PR #314, opened 2015-03-18
Implement a hooks framework

byroot:
@gmalette @davidcornu @cjoudrey I figured you guys might be interested.

It do not publish anything yet, and there is no UI nor API to register hooks, but it's a start.

Every delivery is recorded for ease of debugging. Also hooks can be registered either on a stack or globally.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -50,6 +50,9 @@ Style/SignalException:
 Style/RaiseArgs:
   Enabled: false
 
+Style/ModuleFunction:
+  Enabled: false
+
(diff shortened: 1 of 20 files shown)


mattvleming:
Is there a way to register hooks now, either through a UI or an API? How can one add a webhook with shipit?

byroot:
@mattvleming API (or rails console) is the only way for now: https://github.com/Shopify/shipit-engine/blob/master/test/controllers/api/hooks_controller_test.rb

PR #373, opened 2015-05-21
Implement Status groups

byroot:
The idea here is to fold commits with multiple statuses into one. Either on hover or on click, a popin would display the detail.

The "main" status is defined by order of importance of states: failure -> error -> pending -> success

cc @vernalkick


app/models/status_group.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def to_partial_path
+    'statuses/group'
+  end
+
+  private
+
+  def find_main_status
+    STATES_PRIORITY.each do |state|
```
rafaelfranca: I think you can use `find` here too instead of a `return` (But my personal preference is its alias `detect`).

app/models/status_group.rb
```diff
@@ -0,0 +1,40 @@
...
+      status = @statuses.find { |s| s.state == state }
+      return status if status
+    end
+    fail "No status found, this is not supposed to happen"
```
rafaelfranca: If you chose for the `find` comment above this will have to be changed to a `or` condition.

app/models/status_group.rb
byroot: :facepalm:

app/views/statuses/_group.html.erb
```diff
@@ -0,0 +1,13 @@
+<div class="status-group">
+  <%= render 'statuses/status', status: group %>
+
+  <div class="status-items">
+    <% group.statuses.each do |status| %>
```
rafaelfranca: This partial is a little bit confusing, for what I could get `group` is a instance of `StatusGroup` and `status` is a instance of `Status` right?

If that is the case I don't see why `StatusGroup` have all those methods.

app/models/status_group.rb
byroot: Nevermind. While applying that change I remembered why it isn't a find. If it was this method would return a `String`, while we are looking for a `Status`.

app/views/statuses/_group.html.erb
byroot: > why StatusGroup have all those methods.

So that we can feed it to the regular `_status` partial. Commits with a single status go directly to the `_status` partial. It's a way to use polymorphism to avoid too many `if / else` in views.

app/models/status_group.rb
rafaelfranca: :+1:

app/views/statuses/_group.html.erb
rafaelfranca: :+1:. It was not clear to me that `StatusGroup` were being used in the regular `_status` partial. But just saw the render call on the top of the file after your explanation.

byroot:
Marking this as waiting for design.

designers

app/views/statuses/_group.html.erb
```diff
@@ -0,0 +1,13 @@
...
+      <div class="status-item <%= status.state %>">
+        <strong><%= status.context %></strong>
+        <span><%= status.description %></span>
+        <a href="<%= status.target_url%>" target="_blank">Details</a>
```
rafaelfranca: :nail_care:. Space before the `%>`

byroot: fixed thanks :)

app/views/statuses/_group.html.erb
```diff
@@ -0,0 +1,13 @@
+<div class="status-group">
+  <%= render 'statuses/status', status: group %>
+
+  <div class="status-items">
```
rafaelfranca: It seems that the CSS is using BEM. Following the BEM rules should not this be `status-group__items` since this class only have effect in the presence of the top level element?

byroot: You are right. I think the BEM introduction is an ongoing effort by @vernalkick. Unfortunately he have very little time to put on shipit.

I'm really bad at CSS and design, so I can't really finish the redesign project, but I'll try to find some time to improve the classes naming consistency.

gmalette: @funionnn is this something you can help with?

byroot:
Here's how GitHub does it:

capture d ecran 2015-05-27 a 09 25 51

byroot:
@rafaelfranca also another annoyance with Travis that I haven't figured out yet is that it run the pull request twice... Any idea why it would do that? Shipit is far from the only repo where I saw this.

dwradcliffe:
@byroot Travis is testing the branch commit and the newly created merge commit by GitHub, two different things.

byroot:
Oh! That make sense thanks.

vernalkick:
Will try to take a stab at it tonight.

byroot:
Ok so I did the best I could:
capture d ecran 2015-05-30 a 15 53 00

Not sure if it's good enough to be merged though :/

Sirupsen:
This is pretty awesome.

gmalette:
@byroot can you rebase, maybe @vernalkick can make sure the stacking is OK

Sirupsen:
The `z-index` should probably be bumped so it doesn't fall below the other buttons—but overlays them

byroot:
Rebased.

vernalkick:
Updated the design and fixed the z-index issues.

gmalette:
Error has been yellow in the circle (and is now red in the hover)
screen shot 2015-06-22 at 3 41 36 pm

vernalkick:
Updated it to red

Gemfile.lock
```diff
@@ -10,6 +10,7 @@ PATH
       faraday (~> 0.9.1)
       faraday-http-cache (= 1.1.0)
       jquery-rails (~> 4.0.4)
+      octicons-rails (~> 2.1)
```
gmalette: Is this needed?

byroot: Not anymore it seems. I'll remove it.

byroot:
LGTM, but I'll nitpick one thing: On GitHub `failure` is a red cross and `error` (which very rarely happen) is a black cross.

PR #376, opened 2015-05-21
Add ignore ci action to stacks controller

JakeCataford:
Closes #372 

This does a few things, it hooks up the link to ignore CI from #372 to a new action on the stack controller. Then it adds a settings field for toggling the ignore_ci option on and off.

/cc @byroot @gmalette @funionnn


JakeCataford:
@gmalette updated this so that we use the update action instead, would appreciate another :eyeglasses:

app/controllers/stacks_controller.rb
```diff
@@ -1,5 +1,5 @@
 class StacksController < ShipitController
-  before_action :load_stack, only: %i(update destroy settings sync_webhooks clear_git_cache refresh)
+  before_action :load_stack, only: %i(update destroy settings sync_webhooks clear_git_cache refresh ignore_ci)
```
gmalette: This shouldn't be required anymore

gmalette:
:shipit:

PR #379, opened 2015-05-21
Documentation

byroot:
WIP


documentation/shipit.yml.md
```diff
@@ -0,0 +1,118 @@
...
+  unlock
+
+dependencies
+  bundler
+
+### review
+
+<code>review.checklist</code>
+
+Array of strings. Can contain and HTML tags. Users are required to check each of them before being able to deploy.
```
rafaelfranca: I think there is an extra `and` there. It should be `Can contain HTML tags.`

documentation/shipit.yml.md
```diff
@@ -0,0 +1,118 @@
...
+
+curl --silent https://app.shopify.com/services/ping/version
+
+### deploy
+
+<code>deploy.override</code>
+
+Array of shell commands required to deploy the application. Shipit will try to infer it from the repository structure, but you can change the default inference.
+
+exec ./script/multi_deploy.rb bundle exec cap $ENVIRONMENT deploy
```
rafaelfranca: Is this the default inference? If so, it may be good to make it explicit.

byroot:
Oh I forgot about that PR. I should rebase from master since this was initially an extraction gimme a minute.

byroot:
Yeah It's what I though. That file basically became the README. Closing this.

documentation/shipit.yml.md
byroot: Good catch. I'll update master.

documentation/shipit.yml.md
byroot: No it's a custom one Rory grabbed from one of our repos. The default inference is based on multiple heuristics: https://github.com/Shopify/shipit-engine/tree/master/app/models/deploy_spec

PR #384, opened 2015-05-22
Move secrets.yml to secrets.example.yml and copy on bootstrap

gmalette:


script/bootstrap
```diff
@@ -4,5 +4,6 @@
 set -e
 set -x
 
+cp -n test/dummy/config/secrets.example.yml test/dummy/config/secrets.yml || true
```
byroot: `cp cp -n test/dummy/config/secrets{.example,}.yml` but whatever :p

gmalette:
@byroot

byroot:
:shipit:

Issue #318, opened 2015-03-29
Handle deleted repositories on GitHub side


byroot:
## Exception in Shipit

**Octokit::NotFound** in **resque#default**
GET https://authme.shopify.com/github/repos/shopify/brochure-app4/commits?sha=app4: 404 - Not Found // See: https://developer.github.com/v3

[View on Bugsnag](https://bugsnag.com/shopify/shipit/errors/551313196b1bc35a632e5f6c?event_id=551313981b0b10941cdfd8f3)

## Stacktrace

    app/models/stack.rb:151 - github_commits
    app/jobs/github_sync_job.rb:15 - perform
    app/jobs/background_job.rb:11 - block in perform
    app/jobs/background_job.rb:19 - with_timeout
    app/jobs/background_job.rb:10 - perform

[View full stacktrace](https://bugsnag.com/shopify/shipit/errors/551313196b1bc35a632e5f6c?event_id=551313981b0b10941cdfd8f3)

gmalette:
It kinda sucks that this was in bugsnag

byroot:
No big deal. It's easy to repro. I think the way to go would be:

  - Have a column to mark stacks as not existing GitHub side
  - Flip that boolean when we hit 404s etc.
  - Display a message when it's true
  - Hitting refresh flip it back to false.

WDYT?

gmalette:
I was thinking the same, but I'm wondering if the sync job is the only place it can occur

byroot:
No there is several other places, like webhook setup

gmalette:
> Hitting refresh flip it back to false.

Can you undelete stacks?

byroot:
Fixed a while ago in https://github.com/Shopify/shipit-engine/pull/443

PR #389, opened 2015-05-22
Remove unused code

gmalette:
I was trying to find other places that could be vulnerable to the same problem as #318, and I found this code that doesn't seem to be used.


byroot:
:shipit:

PR #391, opened 2015-05-22
Configure sidekiq properly in the template

byroot:
@arthurnn for review.

I'm not quite fan of the `if Rails.env.production?` so if you have a better idea on how to do that?

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -57,6 +57,18 @@ production:
     # SSH_AUTH_SOCK: /foo/bar # You can set environment variable that will be present during deploys.
 CODE
 
+initializer 'sidekiq.rb', <<-CODE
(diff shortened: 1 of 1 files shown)


template.rb
```diff
@@ -57,6 +57,18 @@
...
+  
+  Sidekiq.configure_server do |config|
+    config.redis = { url: Shipit.redis_url.to_s }
+  end
+
+  Sidekiq.configure_client do |config|
+    config.redis = { url: Shipit.redis_url.to_s }
+  end
+CODE
+
```
arthurnn: template.rb:62:1: C: Trailing whitespace detected.

arthurnn:
I think this works for now :+1:

PR #392, opened 2015-05-22
Disable search and contributor filtering up to 10 stacks

byroot:
Those two features while indispensable for organisations like Shopify with hundreds of stacks, is quite useless and even confusing for small organizations like rubygems that only have an handful of stacks.

@gmalette and @arthurnn for technical review please.

@funionnn and @vernalkick for CSS review please.

Before: 
capture d ecran 2015-05-22 a 14 03 31

After:
capture d ecran 2015-05-22 a 14 03 50

### DIFF
--- a/app/controllers/stacks_controller.rb
+++ b/app/controllers/stacks_controller.rb
@@ -8,7 +8,7 @@ class StacksController < ShipitController
   def index
     @user_stacks = current_user.stacks_contributed_to
 
-    @stacks = Stack.order('(undeployed_commits_count > 0) desc', tasks_count: :desc)
+    @stacks = Stack.order('(undeployed_commits_count > 0) desc', tasks_count: :desc).to_a
   end
 
   def show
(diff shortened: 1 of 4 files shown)


gmalette:
:+1:

app/controllers/stacks_controller.rb
```diff
@@ -8,7 +8,7 @@ def new
   def index
     @user_stacks = current_user.stacks_contributed_to
 
-    @stacks = Stack.order('(undeployed_commits_count > 0) desc', tasks_count: :desc)
+    @stacks = Stack.order('(undeployed_commits_count > 0) desc', tasks_count: :desc).to_a
```
arthurnn: pourquoi `to_a` ?

byroot: Because we call `size` twice on `@stacks` in the view. So we might as well fetch it once and for all in the controller.

funionnn:
Looks good!

app/controllers/stacks_controller.rb
arthurnn: :+1:

arthurnn:
:shipit:

vernalkick:
I don't think I would disable search, otherwise LGTM.

byroot:
> I don't think I would disable search, otherwise LGTM.

Hum, I'm not sure either. Let's try like that anyway. It's easy to change.

PR #394, opened 2015-05-22
Add git commands to template

arthurnn:
More convenient to init the repo with git.

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -75,4 +75,8 @@ CODE
 
 after_bundle do
   rake 'railties:install:migrations db:create db:migrate'
+
+  git :init
+  git add: '.'
+  git commit: "-a -m 'Initial commit'"
 end
--- a/template.rb
+++ b/template.rb
@@ -13,7 +13,6 @@ gem 'shipit-engine', github: 'Shopify/shipit-engine'
 say("These configs are for development, you will have to generate them again for production.",
     Thor::Shell::Color::GREEN, true)
 
-
 say("Shipit requires a GitHub application to authenticate users. "\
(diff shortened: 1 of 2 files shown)


arthurnn:
review @dwradcliffe @byroot

byroot:
Is the secrets.yml ignored in dev though?

arthurnn:
> Is the secrets.yml ignored in dev though?

I guess we should use ENV variables even for dev, and probably use foreman locally? WDYT ?

byroot:
I'm not super familiar with foreman in dev. Does it load env from a .env file or something? If so then yes :+1:

arthurnn:
Yes, it loads the `.env` file.. we can create that file on the template and ignore it on git.

byroot:
Go for it man!

arthurnn:
Ok done.. So this allows some good things:
- `foreman start` will now start sidekiq and rails in development, so we can use sidekiq locally
- foreman will use the .env file. Which is also used for the docker container I am creating. So if you already have the `.env` file from a development version, you can just copy/paste it to the container version. Not that useful, but at least is consistent. 
- We can push the secrets file now, as it doesnt contain sensitive data.

template.rb
```diff
@@ -8,12 +8,11 @@
 
 gem 'sidekiq'
 gem 'thin'
-gem 'shipit-engine'
+gem 'shipit-engine', github: 'Shopify/shipit-engine'
```
gmalette: should probably change before merge?

gmalette:
nice! I can't wait

PR #397, opened 2015-05-22
Improve the layout of the settings page

vernalkick:
Still not perfect, but a good improvement over the current design.

cc @gmalette


gmalette:
:+1:

PR #401, opened 2015-05-22
Fix migrations for postgres

arthurnn:
when trying to run shipit on postgres was getting the following error on migration run:
```
StandardError: An error has occurred, this and all later migrations canceled:

PG::SyntaxError: ERROR:  syntax error at or near "("
LINE 1: SELECT 'boolean(1)'::regtype::oid
```

Removing the limit on the boolean column should fix the problem.

review @byroot

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -4,6 +4,18 @@
 set -e
 set -x
 
+#!/bin/bash
+
 cp -n test/dummy/config/secrets.example.yml test/dummy/config/secrets.yml || true
+
+case $CIRCLE_NODE_INDEX in
+  1)
+    cp test/config/database.postgres.yml test/dummy/config/database.yml
+    ;;
+  2)
+    cp test/config/database.sqlite.yml test/dummy/config/database.yml
+    ;;
+esac
+
 bundle install
 bundle exec rake db:create db:schema:load db:seed
--- a/script/circleci
+++ b/script/circleci
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+case $CIRCLE_NODE_INDEX in
+  0 )
+    bundle exec rake test
+    ;;
+  1)
+    bundle exec rake test
+    ;;
+  2)
+    bundle exec rake test
+    ;;
+  3)
+    bundle exec rubocop
+    bundle exec rake db:seed # Make sure db:seed still works
+    ;;
+esac
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,8 @@ source 'https://rubygems.org'
 gemspec
 
 gem 'mysql2'
+gem 'pg'
+gem 'sqlite3'
 
 group :development, :test do
   gem 'faker'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -160,6 +160,7 @@ GEM
       omniauth (~> 1.2)
     parser (2.2.2.3)
       ast (>= 1.1, < 3.0)
+    pg (0.18.2)
     powerpack (0.1.1)
     pry (0.10.1)
       coderay (~> 1.1.0)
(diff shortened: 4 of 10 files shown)


gmalette:
Is this replayable, or will new schemas generated on mysql have the limit?

gmalette:
:+1:

arthurnn:
i tested on sqlite3 and PG,.. should be fine on mysql too

gmalette:
:+1:

byroot:
I modified this PR to run test under MySQL/ PG / Sqlite: https://circleci.com/gh/Shopify/shipit-engine/1636

We have one bug with sqlite it seems (look very much like our intermittent issue but somehow consistent).

arthurnn:
> I modified this PR to run test under MySQL/ PG / Sqlite:

Awesome

script/bootstrap
```diff
@@ -4,6 +4,18 @@
 set -e
 set -x
 
+#!/bin/bash
```
rafaelfranca: Is this shebang needed? It seems to be the same as the one on the top level of the file.

script/circleci
```diff
@@ -0,0 +1,17 @@
...
+    ;;
+  1)
+    bundle exec rake test
+    ;;
+  2)
+    bundle exec rake test
+    ;;
+  3)
+    bundle exec rubocop
+    bundle exec rake db:seed # Make sure db:seed still works
```
rafaelfranca: You already test seed at the bootstrap. It does make sense to test again?

script/bootstrap
byroot: Yeah look like a mistake.

script/circleci
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+case $CIRCLE_NODE_INDEX in
+  0 )
+    bundle exec rake test
```
rafaelfranca: Maybe it would be good to print which database we are testing here. Right we don't have this information on the ci's `script/circleci` output

script/circleci
byroot: Nope it doesn't :)

script/circleci
gmalette: if this is the "good way" to do it on Circle, can we not use Travis now?

byroot: Excellent idea.

byroot: Also note to self:  the `bundle exec rake test` duplication can easily be solved with `*)`

rafaelfranca: > if this is the "good way" to do it on Circle, can we not use Travis now?

Travis build matrix seems better than Circle CI matrix so far, and it has the advantage that every single OSS project on GitHub use it. If you guys want to give travis a try I can work on it so we can see how it looks like.

byroot: > if this is the "good way" to do it on Circle, can we not use Travis now?

Yeah we could. I just tend to be despaired by Travis queueing, but maybe it's just me.

gmalette: Yeah but as @rafaelfranca pointed out, at least they have decent build matrixes. Is there a way to get around the queues using :dollar: ?

gmalette: Or... buildkite?

byroot: > Or... buildkite?

If you give me your credit card to pay for the ec2 instances then sure no problem :p 

More seriously, yes Travis build matrix is definitely better and people are used to it. Also I think Shopify pays for Travis so you should not experience as much queueing than on my personal projects.

@rafaelfranca I'd say let's give a try to travis in this PR. You want to handle it?

rafaelfranca: Sure. It will help me to get used with the code.

byroot: :heart:

byroot:
Replaced by https://github.com/Shopify/shipit-engine/pull/402. 

I'll open another issue for sqlite3 support.

PR #402, opened 2015-05-27
Travis CI setup

rafaelfranca:
Here is an example build https://travis-ci.org/rafaelfranca/shipit-engine/builds/64195217.

In my account the jobs started in less than 10 seconds, so maybe the queue size will be not a problem (but I don't know if this is because people on Travis gave me some special treatment since I was working with them beta testing the new container setup with Rails).

There is still a failure because a problem with datetime comparing on sqlite that I'm still investigating.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,8 @@ source 'https://rubygems.org'
 gemspec
 
 gem 'mysql2'
+gem 'pg'
+gem 'sqlite3'
 
 group :development, :test do
   gem 'faker'
--- a/script/cibuild
+++ b/script/cibuild
@@ -21,3 +21,4 @@ case $DATABASE in
 esac
 
 bundle exec rake db:create db:schema:load test
+bundle exec rake db:seed
--- a/script/cibuild
+++ b/script/cibuild
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+
+# Output all commands, stop on failures
+set -e
+set -x
+
+cp -n test/dummy/config/secrets.example.yml test/dummy/config/secrets.yml || true
+
+case $DATABASE in
+  mysql)
+    echo "Testing PostgreSQL"
+    ;;
+  pg)
+    cp test/dummy/config/database.postgresql.yml test/dummy/config/database.yml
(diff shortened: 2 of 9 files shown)


Gemfile
```diff
@@ -3,6 +3,8 @@ source 'https://rubygems.org'
 gemspec
 
 gem 'mysql2'
+gem 'pg'
+gem 'sqlite3'
```
gmalette: Can anyone explain to me why this is in the `Gemfile` and not in the `gemspec`? I'm not sure I understand which is for what

gmalette:
I think this is much nicer. If queues are an issue we can go back

Gemfile
byroot: Gemspec are them gem dependencies. While  in the gemfile you put things that are only for dev/test

byroot:
LGTM. Feel free to mark sqlite as soft-fail and we can fix it in another PR.

Gemfile
rafaelfranca: You can put development dependencies at gemspec too but you will loose the ability to define bundle groups. The only advantage of having development dependencies is that they will be present on the rubygems pages like this https://rubygems.org/gems/arel.

gmalette: :+1: thanks that clarifies it

rafaelfranca:
Done. I also added `db:seed` to the test script.

Now you just need to enable travis builds.

byroot:
I was able to tell you to merge yourself, I guess I'm not used to external contributors yet :p.

byroot:
Builds are now enabled in travis: https://travis-ci.org/Shopify/shipit-engine/branches

script/cibuild
```diff
@@ -0,0 +1,24 @@
...
+
+# Output all commands, stop on failures
+set -e
+set -x
+
+cp -n test/dummy/config/secrets.example.yml test/dummy/config/secrets.yml || true
+
+case $DATABASE in
+  mysql)
+    echo "Testing PostgreSQL"
```
arthurnn: Shouldnt say mysql here?

rafaelfranca: Yup

PR #404, opened 2015-05-27
Store the milliseconds on the database column

### LINKED ISSUE #403: Test suite doesn't pass with sqlite3
See https://travis-ci.org/rafaelfranca/shipit-engine/jobs/64195221 for example.

I think it's because of the way string comparison works (case sensitivity) but not 100% sure. 

cc @rafaelfranca

rafaelfranca:
This will correct the comparison for databases that have millisecond precision and work like before for databases that don't have.

Also memoize the github_status object on the tests so the milliseconds are the same for the created_at property.

Closes #403

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,6 @@ env:
 rvm:
   - 2.1
 matrix:
-  allow_failures:
-    - env: "DATABASE=sqlite3"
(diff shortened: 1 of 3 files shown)


byroot:
Thanks!

PR #408, opened 2015-05-27
Test with Ruby 2.2

rafaelfranca:


byroot:
Thanks, I'll merge on :green_heart:

byroot:
Not to :trollface: but the CircleCI build finished 8 minutes ago and the Travis one didn't started yet :p

rafaelfranca:
lol. That is what I was looking. Very strange.

rafaelfranca:
Turbograph seems to be the reason to have not started the build yet

byroot:
Yeah I wonder if the core issue is not that Travis limit the number of concurrent builds for an organization and that liquid / turbograft / active_merchant just keep eating all the allocated capacity.

rafaelfranca:
Yes, it has organization limit. One thing that we can do to improve the situation is distribute the limit across the projects equally.

byroot:
How does that work?

rafaelfranca:
On the concurrent jobs option of this screen

screen shot 2015-05-27 at 11 13 57 pm

PR #410, opened 2015-05-30
Restore sidebar for plugins

byroot:
We still need the sidebar for visualisation plugins. It should be hidden by default and show up as soon as a plugin render something.


byroot:
Thanks @jeromegn very helpful!

byroot:
LGTM now. I'll merge on green.

Issue #411, opened 2015-06-01
Current design can't acomodate more than 2 or 3 custom tasks


byroot:
Custom tasks buttons can easily go off screen and create an horizontal scroll.

According to @gmalette, @funionnn and @vernalkick original plan was to go for a dropdown if the task has more than a handful tasks.
- [ ] Improve seeds so that it's easily reproducible 
- [ ] Fix the design.

Maybe @jeromegn can help? ;)

funionnn:
Can someone include a screenshot?

byroot:
Screenshot of the issue
![capture d ecran 2015-06-01 a 09 48 32](https://cloud.githubusercontent.com/assets/44640/7914503/778cb832-0845-11e5-8007-8cbe1036f62d.png)


Mockup by @vernalkick 
![](https://s3.amazonaws.com/f.cl.ly/items/0k3j233h333O402Y1O2l/Image%202015-06-01%20at%2010.01.42%20AM.png)

casperisfine:
This was fixed a long time ago.

PR #412, opened 2015-06-01
Allow stack buttons to wrap on multiple lines

vernalkick:
Will make things a bit better until we have the overflow menu

### DIFF
--- a/app/assets/stylesheets/_structure/_layout.scss
+++ b/app/assets/stylesheets/_structure/_layout.scss
@@ -21,6 +21,7 @@
 
 .header__inner {
   display: flex;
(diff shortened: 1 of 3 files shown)


byroot:
:heart: 

Ref: https://github.com/Shopify/shipit-engine/issues/411

app/assets/stylesheets/_structure/_layout.scss
```diff
@@ -30,7 +31,11 @@
 }
 
 .header__page-actions {
```
rafaelfranca: Any specific reason to keep this style blank?

PR #416, opened 2015-06-01
Fix links in checklist

vernalkick:
cc @gmalette

### DIFF
--- a/app/views/deploys/_checklist.html.erb
+++ b/app/views/deploys/_checklist.html.erb
@@ -4,10 +4,10 @@
   </header>
 
   <ul class="deploy-checklist">
-    <%- stack.checklist.each do |check| -%>
+    <%- stack.checklist.each_with_index do |check, index| -%>
       <li class="deploy-checklist__item">
-        <%= check_box_tag :checklist, nil, false, class: 'deploy-checklist__item__checkbox' %>
-        <label class="deploy-checklist__item__label">
+        <%= check_box_tag :checklist, nil, false, class: 'deploy-checklist__item__checkbox', id: "checkbox_#{index}" %>
+        <label class="deploy-checklist__item__label" for="checkbox_<%= index %>">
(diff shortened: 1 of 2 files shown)


vernalkick:
Noticed we're missing `for=` attrs on the labels

app/views/deploys/_checklist.html.erb
```diff
@@ -6,8 +6,8 @@
   <ul class="deploy-checklist">
     <%- stack.checklist.each do |check| -%>
       <li class="deploy-checklist__item">
+        <%= check_box_tag :checklist, nil, false, class: 'deploy-checklist__item__checkbox' %>
         <label class="deploy-checklist__item__label">
-          <%= check_box_tag :checklist, nil, false, class: 'deploy-checklist__item__checkbox' %>
```
byroot: Correct me if I'm wrong, but this change won't allow you to click on the text to check the box anymore right?

Should we add an `id` and a `for` attribute?

gmalette: yeah he asked me to take care of it. @vernalkick can you add the proper classes such that the "Create Deploy" button will be disabled when it has a `disabled` class?

gmalette:
@vernalkick I added the `id` and `for`

vernalkick:
:cool: :ship: ?

gmalette:
@vernalkick can you add the proper CSS such that the "Create Deploy" button will be disabled when it has a `disabled` class?

gmalette:
then yes, totally :ship:

vernalkick:
Unrelated to this PR?

gmalette:
/shrug depending how you see it. Feel free to merge it if you think it's unrelated

byroot:
@gmalette I think https://github.com/Shopify/shipit-engine/pull/418 fixes this as well.

gmalette:
Ah that makes sense

Issue #414, opened 2015-06-01
Deploys history


byroot:
Right now there is no way to search for previous deploys past the last 10.

We should allow to paginate in the deploy history of each stacks.

cc @pushrax

byroot:
Fixed by #417

PR #417, opened 2015-06-02
Deploy history

byroot:
Fixes: https://github.com/Shopify/shipit-engine/issues/414

This is work in progress but I'm about to onboard.

capture d ecran 2015-06-01 a 21 03 41

capture d ecran 2015-06-01 a 21 03 49

- I removed the rollbacks button because I don't think it's a good idea to rollback that far. But If you guys think I should let them let me know.
- Since I reuse the API paginator I'm limited to `first` and `next` links. We can always improve that latter.
- [x] Needs tests

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -12,7 +12,7 @@ PATH
       jquery-rails (~> 4.0.3)
       octokit (~> 3.8.0)
       omniauth-github (~> 1.1.2)
-      pubsubstub (~> 0.0.11)
+      pubsubstub (~> 0.0.9)
       rails (~> 4.2.1)
       rails-timeago (~> 2.11.1)
       redis-objects (= 1.2.0)
--- a/app/controllers/tasks_controller.rb
+++ b/app/controllers/tasks_controller.rb
@@ -1,5 +1,15 @@
 class TasksController < ShipitController
+  include Pagination
+
   before_action :stack
+
+  self.default_page_size = 20
+
+  def index
+    paginator = paginate(stack.tasks)
+    @tasks = paginator.to_a
+    @links = paginator.links
+  end
 
   def new
     @definition = stack.find_task_definition(params[:definition_id])
--- a/app/views/tasks/index.html.erb
+++ b/app/views/tasks/index.html.erb
@@ -1,0 +1,25 @@
+<%= render partial: "stacks/header", locals: { stack: @stack } %>
+
+<div class="wrapper">
+  <section>
+    <ul class="task-list">
+      <%= render @tasks, read_only: true %>
+    </ul>
+  </section>
+</div>
+
+<div class="wrapper">
+  <section class="pagination">
+    <% if params[:since].present? %>
+      <%= link_to 'Start', @links[:first], class: 'header__btn btn' %>
+    <% else %>
+      <%= link_to 'Start', '', class: 'header__btn btn btn--disabled disabled' %>
+    <% end %>
+
+    <% if url = @links[:next] %>
+      <%= link_to 'Older', url, class: 'header__btn btn' %>
+    <% else %>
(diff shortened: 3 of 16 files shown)


Gemfile.lock
```diff
@@ -12,7 +12,7 @@ PATH
       jquery-rails (~> 4.0.3)
       octokit (~> 3.8.0)
       omniauth-github (~> 1.1.2)
-      pubsubstub (~> 0.0.11)
+      pubsubstub (~> 0.0.9)
```
rafaelfranca: Any specific reason for the downgrade?

byroot: It's a mistake /o\

app/controllers/concerns/paginator.rb
```diff
@@ -0,0 +1,49 @@
+class Paginator
```
rafaelfranca: Maybe this class should be moved to `lib`. It is used on controllers but it is not a concern related just to the controller layer. Also, there is already a lot of logic here to justify unit tests for it.

config/routes.rb
```diff
@@ -44,6 +44,7 @@
     resources :rollbacks, only: %i(create)
     resources :tasks, only: %i(show) do
       collection do
+        get '' => 'tasks#index', as: :index
```
rafaelfranca: I think you can change the `only` option to include `index` and get the same result.

app/controllers/tasks_controller.rb
```diff
@@ -1,6 +1,16 @@
 class TasksController < ShipitController
+  include Pagination
+
   before_action :stack
 
+  self.default_page_size = 20
+
+  def index
+    @paginator = paginate(stack.deploys)
```
rafaelfranca: This variable is not being used in any view so I think we can change it to be a local variable instead a instance variable.

app/views/tasks/index.html.erb
```diff
@@ -0,0 +1,25 @@
...
+
+<div class="wrapper"
+  <section>
+    <ul class="commit-list">
+      <%= render @tasks, read_only: true %>
+    </ul>
+  </section>
+</div>
+
+<div class="wrapper"
```
rafaelfranca: Missing close `>`

app/views/tasks/index.html.erb
```diff
@@ -0,0 +1,25 @@
+<%= render partial: "stacks/header", locals: { stack: @stack } %>
+
+<div class="wrapper"
```
rafaelfranca: Missing close `>`

app/views/tasks/index.html.erb
```diff
@@ -0,0 +1,25 @@
...
+  <section>
+    <% if params[:since].present? %>
+      <%= link_to 'Start', @links[:first], class: 'header__btn btn' %>
+    <% else %>
+      <%= link_to 'Start', '', class: 'header__btn btn btn--disabled disabled' %>
+    <% end %>
+
+    <% if url = @links[:next] %>
```
rafaelfranca: This line give a ruby warning. I hate this warning too but it is a good practice to write warning free code. To do so put the `url = @links[:next]` inside parentheses

rafaelfranca:
Just a few small comments but the feature looks good.

byroot:
All good points thanks!

app/views/tasks/index.html.erb
byroot: Not sure what the warning is supposed to be, I tried:

```ruby
@foo = {}

if url = @foo[:next]
  p url
end
```

And it didn't generate any warning.

rafaelfranca: O_O I can't reproduce this warning anymore but I got it many times on the Rails repository. If I remember the situation I'll let you know but this is fine as it is.

byroot:
I centered the navigation links. I believe this is good to go.

PR #420, opened 2015-06-03
Improve the missing settings page style

byroot:
Still not great but slightly better:

capture d ecran 2015-06-03 a 10 08 02

Since this page will likely be our users first impression of Shipit it's important that it looks good and welcoming.

I did the best I could here, but I think we need a second pass :/

IMO this is the last real blocker for a proper announce of Shipit existence.


gmalette:
@vernalkick can you paste a screenshot of what you had please?

byroot:
I'll merge this as is since it's already an improvement.

PR #421, opened 2015-06-04
Allow overriding --frozen for bundle install

gmalette:
In the case of Rubygems, they use a git submodule to include gems. In some cases, the `Gemfile.lock` and the submodule don't align perfectly, but they still want to be able to deploy to staging. 

@byroot @dwradcliffe

### DIFF
--- a/app/models/deploy_spec/bundler_discovery.rb
+++ b/app/models/deploy_spec/bundler_discovery.rb
@@ -26,7 +26,9 @@ class DeploySpec
     end
 
     def frozen_flag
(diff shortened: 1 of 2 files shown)


arthurnn:
:+1:

dwradcliffe:
:+1:

byroot:
LGTM. But I wonder if we shouldn't simply provide a `bundler_args` like travis. But this will do it for now we can always revisit later.

gmalette:
`dependencies.bundler.orverride` perhaps?

PR #422, opened 2015-06-04
Fix git branch regex

trishume:
Git branches can contain periods, this is used for branches containing version numbers.
The `\w` character class also already covers the underscore.

@byroot

### DIFF
--- a/app/models/commit.rb
+++ b/app/models/commit.rb
@@ -118,7 +118,7 @@ class Commit < ActiveRecord::Base
   end
 
   def parsed
-    @parsed ||= message.match(/\AMerge pull request #(?<pr_id>\d+) from [\w\-\_\/]+\n\n(?<pr_title>.*)/)
+    @parsed ||= message.match(/\AMerge pull request #(?<pr_id>\d+) from [\w\-.\/]+\n\n(?<pr_title>.*)/)
   end
 
   def schedule_continuous_delivery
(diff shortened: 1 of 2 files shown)


gmalette:
Can you add a test please

trishume:
I was thinking of doing that but then I found that there is no test file for `commit.rb`. I guess I'll go yak-shaving and write a bunch of tests for commit.

gmalette:
Thanks :)

trishume:
Nvm they are all in `commits_test.rb`, just no tests for parsing.

app/models/commit.rb
```diff
@@ -118,7 +118,7 @@ def short_sha
   end
 
   def parsed
-    @parsed ||= message.match(/\AMerge pull request #(?<pr_id>\d+) from [\w\-\_\/]+\n\n(?<pr_title>.*)/)
+    @parsed ||= message.match(/\AMerge pull request #(?<pr_id>\d+) from [\w\-\.\/]+\n\n(?<pr_title>.*)/)
```
rafaelfranca: Inside a character class `.` don't need to be escaped.

trishume:
Added test and removed unnecessary escape.

byroot:
LGTM.

gmalette:
Awesome, thanks!

PR #423, opened 2015-06-05
Bump pubsubstub

gmalette:
Ok so... I made a stupid mistake when refactoring heartbeats for 0.0.10. It crashes the spawned thread and never does the heartbeat. 

Here's the diff https://github.com/gmalette/pubsubstub/compare/325a89070e5569f51ed164c38387c3fb9377f18c...master


byroot:
LGTM.

PR #424, opened 2015-06-07
add --recursive option to git clone

dwradcliffe:
Any repos that have submodules probably need them to build properly.

This option fetches all submodules after the clone.

From the git docs:
> After the clone is created, initialize all submodules within, using their default settings. This is equivalent to running git submodule update --init --recursive immediately after the clone is finished.

@byroot @gmalette


byroot:
:shipit:

rafaelfranca:
:shipit:

PR #427, opened 2015-06-15
Emergency mode

byroot:
We talked about that feature a while ago with @gmalette, the idea being to toogle a flag that allows you to trigger deploys in any circumstances.

@graemej just asked for that feature again, so here it is.

The first commit is a small refactoring of the button, the second is of higher importance.

@gmalette @davidcornu mind giving this a look please?


app/helpers/stacks_helper.rb
```diff
@@ -1,4 +1,16 @@
 module StacksHelper
+  DISABLED_BUTTON = %w(btn--disabled disabled).freeze
```
davidcornu: Is this used anywhere? You're using `classes = %W(btn btn--primary deploy-action #{commit.state})` on line `6`.

davidcornu: Also I'd call this `DISABLED_BUTTON_CLASSES` or something to that effect

app/helpers/stacks_helper.rb
```diff
@@ -26,4 +38,24 @@ def pull_request_link(commit)
   def render_raw_commit_id_link(commit)
     link_to(commit.short_sha, github_commit_url(commit), target: '_blank', class: 'number')
   end
+
+  private
+
+  def deploy_button_disabled?(commit)
```
davidcornu: `can_deploy_commit?` maybe?

app/helpers/stacks_helper.rb
```diff
@@ -26,4 +38,24 @@ def pull_request_link(commit)
...
+    if commit.deployable?
+      return 'Deploy in progress...' if commit.stack.deploying?
+      return 'Deploy'
+    end
+
+    return 'CI Pending...' if commit.pending?
+    return 'CI Failure' if commit.failure?
+    return 'CI Error' if commit.error?
+    'Not Run'
+  end
```
davidcornu: Might read slightly better as a `case`. Up to you.

```ruby
case
when commit.stack.locked? then "Locked"
when commit.stack.deploying? then "Deploy in progress..."
when commit.stack.deployable? then "Deploy"
when commit.pending? then "CI Pending..."
when commit.failure? then "CI Failure"
when commit.error? then "CI Error"
else "Not Run"
end
```

davidcornu:
Minor stuff. Haven't seen UI. Other than that :ship:

gmalette:
> Haven't seen UI.

screen shot 2015-06-15 at 11 37 30 am

byroot:
Here's the UI:

capture d ecran 2015-06-15 a 17 37 14

capture d ecran 2015-06-15 a 17 37 21

davidcornu:
Lol thanks guys. UI LGTM

app/views/stacks/show.html.erb
```diff
@@ -37,6 +37,14 @@
   <section>
     <header class="section-header">
       <h2>Undeployed Commits</h2>
+
+      <div class="commit-list-actions">
+        <% if params[:force] %>
+          <%= link_to 'Exit emergency mode', stack_path(@stack) %>
+        <% else %>
+          <%= link_to 'Enable emergency mode', stack_path(@stack, force: 1) %>
```
gmalette: I find `force` ambiguous, why not use `emergency`?

gmalette:
:+1:

app/views/stacks/show.html.erb
byroot: It was to mirror the deploy API `force` parameter, but you're probably right.

byroot:
Thanks for the review guys!

app/helpers/stacks_helper.rb
rafaelfranca: I think `deploy_button_disabled?` fits more with this case. Usually I use `can_\w+?` something when we method receiver is an actor, like a user, in this case the receiver is the helper and it is related to Stack that is not the actor of the action.

app/helpers/stacks_helper.rb
rafaelfranca: :+1:

app/helpers/stacks_helper.rb
```diff
@@ -1,4 +1,14 @@
 module StacksHelper
+  def deploy_button(commit)
```
rafaelfranca: Maybe some tests for this helper?

dwradcliffe:
:+1: nice feature

app/helpers/stacks_helper.rb
byroot: Indeed. That's a bad habit I have to almost never test views and helpers /o\

app/helpers/stacks_helper.rb
byroot: Sorry I fixed it, but look like I forgot to commit the change. I just pushed it on master.

PR #428, opened 2015-06-15
Retry EventSource on resume

gmalette:
I don't like to reload. Shipit should just stay on top of it, and reconnect when the EventSource is closed by the network going down. 

This PR adds a periodic timer to see if the current source has been able to connect. If it has, it will monitor when it's been disconnected. When it detects a disconnects, it tries to reconnect (up to twice, but it could be more than that). 

I've tried this over the weekend, and it seemed to work as expected for both EM and non-em.

### DIFF
--- a/app/assets/javascripts/shipit/stacks.js.coffee
+++ b/app/assets/javascripts/shipit/stacks.js.coffee
@@ -24,11 +24,20 @@ jQuery ($) ->
     payload = JSON.parse(message.data)
     $('[data-layout-content]').load("#{payload.url} [data-layout-content] > *", -> $('time[data-time-ago]').timeago())
 
+  retries = 0
   listenToEventSource = (url) ->
     source = new EventSource(url)
-    reconnect = -> listenToEventSource(url)
-    #source.onerror = -> setTimeout(reconnect, 3000)
     source.addEventListener 'stack.update', updatePage
+    interval = setInterval ->
+      switch source.readyState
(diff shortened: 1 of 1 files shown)


gmalette:
To top it off, we could probably have some sort of small message saying "There is a problem with real-time updates" if connection/reconnection fails

byroot:
LGTM but I'll try to :tophat: it tomorrow.

davidcornu:
As per IRL conversation with @gmalette, it might be possible to handle this case using `EventSource`'s [`onerror`](https://developer.mozilla.org/en-US/docs/Web/API/EventSource#Attributes) which would make it a bit cleaner (we just have to make sure `readyState` isn't set to `CLOSED` when the normal retry mechanism is being used).

gmalette:
I know the normal retry will keep readyState at Connecting. Just need to verify its at Closed if the normal retry gets exhausted

gmalette:
So the `onerror` doesn't seem to work properly in Chrome (it's fine in Firefox). [I filed an issue ](https://code.google.com/p/chromium/issues/detail?id=501969)

davidcornu:
@gmalette nice one. Hopefully they'll fix it. Do you want to use this approach in the mean time?

gmalette:
I don't mind. It's shit, but it's basically the best I can do given the limitations

davidcornu:
:sheep: :dash:

Issue #426, opened 2015-06-09
Add dynamic "confirm" step before deploy


wvanbergen:
A "confirm" step would be really useful to make sure it's safe to deploy. The content to display would be collecting by running a task on the source code. The main use case for me is to show a list of pending changes, so you know nothing crazy will happen when you deploy.

#### Process

1. Checkout SHA
2. Run a task as defined in shipit.yml
3. Capture output of said task, and display to user.
4. User can press **Continue** or **Abort** [1]
5. **Continue** will continue with the normal deploy procedure

[1] Potentially, if the task returns a non-zero status, the "Continue" button could be disabled. 

#### Examples

- Show pending database migrations that will be applied by a deploy
- Show DNS record changes that will be applied when continuing the deploy.
- Run a task that makes sure the initial state is what you expect it to be.

#### Notes

- The reason you cannot do this as part of a CI PR is that you don't know when a PR will be merged and/or deployed, so this needs to run right before deploying.
- Potentially, this could be merged with, or could be an extension of, the current checklist functionality

Thoughts?

cc @graemej @dalehamel @thegedge @EmilS

byroot:
Interesting. I'll sleep on it

gmalette:
> Potentially, this could be merged with, or could be an extension of, the current checklist functionality

Could it happen after you've created the deploy but before the deployment steps are started? Why is it less than ideal?

byroot:
Yeah the biggest issue with this is that it requires to checkout the repository which currently only happen later.

byroot:
So after putting it some thoughts I think it's an awesome idea but it has major architectural challenges:

  - This must be backgrounded because we have a tacit rule of not shelling out nor reaching any 3rd party service from web actions.
  - Which mean you would need to trigger the check when you display the page, put some kind of spinner and update the page asynchronously.
  - You likely need a debounce / caching mechanism of some sort so that you don't overload the app by refreshing the deploys/new page.

dalehamel:
Some back-story here - we'd like to use shipit to deploy infrastructure using [terraform](https://terraform.io/). It would be ideal if the infrastructure that a project needs were embedded directly into the repo, so that you could have a separate 'infrastructure' deploy and a separate 'code' deploy.

Before deploying the infrastructure, we want to run 'terraform plan' which will show what is expected to happen so that you can confirm and execute the plan as the actual deploy.

[...]

PR #429, opened 2015-06-16
Dynamic confirm step

byroot:
Fixes: https://github.com/Shopify/shipit-engine/issues/426

This is still an heavy work in progress, but I think I solved the architectural issues. I'll continue the work on this tomorrow.

@wvanbergen @dalehamel FYI.


app/models/stack.rb
```diff
@@ -205,6 +205,11 @@ def checklist
     cached_deploy_spec.review_checklist.map(&:strip).select(&:present?)
   end
 
+  def checks?
+    return false unless cached_deploy_spec
```
gmalette: Caching the whole spec was a great idea

lib/shipit.rb
```diff
@@ -110,8 +113,8 @@ def all_settings_present?
     ].all?(&:present?)
   end
 
-  def extra_env
-    secrets.env || {}
+  def env
+    {'SHIPIT' => '1'}.merge(secrets.env || {})
```
gmalette: A lot of this is actually just a great refactoring, would you mind splitting it off?

byroot: Indeed. I'll extract a couple PRs out of this one.

byroot:
@gmalette @davidcornu @rafaelfranca I believe this is good to go (There is one test that fail on CI but I'll figure that out soonish).

Here's what it look like:

capture d ecran 2015-06-29 a 14 27 48

The output is streamed using the same code than for deploys, and the status is reflected by the border color: yellow: in progress, red: failed, green: success.

One thing we might need to add in a followup is a retry button.

@wvanbergen @dalehamel thoughts?

gmalette:
Should we collapse non-stdout or something? @dalehamel for your specific case, what does the checks look like?

byroot:
> Should we collapse non-stdout or something? 

[Buildkite's terminal](http://buildkite.github.io/terminal/) have an interesting feature: it recognize `---` and `+++` as commands to collapse the output. We might do something similar later, but I think this will do for now.

app/assets/javascripts/task/stream.js.coffee
```diff
@@ -24,6 +24,7 @@ class @Stream
     @status = 'running'
 
   init: ({url, text, status}) ->
+    console.log(url, text, status)
```
davidcornu: `console?.log`?

dalehamel:
@byroot @gmalette this looks conceptually good - how many checks can we define? Each is just one command, displayed separately?

I think showing both stderr and stdout has merit. 

@gmalette our use case is that we'd like to use shipit for infrastructure (building AWS instances, etc), but we want to preview the changes before we actually apply them. This can be done in CI as well, but things might change between CI and actually shipping.

app/assets/javascripts/task/stream.js.coffee
byroot: That's a mistake. I'll :fire: it.

byroot:
> how many checks can we define? Each is just one command, displayed separately?

As much as you want, they will be executed sequentially until one returns a non-zero status or that all have been successfully executed.

app/models/commit_checks.rb
```diff
@@ -0,0 +1,92 @@
...
+    @status = status
+  end
+
+  def finished?
+    FINAL_STATUSES.include?(status)
+  end
+
+  def output(since: 0)
+    return fetch if since.zero?
+    redis.getrange('output', since, -1)
```
davidcornu: How much output are we actually planning on having? Supporting `since` seems like overkill no?

davidcornu:
```
Offenses:
app/jobs/perform_commit_checks_job.rb:2:22: C: Space inside parentheses detected.
  def perform(commit: )
                     ^
app/models/commit_checks.rb:62:38: C: Space inside parentheses detected.
  def build_commands(commands, chdir: )
```

:trollface:

app/models/commit_checks.rb
byroot: I just don't want to bork the app if a command start to output megabytes of text.

> “There's an old saying in Tennessee — I know it's in Texas, probably in Tennessee — that says, fool me once, shame on — shame on you. Fool me — you can't get fooled again.”

> -- George W Bush

app/assets/javascripts/task/notifications.js.coffee.erb
```diff
@@ -5,12 +5,15 @@ class @Notifications
     error: '<%= image_path "deploy_error.jpg" %>'
 
   @init: (outputStream) ->
-    outputStream.addEventListener('status', new this().statusUpdated)
+    if repoName = $('[data-repo-full-name]').attr('data-repo-full-name')
+      outputStream.addEventListener('status', new this(repoName).statusUpdated)
```
davidcornu: `new this` looks odd to me but I guess that's the only way of supporting inheritance.

app/assets/javascripts/task/notifications.js.coffee.erb
```diff
@@ -5,12 +5,15 @@ class @Notifications
...
+
+  constructor: (@repoName) ->
 
   statusUpdated: (task) =>
     return if task.status in ['pending', 'running']
     return unless $.notifyCheck() == $.NOTIFY_ALLOWED
-    $.notify(@url(task), @title(task), @message(task))
+    $.notify(@url(task), repoName, @message(task))
```
davidcornu: `@repoName`?

app/models/commit_checks.rb
```diff
@@ -0,0 +1,92 @@
...
+    commands = StackCommands.new(stack)
+    commands.with_temporary_working_directory(commit: commit) do |directory|
+      deploy_spec = DeploySpec::FileSystem.new(directory, stack.environment)
+      Bundler.with_clean_env do
+        capture_all(build_commands(deploy_spec.dependencies_steps, chdir: directory))
+        capture_all(build_commands(deploy_spec.review_checks, chdir: directory))
+      end
+    end
+  rescue Command::Error
+    self.status = 'failed'
```
davidcornu: Does this get logged anywhere?

davidcornu: I guess `capture_all` would give you debug info

byroot: Yeah the status is so the client side knows it's over. The command status etc is appended to the output.

app/models/commit_checks.rb
```diff
@@ -0,0 +1,92 @@
...
+  rescue
+    self.status = 'error'
+    raise
+  else
+    self.status = 'success'
```
davidcornu: Any reason not to put this after line `18`?

app/models/commit_checks.rb
```diff
@@ -0,0 +1,92 @@
...
+  end
+
+  def fetch
+    @output ||= begin
```
davidcornu: Having both a `@output` instance variable that is populated by `fetch` and a `output` method that does something completely different seems odd to me.

wvanbergen:
@byroot Does displaying this page create a deploy lock? Or is is potentially possible to have a bunch of people requesting this page at the same time, and/or somebody else deploying when showing this page?

app/models/commit_checks.rb
byroot: Not really no.

app/models/commit_checks.rb
```diff
@@ -0,0 +1,92 @@
...
+    redis.set('status', status, ex: OUTPUT_TTL)
+    @status = status
+  end
+
+  def finished?
+    FINAL_STATUSES.include?(status)
+  end
+
+  def output(since: 0)
+    return fetch if since.zero?
```
davidcornu: Doesn't this mean you can call `output` and get nothing back if `since > 0` and the output hasn't been populated yet?

byroot:
> Does displaying this page create a deploy lock? Or is is potentially possible to have a bunch of people requesting this page at the same time, and/or somebody else deploying when showing this page?

@wvanbergen yes this page is totally independent from a deploy. The commands are executed in a temporary clone of the stack.

If 2 people are visualizing this page for the same commit at the same, they'll share the output, but that's all.

app/models/commit_checks.rb
byroot: Indeed.

app/models/commit_checks.rb
davidcornu: Sure

app/models/commit_checks.rb
byroot: I'll see if I can improve this class encapsulation.

davidcornu: Shouldn't we instead check of the command has been run?

app/models/commit_checks.rb
davidcornu: Seems odd to put the success case after all the error cases as a `rescue/else`.

davidcornu:
re: rubocop

I prefer having a trailing space after required keyword arguments but that might just be me

```ruby
def my_method(required_arg: )
                           ^
```

davidcornu:
Some minor concerns around lazy-loading the command output but overall LGTM.

gmalette:
LGTM. I didn't :tophat: but I saw this was using `OutputStream` which means it'll have colorization, right?

byroot:
> which means it'll have colorization, right?

Indeed.

byroot:
@davidcornu 5a7b782 refactored `CommitChecks`. I think you'll like it better.

I'll merge on green CI.

One last thing, as of now these checks only execute for deploys, meaning not for rollbacks. Not sure if they should as you don't want to delay rollbacks too much. But if you think they should let me know I can add them.

dalehamel:
:heart: :sparkles: :heart: 

:clap: 

:gift: 

:tada:

app/models/commit_checks.rb
davidcornu: I still don't like this

davidcornu:
> @davidcornu 5a7b782 refactored `CommitChecks`. I think you'll like it better.

:+1:

app/models/commit_checks.rb
byroot: What the `else`? It's better than putting it inside the main block, that way this statement can't be rescued by the block.

davidcornu: > that way this statement can't be rescued by the block.

What does that add though? It's just an `attr_accessor`, if that breaks I'd be a lot more worried.

byroot: > It's just an `attr_accessor`

It's not :wink:

davidcornu: Ah. My bad

Issue #430, opened 2015-06-17
Add support for Github Entreprise


mathieugagne:
Hey guys

The company I work at has a private Github Entreprise. How can I point Shipit to a different Github instance?

Thanks

byroot:
Of course. You just have to set `github_api.api_endpoint` in the `secrets.yml`.

I'll let this open because it means we should update the documentation.

mathieugagne:
I'm still being redirected to github.com for authentication.

Seems omniauth-github is not set to take extra parameters needed for Github Entreprise.

From https://github.com/intridea/omniauth-github#github-enterprise-usage
```
provider :github, ENV['GITHUB_KEY'], ENV['GITHUB_SECRET'],
    {
      :client_options => {
        :site => 'https://github.YOURDOMAIN.com/api/v3',
        :authorize_url => 'https://github.YOURDOMAIN.com/login/oauth/authorize',
        :token_url => 'https://github.YOURDOMAIN.com/login/oauth/access_token',
      }
    }
```

Current implementation in [lib/shipit/engine:25](https://github.com/Shopify/shipit-engine/blob/f14cfc401273df52f2e57e96c0e0fc9a67dd8868/lib/shipit/engine.rb#L25)
```
app.middleware.use OmniAuth::Builder do
  provider(:github, Shipit.github_oauth_id, Shipit.github_oauth_secret, scope: 'email')
end
```

Seems we need an additional parameter for login or a way to grab the login url, e.g.
```yml
github_api:
  api_endpoint: 'https://github.YOURDOMAIN.com/api/v3/'
  login_endpoint: 'https://github.YOURDOMAIN.com/'
```
Other implementation could use /api/v3 by default, or have a api_version 'v3' param instead.

mathieugagne:
Tested on a fork and worked with hardcoded `:site`, `:authorize_url` and `:token_url`

byroot:
Yeah we need to add support for this in the `secrets.yml`. Also the links to gtihub will be broken.

I'm fixing this right away.

mathieugagne:
Tracking a few 'github.com' mentions:
https://github.com/Shopify/shipit-engine/blob/master/app/views/stacks/new.html.erb#L11

[stacks/_header](https://github.com/Shopify/shipit-engine/blob/master/app/views/stacks/_header.html.erb#L28) link points to a github.com repository

byroot:
@mathieugagne there is also the git url:

```ruby
  def repo_git_url
[...]

PR #431, opened 2015-06-17
Fix documentation for GitHub enterprise

byroot:
Fixes: https://github.com/Shopify/shipit-engine/issues/430

@mathieugagne is this  good for you?

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,7 +283,7 @@ development:
 
 If you specify an `access_token`, you don't need a `login` and `password`. The opposite is also true:  if you specify a `login` and `password`, then you don't need an `access_token`.
 
-The `api_endpoint` is optional, and required only if you want to use a proxy.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -283,7 +283,7 @@ development:
 
 If you specify an `access_token`, you don't need a `login` and `password`. The opposite is also true:  if you specify a `login` and `password`, then you don't need an `access_token`.
 
-The `api_endpoint` is optional, and required only if you want to use a proxy.
```
gmalette: `If you use GitHub Enterprise or a proxy` ?

byroot: I don't think the proxy use case is common enough to be worth a special case.

gmalette: I guess, yeah, it's a lot of trouble

mathieugagne:
:ok_hand: I looked at the doc but I guessed I must have skipped this line. Thanks for the prompt response guys.

PR #437, opened 2015-06-17
Fix: Non Github Enterprise omniauth issue

pjaspers:
If using a non-Github Enterprise, omniauth was expecting a hash to be set
in its `client_options`, instead it got nil.

This fixes that.

Otherwise it would throw an error, in the `deep_symbolize` inside omniauth.


byroot:
Thanks for this!

PR #438, opened 2015-06-17
Add support for OSX.

pjaspers:
OSX is nitpicky when it comes to `sed`. It would throw an error when running the command.

[0] http://stackoverflow.com/questions/16745988/sed-command-works-fine-on-ubuntu-but-not-mac


rafaelfranca:
Thank you

byroot:
Thanks!

Issue #425, opened 2015-06-09
use "committed" instead of "merged" when it's not a merge


dwradcliffe:
For commit items that aren't merge commits, it would be nice if the UI displayed "committed" instead of "merged".

![shipit - rubygems org 2015-06-08 23-23-33](https://cloud.githubusercontent.com/assets/118850/8050086/709bbe92-0e35-11e5-99a7-72c4e8f873a0.png)

byroot:
Good point. We kinda miss the info though, expect if we consider that everything that isn't a PR merge commit is a regular commit.

pushrax:
How about "pushed"?

gmalette:
"" may also work. the word seems superfluous

dwradcliffe:
Yeah I think we could just remove the word

PR #440, opened 2015-06-23
Remove "merged" from commit details view

girasquid:
@byroot @pushrax 

Pretty sure this is the only change you need for this, right?

Fixes https://github.com/Shopify/shipit-engine/issues/425.


dwradcliffe:
Thanks @girasquid :smiley:

byroot:
It won't be consistent with deploy / rollbacks anymore, but yeah ok let's try that.

girasquid:
What do I need to change for those?

byroot:
You shouldn't change those. Otherwise there would be no way to make a distinction between a deploy and a rollback.

pushrax:
:+1:

PR #442, opened 2015-06-29
Prevent fetch deployed revision during rollbacks

byroot:
cc @eapache since IIRC you reported this bug a while ago.

I finally figured it out. We have this STI hierarchy: `Task -> Deploy -> Rollback`.

But `has_many :deploys` was selecting only the `Deploy` instances.

@rafaelfranca you may know a better way to express this?

### DIFF
--- a/app/models/stack.rb
+++ b/app/models/stack.rb
@@ -7,6 +7,7 @@ class Stack < ActiveRecord::Base
   has_many :tasks, dependent: :destroy
   has_many :deploys
   has_many :rollbacks
+  has_many :deploys_and_rollbacks, -> { where(type: %w(Deploy Rollback)) }, class_name: 'Task'
   has_many :github_hooks, dependent: :destroy, class_name: 'GithubHook::Repo'
   has_many :hooks, dependent: :destroy
   has_many :api_clients, dependent: :destroy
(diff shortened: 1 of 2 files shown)


davidcornu:
:+1:

rafaelfranca:
> @rafaelfranca you may know a better way to express this?

Not that I know

byroot:
> Not that I know

Ok, no problem. I was just surprised by this:

```ruby
>> stack.deploys.count
   (0.2ms)  SELECT COUNT(*) FROM "tasks" WHERE "tasks"."type" IN ('Deploy') AND "tasks"."stack_id" = ?  [["stack_id", 1]]
=> 3
>> Deploy.where(stack_id: stack.id).count
   (0.3ms)  SELECT COUNT(*) FROM "tasks" WHERE "tasks"."type" IN ('Deploy', 'Rollback') AND "tasks"."stack_id" = ?  [["stack_id", 1]]
=> 4
```

gmalette:
@byroot that's the reason I had `require_dependency` somewhere for those. For production it *should* be fine since the classes were loaded

byroot:
@gmalette no it's not a loading issue. `has_many :deploys` won't include Deploy descendants.

eapache:
belated :+1:

PR #443, opened 2015-06-30
Mark stacks as inaccessible if GitHub returns 404

byroot:
Fix: #318 

This isn't really the final solution, but it's more like a first step. There might be other places to handle 404, and we could leverage this column to bail out from some calls. I'll keep and eyes on exceptions and followups as needed.

### DIFF
--- a/app/models/stack.rb
+++ b/app/models/stack.rb
@@ -244,6 +244,14 @@ class Stack < ActiveRecord::Base
     Rails.cache.write(ci_enabled_cache_key, true)
   end
 
+  def mark_as_inaccessible!
+    update!(inaccessible_since: Time.now) unless inaccessible?
+  end
+
+  def inaccessible?
+    inaccessible_since?
+  end
+
   private
 
   def clear_cache
--- a/app/views/stacks/show.html.erb
+++ b/app/views/stacks/show.html.erb
@@ -17,6 +17,21 @@
   </div>
 <% end %>
 
+<% if @stack.inaccessible? %>
+  <div class="banner banner--red">
+    <div class="banner__inner wrapper">
+      <div class="banner__content">
+        <h2 class="banner__title">This repository seems to be missing!</h2>
+        <p class="banner__text">
+          Since <%= time_ago_in_words(@stack.inaccessible_since) %>
+          <%= link_to @stack.github_repo_name, github_repo_url(@stack.repo_owner, @stack.repo_name) %> is inaccessible.
+          This could be a permission issue, or the repository have been deleted on GitHub.
+        </p>
+      </div>
+    </div>
+  </div>
+<% end %>
+
 <% if @stack.locked? %>
   <div class="banner banner--orange">
     <div class="banner__inner wrapper">
--- a/test/dummy/db/schema.rb
+++ b/test/dummy/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20150518214944) do
+ActiveRecord::Schema.define(version: 20150630171127) do
 
   create_table "api_clients", force: :cascade do |t|
     t.text     "permissions", limit: 65535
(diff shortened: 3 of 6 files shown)


gmalette:
nice

app/models/stack.rb
```diff
@@ -244,6 +244,14 @@ def enable_ci!
     Rails.cache.write(ci_enabled_cache_key, true)
   end
 
+  def mark_as_inaccessible!
+    update!(inaccessible_since: Time.now) unless inaccessible?
+  end
+
+  def inaccessible?
+    inaccessible_since?
```
davidcornu: Isn't `inaccessible_since` a datetime? Does Rails generate `?` methods for all attributes?

test/dummy/db/schema.rb
```diff
@@ -39,10 +39,10 @@
     t.integer  "deletions",    limit: 4
   end
 
-  add_index "commits", ["author_id"], name: "index_commits_on_author_id", using: :btree
-  add_index "commits", ["committer_id"], name: "index_commits_on_committer_id", using: :btree
-  add_index "commits", ["created_at"], name: "index_commits_on_created_at", using: :btree
-  add_index "commits", ["stack_id"], name: "index_commits_on_stack_id", using: :btree
```
davidcornu: Odd. Did we bump Rails?

test/jobs/github_sync_job_test.rb
```diff
@@ -48,4 +48,11 @@ class GithubSyncJobTest < ActiveSupport::TestCase
     assert_equal first, parent
     assert_equal [last], commits
   end
+
+  test "if GitHub returns a 404, the stacks is marked as unaccessible" do
```
davidcornu: `inaccessible`

app/models/stack.rb
byroot: Yup to both.

davidcornu: TIL.

:+1:

app/views/stacks/show.html.erb
```diff
@@ -17,6 +17,21 @@
...
+        <h2 class="banner__title">This repository seems to be missing!</h2>
+        <p class="banner__text">
+          Since <%= time_ago_in_words(@stack.inaccessible_since) %>
+          <%= link_to @stack.github_repo_name, github_repo_url(@stack.repo_owner, @stack.repo_name) %> is inaccessible.
+          This could be a permission issue, or the repository have been deleted on GitHub.
+        </p>
+      </div>
+    </div>
+  </div>
+<% end %>
```
davidcornu: Screenshot?

davidcornu:
Some minor stuff, otherwise LGTM

app/views/stacks/show.html.erb
byroot: capture d ecran 2015-06-30 a 14 03 07

test/dummy/db/schema.rb
byroot: Arf. No it's because I use sqlite in dev. I don't think it's a big deal honestly.

test/dummy/db/schema.rb
```diff
@@ -151,14 +152,14 @@
     t.integer  "additions",       limit: 4,     default: 0
     t.integer  "deletions",       limit: 4,     default: 0
     t.text     "definition",      limit: 65535
-    t.binary   "gzip_output",     limit: 16777215
```
byroot: Actually this is pretty bad.

byroot:
@davidcornu I fixed the schema.rb. Is the screenshot good looking for ya?

app/views/stacks/show.html.erb
davidcornu: > **repo** has been inaccessible for 8 days. This could either be due to a permission issue or because the repository has been removed from GitHub.

Might be better copy

davidcornu:
@byroot screenshot looks good. I suggested a minor copy tweak.

app/views/stacks/show.html.erb
byroot: Indeed.

PR #444, opened 2015-06-30
Update the cached deploy spec on every pushes

byroot:
As discussed offline, the current situation where you need to deploy to update the cached deploy spec (the one that defines custom tasks, rollback ability, review elements, hidden ci status, etc), is just a relic of previous technical limitations. It has no reasons to still exist.

This PR update the cache everytime a new commit is received.


davidcornu:
LGTM

gmalette:
:+1:

Issue #363, opened 2015-05-20
Finish the API.


byroot:
Out of the top of my head we're missing:
- [ ] Rollback
- [ ] Abort task
- [ ] List custom tasks
- [x] Trigger custom tasks (https://github.com/Shopify/shipit-engine/pull/447).

teu:
+1 on the abort and rollback

byroot:
You need to abort deploys and rollback them from the REST API? If so I'd love to hear about your use case.

PR #447, opened 2015-07-02
Implement an API to trigger custom tasks

byroot:
Part of: https://github.com/Shopify/shipit-engine/issues/363

Requested by @dwradcliffe 

Endpoint `POST /api/stacks/:organization/:repository/:environment/task/:task_name`

 e.g: `POST /api/shopify/shipit/production/task/restart`.

@gmalette @davidcornu @dwradcliffe for review please.

### DIFF
--- a/app/controllers/api/tasks_controller.rb
+++ b/app/controllers/api/tasks_controller.rb
@@ -1,6 +1,7 @@
 module Api
   class TasksController < BaseController
     require_permission :read, :stack
+    require_permission :deploy, :stack, only: :trigger
 
     def index
       render_resources stack.tasks
@@ -8,6 +9,10 @@ module Api
 
     def show
       render_resource stack.tasks.find(params[:id])
+    end
+
+    def trigger
+      render_resource stack.trigger_task(params[:task_name], current_user), status: :accepted
     end
   end
 end
--- a/test/controllers/api/tasks_controller_test.rb
+++ b/test/controllers/api/tasks_controller_test.rb
@@ -13,4 +13,24 @@ class Api::TasksControllerTest < ActionController::TestCase
     assert_response :ok
     assert_json '0.id', task.id
   end
+
+  test "#show returns a task" do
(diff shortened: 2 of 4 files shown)


test/controllers/api/tasks_controller_test.rb
```diff
@@ -13,4 +13,24 @@ class Api::TasksControllerTest < ActionController::TestCase
...
+  end
+
+  test "#trigger triggers a custom task" do
+    post :trigger, stack_id: @stack.to_param, task_name: 'restart'
+    assert_response :accepted
+    assert_json 'type', 'task'
+    assert_json 'status', 'pending'
+  end
+
+  test "#trigger when the task doesn't exist" do
```
davidcornu: > #trigger returns a 404 when the task doesn't exist

app/controllers/api/tasks_controller.rb
```diff
@@ -9,5 +9,9 @@ def index
     def show
       render_resource stack.tasks.find(params[:id])
     end
+
+    def trigger
+      render_resource stack.trigger_task(params[:task_name], current_user), status: :accepted
+    end
```
davidcornu: We might want to update the `require_permission` call for `TasksController` as `trigger` is more then a `:read` on `:stack`.

davidcornu:
Aside from the permissions issue LGTM

app/controllers/api/tasks_controller.rb
byroot: :s very good point /o\

byroot:
I've restricted it to `deploy:stack`. Since custom tasks can possibly do anything, I think deploying, rolling back and running a custom task should be the same permission.

gmalette:
:+1:

davidcornu:
> I've restricted it to deploy:stack. Since custom tasks can possibly do anything, I think deploying, rolling back and running a custom task should be the same permission.

:+1:

PR #451, opened 2015-07-08
Implement an "Abort and Rollback" button on deploy log

### LINKED ISSUE #449: Abort and Rollback button
### Context:

Shipit have been built around the idea that the deploy scripts are transactional, meaning that if you abort the script, it will go back to it's previous step like capistrano 2 used to do.

Unfortunately in real life not all deploy scripts are transactional.

### Proposal

If we acknowledge this fact, then it makes sense to have an `Abort and rollback` button on the deploy log page. It would kill the current deploy, trigger a rollback, and redirect the user to the rollback log.

The benefit would be a faster reaction in case of deploy related outages.

@gmalette thoughts?

cc @KnVerey

byroot:
Fix #449 

### Workflow

  - When a deploy is running and the stack supports rollbacks, the "Abort and Rollback Deploy" button is visible.
  - Clicking it trigger the abort just like the old button, but also flip a flag
  - Once the deploy successfuly aborted, a rollback is triggered
  - The output polling endpoint then include a link to the newly triggered rollback
  - The user is redirected to the rollback's output

### What it looks like: 

The button in the deploy header:
capture d ecran 2015-07-08 a 13 56 10

The 2 new states in the deploys list (the first one is `aborting`, the other is `aborted`):
capture d ecran 2015-07-08 a 13 47 15

@gmalette @davidcornu @rafaelfranca for review please. As well as @KnVerey for the UX part.

I did my best in term of design, but it could be improved I guess...

### DIFF
--- a/app/assets/javascripts/shipit/deploy.js.coffee
+++ b/app/assets/javascripts/shipit/deploy.js.coffee
@@ -1,7 +1,37 @@
-$(document).on 'click', '.deploy-abort .deploy-action', (event) ->
-  event.preventDefault()
-  $target = $(event.currentTarget)
-  return if $target.hasClass('pending')
-  $target.addClass('pending')
-  enable = -> $target.removeClass('pending')
-  $.post($target.attr('href')).done(-> setTimeout(enable, 3000))
+class AbortButton
+  SELECTOR = '[data-action="abort"]'
+
+  @listen: ->
+    $(document).on('click', SELECTOR, @handle)
+
+  @handle: (event) =>
+    event.preventDefault()
+    button = new this($(event.currentTarget))
+    button.trigger()
+
+  constructor: (@$button) ->
+    @url = @$button.find('a[href]').attr('href')
+    @shouldRollback = @$button.data('rollback')
+
+  trigger: ->
+    return false if @isDisabled()
+
+    @disable()
+    @waitForCompletion()
+    $.post(@url).success(@waitForCompletion).error(@reenable)
+
+  waitForCompletion: =>
+    setTimeout(@reenable, 3000)
+
+  reenable: =>
+    @$button.removeClass('pending')
+    @$button.siblings(SELECTOR).removeClass('disabled')
+
+  disable: ->
+    @$button.addClass('pending')
+    @$button.siblings(SELECTOR).addClass('disabled')
+
+  isDisabled: ->
+    @$button.hasClass('pending') || @$button.hasClass('disabled')
+
+AbortButton.listen()
--- a/app/assets/images/aborted.svg
+++ b/app/assets/images/aborted.svg
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg width="16px" height="16px" viewBox="0 0 16 16" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <defs></defs>
+    <g stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g transform="translate(1.000000, 1.000000)" stroke-width="2" stroke="#EDAA4E">
+            <path d="M0,0 L14,14"></path>
+            <path d="M14,0 L0,14"></path>
+        </g>
+    </g>
+</svg>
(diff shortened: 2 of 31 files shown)


app/assets/javascripts/shipit/deploy.js.coffee
```diff
@@ -1,7 +1,32 @@
...
-  return if $target.hasClass('pending')
-  $target.addClass('pending')
-  enable = -> $target.removeClass('pending')
-  $.post($target.attr('href')).done(-> setTimeout(enable, 3000))
+class AbortButton
+  SELECTOR = '[data-action="abort"]'
+
+  @handle: (event) =>
+    event.preventDefault()
+    $element = $(event.currentTarget).closest(SELECTOR)
```
davidcornu: Is the `.closest` necessary given that you're binding to a click event on `[data-action="abort"]`?

app/assets/javascripts/shipit/deploy.js.coffee
```diff
@@ -1,7 +1,32 @@
...
+  reenable: =>
+    @$button.removeClass('pending')
+
+  disable: ->
+    @$button.addClass('pending')
+
+  isDisabled: ->
+    @$button.hasClass('pending')
+
+$(document).on('click', '[data-action=abort]', AbortButton.handle)
```
davidcornu: Seems odd to me to build an `AbortButton` by passing `AbortButton.handle` as a callback. IMHO having something like `AbortButton.listen()` handle the binding would be cleaner.

Also, shouldn't this be `$(document).on('click', AbortButton.SELECTOR, AbortButton.handle)`?

KnVerey:
Yay! :clap: 

A few questions:
- after you click one of the new buttons, what happens to them? I'm mostly thinking of making sure it is clear after the fact which one was clicked, and not possible to click both.
- is the case where capistrano needs to be unlocked, like what I hit yesterday, handled?
- are there tooltips on those states in the deploys list? I don't think it is super clear what those icons mean, but I'm no designer myself and don't really have any better ideas. :-/

davidcornu:
Those buttons look like they're floating about IMHO. Not entirely sure where they should live though.

byroot:
> after you click one of the new buttons, what happens to them?

The one you clicked will be disabled until the server answers, but yeah I need to improve the UX there.

> is the case where capistrano needs to be unlocked, like what I hit yesterday, handled?

That is the responsibility of the deployed app. I will handle that in capistrano.

> are there tooltips on those states in the deploys list?

Yes there is tooltips.

byroot:
I updated the JS in regards to @davidcornu's comments and it also now hide the other button during the recovery period to avoid confusion.

davidcornu:
> I updated the JS in regards to @davidcornu's comments 

:+1: 

> and it also now hide the other button during the recovery period to avoid confusion.

Should we disable rather than hiding? It's a bit weird to have something disappear all of a sudden.

davidcornu:
@byroot any design :eyes: we can summon?

byroot:
Well, the whole header is a bit screwy. Maybe @jeromegn could lend us a hand here :)

byroot:
So I managed to improve the buttons look myself. It's not super great but that will do.

I'll merge after lunch unless you guys have concerns?

KnVerey:
Not a big deal, but in the absence of designer input, I agree with @davidcornu that disabling the non-clicked button would be a nicer experience than having it disappear, especially if we have some sort of loading indicator on the clicked button so it's still clear which one was clicked. I also just had the idea that it might be useful to have one of the buttons look "primary" to help someone underinformed quickly make the right decision about which to click in most cases.

byroot:
capture d ecran 2015-07-09 a 11 56 52

Here what it looks like now after you clicked the button.

@KnVerey @davidcornu good enough for you?

KnVerey:
> good enough for you?

Yep :+1:

KnVerey:
Oh wait, maybe it should say "Aborting with rollback..." or something to distinguish it from the other button.

byroot:
> it should say "Aborting with rollback..."

Done.

KnVerey:
:tada: thanks so much for this @byroot !

davidcornu:
> I also just had the idea that it might be useful to have one of the buttons look "primary" to help someone underinformed quickly make the right decision about which to click in most cases.

Agreed. Might be worth addressing this if there's ever a header cleanup.

PR #453, opened 2015-07-09
Change labels for abort buttons

volmer:
Now that we have 2 buttons to abort deploys I find the labels quite confusing, so I suggest changing "Abort Deploy" and "Abort Deploy and Rolback" to simply "Abort" and "Abort and Rollback", respectively.

### DIFF
--- a/app/views/tasks/_task_output.html.erb
+++ b/app/views/tasks/_task_output.html.erb
@@ -16,7 +16,7 @@
 
       <span class="deploy-abort first" data-action="abort" data-status="<%= task.status %>">
         <%= link_to abort_stack_task_path(@stack, task), class: 'btn btn--delete deploy-action' do %>
-          <span class="caption--ready">Abort Deploy</span>
+          <span class="caption--ready">Abort</span>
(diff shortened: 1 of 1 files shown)


davidcornu:
@KnVerey also [suggested](https://github.com/Shopify/shipit-engine/pull/451#issuecomment-120002686) visually making one of them the primary action and the other a secondary one.

gmalette:
Both make sense. I think if both are present, `Abort` can be lighter in colour or something

byroot:
:+1:

KnVerey:
While deploying just now, I noticed that the two buttons jump around depending on the length of the "2 minutes ago" phrase. Sometimes the second button falls to the next line, and sometimes it gets cut off, like so:


Can we fix this too?

byroot:
> Can we fix this too?

I did the best I could in term of CSS, but a PR is more than welcome.

KnVerey:
I included this change in https://github.com/Shopify/shipit-engine/pull/455 :smile:

PR #455, opened 2015-07-11
Improve abort button visuals

KnVerey:
@byroot @volmer @davidcornu @gmalette 
I'm not very good with css, but here's my best effort at resolving my own complaints!
Makes some visual tweaks to the deploy banner.


app/views/tasks/_task_output.html.erb
```diff
@@ -7,29 +7,27 @@
...
           <span class="caption--pending">Aborting...</span>
         <% end %>
-      </span>
 
-      <% if task.supports_rollback? %>
-        <span class="deploy-abort" data-action="abort" data-rollback="true" data-status="<%= task.status %>">
-          <%= link_to abort_stack_task_path(@stack, task, rollback: true), class: 'btn btn--delete deploy-action' do %>
-            <span class="caption--ready">Abort Deploy and Rollback</span>
+        <% if task.supports_rollback? %>
+          <%= link_to abort_stack_task_path(@stack, task, rollback: true), class: "btn btn--delete action-button", "data-action" => "abort", "data-rollback" => "true", "data-status" => "#{task.status}" do %>
```
byroot: For concision you can pass `data-*` parameters as a sub Hash:

```ruby
link_to ...., data: {action: 'abort', rollback: true, etc}
``` 

Also about ` "#{task.status}"`, interpolation isn't needed here, you can just pass `task.status` directly.

byroot:
One minor code style thingy, but otherwise LGTM. I also :tophat:ed and it's way nicer.

Thanks a lot for this!

KnVerey:
Made the suggested style changes in https://github.com/Shopify/shipit-engine/commit/ae6b1bc66845528f4c88107bf98f108bd9a2f56d :smile:

PR #459, opened 2015-07-17
Fix rails version in gemspec

cmckni3:
Changes the rails version so 4.2.0 and 4.2.1 can be used

### DIFF
--- a/shipit-engine.gemspec
+++ b/shipit-engine.gemspec
@@ -16,7 +16,7 @@ Gem::Specification.new do |s|
   s.files = Dir["{app,config,db,lib}/**/*", "MIT-LICENSE", "Rakefile", "README.rdoc"]
   s.test_files = Dir["test/**/*"]
 
-  s.add_dependency 'rails', '~> 4.2.2'
+  s.add_dependency 'rails', '~> 4.2.0'
   s.add_dependency 'securecompare', '~> 1.0.0'
   s.add_dependency 'validate_url', '~> 1.0.0'
   s.add_dependency 'explicit-parameters', '~> 0.0.3'


cmckni3:
Are any of the Rails 4.2 features being used? I mainly ask to see if the version can be relaxed to 4.x.

shipit-engine.gemspec
```diff
@@ -16,7 +16,7 @@ Gem::Specification.new do |s|
   s.files = Dir["{app,config,db,lib}/**/*", "MIT-LICENSE", "Rakefile", "README.rdoc"]
   s.test_files = Dir["test/**/*"]
 
-  s.add_dependency 'rails', '~> 4.2.2'
+  s.add_dependency 'rails', '~> 4.2.0'
```
byroot: Since we use the pessimistic version comparator, you can already use any of the `4.2.x releases of rails.`

byroot:
> Are any of the Rails 4.2 features being used?

Yes, we use active job among other 4.2 features.

>  I want to see if the Rails version can be relaxed to 4.x.

May I ask why you'd want to do this?

shipit-engine.gemspec
cmckni3: I ran into a conflict because I ran `rails new` with Rails 4.2.1

rafaelfranca: If it is worth noting, 4.2.1 is vulnerable to some security issues and you
should not be using. Even more in a new application.

byroot: > If it is worth noting, 4.2.1 is vulnerable to some security issues

Yeah I guess we can merge this anyway.

rafaelfranca: Yeah, definitely not this gem job to make sure people don't use vulnerable
versions :+1:

cmckni3:
>    Yes, we use active job among other 4.2 features.

Nevermind about relaxing the Rails version in that case

shipit-engine.gemspec
cmckni3: > If it is worth noting, 4.2.1 is vulnerable to some security issues

Yeah. I adjusted my `Gemfile` to use the pessimistic operator for `4.2.x` and ran `bundle update 'rails'` update my `Gemfile.lock` to get Rails 4.2.3.

> Yeah, definitely not this gem job to make sure people don't use vulnerable
versions :+1:

I agree

byroot: > I ran into a conflict because I ran rails new with Rails 4.2.1

Oh so you didn't use the template? https://github.com/Shopify/shipit-engine/blob/master/template.rb

cmckni3: I used the template but I had Rails 4.2.1 already installed from another project, so, `rails new` put `gem '4.2.1'` into my `Gemfile`

byroot: That's good to know thanks. I'll see if we can improve that.

PR #464, opened 2015-07-22
Use ActiveJob for BackgroundJob::Unique lock key serialization

etiennebarrie:
[`ActiveJob::Arguments.serialize`](http://api.rubyonrails.org/v4.2.3/classes/ActiveJob/Arguments.html#method-i-serialize) can take care of serializing using `to_global_id` when possible for us.
~~No tests, sorry~~ Added a test

It changes slightly the lock key:
```patch
-"ChunkRollupJob-gid://shipit/Task/783434608"
+"ChunkRollupJob-{\"_aj_globalid\"=>\"gid://shipit/Task/783434608\"}"
```

### DIFF
--- a/app/jobs/background_job/unique.rb
+++ b/app/jobs/background_job/unique.rb
@@ -20,14 +20,7 @@ class BackgroundJob
     end
 
     def lock_key(*args)
-      ([self.class.name] + args).map { |arg| hash_argument(arg) }.join('-')
-    end
-
-    private
-
-    def hash_argument(argument)
-      return argument.to_global_id.to_s if argument.respond_to?(:to_global_id)
-      argument.to_s
+      ActiveJob::Arguments.serialize([self.class.name] + args).join('-')
     end
   end
 end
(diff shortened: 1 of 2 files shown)


gmalette:
We should do this early and merge it when no one is using Shipit because the locks will not be respected.

byroot:
To be honest I'm not sure which I prefer. This PR make the code simpler and nicer, but the lock key is rather ugly...

gmalette:
I don't think we should be optimizing for what the key looks like

byroot:
You have a point :) :+1: 

> We should do this early and merge it when no one is using Shipit because the locks will not be respected.

I can't recall of any lock that would cause a major issue if it was lost. Some jobs would fail, but the application would come back a a proper stack rather quickly.

gmalette:
:+1:

davidcornu:
> I can't recall of any lock that would cause a major issue if it was lost. Some jobs would fail, but the application would come back a a proper stack rather quickly.

Can we keep using the old key for existing jobs and do a progressive migration? Also, bear in mind there are other users of Shipit that could be affected by this

byroot:
Yeah but Im not sure if it's worth providing no downtime upgrades. We don't even do those ourselves.

PR #467, opened 2015-08-05
Fix nil significant-status

gmalette:
This fixes the bug we had today where the commit had status but none significant, that caused Shipit to go down.

@davidcornu @camilo

### DIFF
--- a/app/models/unknown_status.rb
+++ b/app/models/unknown_status.rb
@@ -1,10 +1,4 @@
-class UnknownStatus
-  attr_reader :commit
-
-  def initialize(commit)
-    @commit = commit
-  end
-
+UnknownStatus = Struct.new(:commit) do
   def state
     'unknown'
   end
--- a/test/models/stacks_test.rb
+++ b/test/models/stacks_test.rb
@@ -293,6 +293,26 @@ class StacksTest < ActiveSupport::TestCase
     end
   end
 
+  test "#filter_visible_statuses removes statuses from hidden contexts" do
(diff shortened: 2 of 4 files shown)


test/models/stacks_test.rb
```diff
@@ -293,6 +293,26 @@ def setup
...
+
+    assert_equal [commit1, commit2], stack.filter_visible_statuses([hidden, commit1, commit2])
+  end
+
+  test "#filter_meaningful_statuses removes statuses from soft-failing contexts" do
+    stack = stacks(:cyclimse)
+    stack.stubs(soft_failing_statuses: ['ci/soft-fail'])
+    commit1 = Status.new(state: 'pending', context: 'ci/valid')
+    commit2 = Status.new(state: 'pending', context: 'ci/valid')
+    soft_fail= Status.new(state: 'pending', context: 'ci/soft-fail')
```
rafaelfranca: missing space here before the `=`

rafaelfranca:
:+1:

app/models/unknown_status.rb
```diff
@@ -1,10 +1,6 @@
-class UnknownStatus
+UnknownStatus = Struct.new(:commit) do
   attr_reader :commit
```
davidcornu: Is this necessary for a `Struct`?

gmalette: oopsie

davidcornu:
:shipit:

PR #469, opened 2015-08-13
Bootstrap integration WIP

jeromegn:
Yesterday, @byroot and I did the initial work to replace the current HTML / CSS with Bootstrap conventions and styles.

The goal of this is to reduce design-related blockers by enabling anybody to create sufficiently well-styled pages and features without much effort.

To keep the same(ish) design for ShipIt, we should build a theme on top of Bootstrap, overriding some of its variables and styles.

@byroot asked me to PR this even if it's not quite ready yet. Right now the only page affected by the changes is the "Create Stack" page.

 This is what it looks like with the code of this PR:

image

Since we have to rewrite most of the HTML, we thought we'd switch to a saner way to write it by using [Slim](http://slim-lang.com/) (obviously, this is a matter of opinion.)

This only took an hour's work.

This PR should not affect the styling of other sections as everything is contained in a new layout and new stylesheets.


byroot:
@gmalette @rafaelfranca @davidcornu you guys have any thoughts on this?

and thanks again @jeromegn!

app/controllers/stacks_controller.rb
```diff
@@ -1,8 +1,10 @@
 class StacksController < ShipitController
+
```
byroot: Rubocop doesn't like this empty line.

rafaelfranca:
I have some concerns about adding slim as dependency since it doesn't buy us too much, otherwise this look good. Having a know style convention would make easier to contributors to change style.

arthurnn:
http://primercss.io/ ?

byroot:
> I have some concerns about adding slim as dependency since it doesn't buy us too much

I always regretted not to start this project with Slim / HAML. So since this endeavour implies to rewrite all views, I though it was the occasion to migrate.

I know not everybody likes those alternative template engines, but here's my arguments for it:

  - It's more concise, more DRY if you will.
  - It prevent you from writing invalid HTML (missing closing tag, etc), see how far we went to validate ERB templates in Shopify: https://github.com/Shopify/erb-mux (sorry private project).
  - It requires you to have perfect indentation. If you go through the views of Shipit right now, you'll see that it really lacks consistency.

But I don't want to force it, so really unless the majority agrees, we can go back on ERB no problem.

> http://primercss.io/

Oh, I never heard of this. Initially the plan was to replicate the current style, but given how much Shipit is GitHub centric, it could make sense to use this to achieve consistency with GitHub UI.

davidcornu:
> The goal of this is to reduce design-related blockers by enabling anybody to create sufficiently well-styled pages and features without much effort.

:+1: Having everyone just follow bootstrap conventions would make contributing anything UI-related a lot easier (and we can achieve a decent UI before involving designers)

> I have some concerns about adding slim as dependency since it doesn't buy us too much

I do too. Unless we're going to make the effort to port everything in a reasonable time frame, I'd like to avoid having a split code base.

gmalette:
> it could make sense to use this to achieve consistency with GitHub UI

I kind of like the current UI...

byroot:
#566 Is superseeding this

Issue #181, opened 2014-07-18
Allow deployment of branch


sirupsen:
It'd be really nice if you could deploy branches, Capistrano already deploys this with SHA and STRICT_DEPLOYS. Awesome strategy for deploying risky branches would be something like:
1. Deploy `risky_branch` from my branch after rebasing with latest deployed SHA.
2. See that it works.
3. Deploy latest deployed sha again.
4. Merge to master, where it will be part of a future deploy.

pjaspers:
Is there more info on allowing inputs?

byroot:
@pjaspers we do have on our internal roadmap the plan to support deploying pull request / branches like GitHub does: http://githubengineering.com/deploying-branches-to-github-com/

But right now we only support the "merge in master branch" model.

pjaspers:
That's exactly what I was planning to use it for :)

byroot:
We'll ping you back when we have it then :)

pjaspers:
For now the "merge in master" will do, but quickly deploying a branch has become a part of our workflow, so it will be nice when it's there.

simi:
Are there still any plans on this feature?

byroot:
It's more or less what Review Stacks provide: https://github.com/Shopify/shipit-engine/blob/master/docs/review_stacks.md, it's not super well documented though, but they allow to deploy pull requests to staging environments or similar.

simi:
@byroot thanks for the info.

benlangfeld:
> It's more or less what Review Stacks provide: https://github.com/Shopify/shipit-engine/blob/master/docs/review_stacks.md, it's not super well documented though, but they allow to deploy pull requests to staging environments or similar.

If anyone needs clarification of the documentation, please let me know and I'll be happy to provide it.

simi:
It would be great to provide complete example using `krane` (if that's the way to go with Kubernetes today). I'm not sure what to do in `up`/`down` blocks. Just run some shell scripts?

benlangfeld:
> It would be great to provide complete example using `krane` (if that's the way to go with Kubernetes today). I'm not sure what to do in `up`/`down` blocks. Just run some shell scripts?

In our case, we have a lot of complex logic around cluster allocation and namespace creation, including copying resources from a template. This part hasn't been included in Krane because it is too specific and would not be accepted. I can put together a blog post about how we do it to serve as an example, though.

simi:
[...]

PR #471, opened 2015-08-14
Define optional environment variables in shipit.yml

byroot:
The idea is to be able to toggle flags and / or adjust settings in the deploy script from the Shipit UI.

e.g.

``` yml
deploy:
  variables:
    - name: RAILS42_PERCENTAGE
      title:  Percentage of servers running Rails 4.2
      default: 0
```

Will produce on the deploy page:

capture d ecran 2015-08-14 a 17 29 04

@rafaelfranca @davidcornu @gmalette 

Possible followups:
- Do the same for tasks
- Allow to persist the value that was set for the following deploys
- Value validation / sanitization e.g. (`type: number`)

### DIFF
--- a/app/models/deploy_spec.rb
+++ b/app/models/deploy_spec.rb
@@ -61,6 +61,10 @@ class DeploySpec
     deploy_steps || cant_detect!(:deploy)
   end
 
+  def deploy_variables
+    Array.wrap(config('deploy', 'variables')).map(&VariableDefinition.method(:new))
+  end
+
   def rollback_steps
     config('rollback', 'override') || discover_rollback_steps
   end
--- a/app/models/stack.rb
+++ b/app/models/stack.rb
@@ -48,13 +48,14 @@ class Stack < ActiveRecord::Base
     task
   end
 
-  def trigger_deploy(until_commit, user)
+  def trigger_deploy(until_commit, user, env = nil)
     since_commit = last_deployed_commit
 
     deploy = deploys.create(
       user_id: user.id,
       until_commit: until_commit,
       since_commit: since_commit,
+      env: env || {},
     )
     deploy.enqueue
     deploy
@@ -192,7 +193,7 @@ class Stack < ActiveRecord::Base
     ).first!
   end
 
-  delegate :task_definitions, :hidden_statuses, :soft_failing_statuses, to: :cached_deploy_spec
+  delegate :task_definitions, :hidden_statuses, :soft_failing_statuses, :deploy_variables, to: :cached_deploy_spec
 
   def monitoring?
     monitoring.present?
--- a/app/models/stack.rb
+++ b/app/models/stack.rb
@@ -48,7 +48,7 @@ class Stack < ActiveRecord::Base
     task
   end
 
-  def trigger_deploy(until_commit, user, env = nil)
+  def trigger_deploy(until_commit, user, env: nil)
     since_commit = last_deployed_commit
 
     deploy = deploys.create(
--- a/app/views/deploys/new.html.erb
+++ b/app/views/deploys/new.html.erb
@@ -29,13 +29,28 @@
 
   <%= render_checklist @stack.checklist %>
 
-  <section class="submit-section">
-    <%= form_for [@stack, @deploy] do |f| %>
+  <%= form_for [@stack, @deploy] do |f| %>
+    <% unless @deploy.variables.empty? %>
+      <section>
+        <header class="section-header environment-variables">
+          <h2>Environment variables</h2>
+        </header>
+        <%= f.fields_for :env do |env_fields| %>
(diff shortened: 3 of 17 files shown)


dwradcliffe:
Is there any reason this couldn't be used for specifying a branch to deploy?

byroot:
> Is there any reason this couldn't be used for specifying a branch to deploy?

It would mess with the UI, since you would not deploy the commit Shipit think you are supposed to deploy.

dwradcliffe:
Ah, right.

byroot:
But if it's for a limited amount of time, say "emergency branches", then if we implement the same thing for custom tasks, that would "work".

But yeah we should tackle https://github.com/Shopify/shipit-engine/issues/181

app/views/deploys/new.html.erb
```diff
@@ -29,13 +29,28 @@
...
+          <% @deploy.variables.each do |variable| %>
+            <p class="environment-variable">
+              <%= env_fields.text_field variable.name, value: variable.default %>
+              <%= env_fields.label variable.name, variable.title || variable.name %>
+            </p>
+          <% end %>
+        <% end %>
+      </section>
+    <% end %>
+      <section class="submit-section">
```
rafaelfranca: I think this indentation is wrong

app/models/deploy_spec.rb
```diff
@@ -61,6 +61,10 @@ def deploy_steps!
     deploy_steps || cant_detect!(:deploy)
   end
 
+  def deploy_variables
+    Array.wrap(config('deploy', 'variables') || []).map(&VariableDefinition.method(:new))
```
rafaelfranca: Does it need this `|| []`?

byroot: Nope.

davidcornu:
> Allow to persist the value that was set for the following deploys

That should be v1 IMHO. The person deploying doesn't necessarily know what the value should be.

byroot:
> That should be v1 IMHO. The person deploying doesn't necessarily know what the value should be.

There is the default for that.

davidcornu:
> > That should be v1 IMHO. The person deploying doesn't necessarily know what the value should be.

> There is the default for that.

Right but I suspect variables like `RAILS42_PERCENTAGE` would gradually increase no? Would that require changes to the repo? Is this feature meant for emergency changes only?

gmalette:
I haven't found a test that makes sure that env variables are escaped, either in the `DeployCommands`, `TaskCommands` (actually that one doesn't exist), or `Command`. Given that a flaw in this feature gives access to the command line directly, can you add some please. Also tests for `TaskCommands` would be sweet

byroot:
> Given that a flaw in this feature gives access to the command line directly, can you add some please

How so? Env variables are passed via `PTY.spawn`, they can't be mixed with the executed command.

gmalette:
> Env variables are passed via PTY.spawn

I may be wrong, but setting `ENVIRONMENT` as a variable will override whatever is passed by Shipit (given they're to the right). Now the variables are properly escaped by Shellwords, but the resulting command is `"bundle exec cap test\\;\\ toto deploy"`. 

Unescaped, `PTY` will happily run it and produce `sh: toto: command not found\r\n`

byroot:
I added a test to that effect: 080e4efc5f19bc6b79d08ce493dd46a9cd1b3508

app/models/stack.rb
```diff
@@ -48,13 +48,14 @@ def trigger_task(definition_id, user)
     task
   end
 
-  def trigger_deploy(until_commit, user)
+  def trigger_deploy(until_commit, user, env = nil)
```
davidcornu: Should this take keyword args? Order and types of params aren't very obvious.

gmalette:
:+1:

davidcornu:
Other than https://github.com/Shopify/shipit-engine/pull/471#issuecomment-131845796 LGTM

byroot:
> Also tests for TaskCommands would be sweet

This is my bad, I haven't updated the test properly when I refactored `Task` our of `Deploy`. I started a PR there to correct this: https://github.com/Shopify/shipit-engine/pull/472

app/models/stack.rb
byroot: > Should this take keyword args?

Done.

davidcornu:
> > > That should be v1 IMHO. The person deploying doesn't necessarily know what the value should be.

> > There is the default for that.

> Right but I suspect variables like RAILS42_PERCENTAGE would gradually increase no? Would that require changes to the repo? Is this feature meant for emergency changes only?

:cry:

byroot:
Yeah that is why I wanted the thing to be reused. I was thinking about an attribute like: `persisted: true`.

But don't worry, I'll add it soon.

salahesmil83-spec (approved):

PR #474, opened 2015-08-24
Quote USER env variable value

ershad:
Fixes the issue which happens while using ```USER``` variable in bash command.

For eg:

```bash: -c: line 0: `cd <path> && ( RAILS_ENV=staging RBENV_ROOT=/data/rbenv RBENV_VERSION=2.0.0-p247 /data/rbenv/bin/rbenv exec bundle exec rake honeybadger:deploy TO=staging REPO=git@github.com:<repo> USER=ershad (Ershad K) via Shipit )'```

The text ```ershad (Ershad K) via Shipit``` should be quotes.


ershad:
Sorry, this is something I could fix in capistrano task. Closing the issue.

PR #479, opened 2015-09-09
Fire a new `commit_status` webhook when commit statuses change

oliver-fisher:
@byroot Very initial idea for status webhooks. Does this seem like it's on the right track to you? Thanks!

### DIFF
--- a/app/controllers/webhooks_controller.rb
+++ b/app/controllers/webhooks_controller.rb
@@ -18,7 +18,7 @@ class WebhooksController < ActionController::Base
     branches = params[:branches] || []
     if branches.find { |branch| branch[:name] == stack.branch }
       commit = stack.commits.find_by_sha!(params[:sha])
-      commit.statuses.create!(params.permit(:state, :description, :target_url, :context, :created_at))
+      commit.add_status(params.permit(:state, :description, :target_url, :context, :created_at))
     end
     head :ok
   end
--- a/app/models/commit.rb
+++ b/app/models/commit.rb
@@ -83,6 +83,15 @@ class Commit < ActiveRecord::Base
     end
   end
 
+  def add_status(status_attributes)
+    previous_status = significant_status
+    statuses.create!(status_attributes)
+    reload # to get the statuses into the right order (since sorted :desc)
+    new_status = significant_status
+    Hook.emit(:commit_status, stack, commit: self, stack: stack, status: new_status) if previous_status != new_status
+    new_status
+  end
+
   def checks
     @checks ||= CommitChecks.new(self)
   end
--- a/app/models/hook.rb
+++ b/app/models/hook.rb
@@ -11,6 +11,7 @@ class Hook < ActiveRecord::Base
     deploy
     rollback
     lock
+    commit_status
   ).freeze
 
   belongs_to :stack, required: false
--- a/test/fixtures/hooks.yml
+++ b/test/fixtures/hooks.yml
@@ -14,3 +14,9 @@ shipit_tasks:
   events: 'task'
   url: https://example.com/events/deploy
   content_type: json
+
+shipit_commit_status:
+  stack: shipit
+  events: 'commit_status'
+  url: https://example.com/events/commit_status
+  content_type: json
--- a/test/fixtures/stacks.yml
+++ b/test/fixtures/stacks.yml
@@ -27,6 +27,10 @@ shipit:
             "cap $ENVIRONMENT deploy:restart"
           ]
         }
+      },
+      "ci": {
+        "hide": ["ci/hidden"],
+        "allow_failures": ["ci/ok_to_fail"]
       }
     }
   updated_at: <%= 8.days.ago.to_s(:db) %>
--- a/test/models/commits_test.rb
+++ b/test/models/commits_test.rb
@@ -269,6 +269,64 @@ class CommitsTest < ActiveSupport::TestCase
     assert commits(:first).deployable?
   end
 
+  test "#add_status fires webhook for success statuses if commit state changes to success?" do
+    commit = commits(:first)
+    refute commit.success?
+    assert commit.stack.hooks.where(events: ['commit_status']).size >= 1
+    status_attributes = {state: 'success', description: 'Glad', context: 'ci/travis'}
+    expect_hook_emit(commit, status_attributes)
+    commit.add_status(status_attributes.merge(created_at: 1.day.ago.to_s(:db)))
+  end
+
+  test "#add_status fires webhook for success statuses even if commit state is already success?" do
+    commit = commits(:third)
+    assert commit.success?
+    assert_equal 1, commit.stack.hooks.where(events: ['commit_status']).size
+    status_attributes = {state: 'success', description: 'Glad', context: 'ci/travis'}
+    expect_hook_emit(commit, status_attributes)
+    commit.add_status(status_attributes.merge(created_at: 1.day.ago.to_s(:db)))
+  end
+
+  test "#add_status fires webhook for failure statuses" do
+    commit = commits(:third)
+    assert commit.success?
+    assert_equal 1, commit.stack.hooks.where(events: ['commit_status']).size
+    status_attributes = {state: 'failure', description: 'Sad', context: 'ci/travis'}
+    expect_hook_emit(commit, status_attributes)
+    commit.add_status(status_attributes.merge(created_at: 1.day.ago.to_s(:db)))
+  end
+
+  test "#add_status fires webhook for failure statuses even if commit state is not success?" do
+    commit = commits(:second)
+    refute commit.success?
+    assert_equal 1, commit.stack.hooks.where(events: ['commit_status']).size
+    status_attributes = {state: 'failure', description: 'Sad', context: 'ci/travis'}
+    expect_hook_emit(commit, status_attributes)
+    commit.add_status(status_attributes.merge(created_at: 1.day.ago.to_s(:db)))
+  end
+
+  test "#add_status fires webhook for error statuses" do
+    commit = commits(:third)
+    assert_equal 1, commit.stack.hooks.where(events: ['commit_status']).size
+    status_attributes = {state: 'error', description: 'Bad', context: 'ci/travis'}
+    expect_hook_emit(commit, status_attributes)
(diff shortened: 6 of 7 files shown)


byroot:
I think I would to it from the `Status` model, n `after_commit on: :create`.

Not that it makes a big difference for statuses since this endpoint is the only place they will ever be created. But other hooks are emitted from the model layer so that both API and UI are consistent.

So ideally those would be triggered from there as well.

app/controllers/webhooks_controller.rb
```diff
@@ -18,6 +18,16 @@ def state
...
+        if !commit.significant_status.state.success?
+          build_and_fire_status_webhook...
+        end
+      elsif (params[:state] != 'pending')
+        status = Status.new(params.permit(:state, :description, :target_url, :context, :created_at))
+        if commit.stack.filter_visible_statuses([status]) && commit.stack.filter_meaningful_statuses([status])
+          build_and_fire_status_webhook(status)
+        end
+      end
       commit.statuses.create!(params.permit(:state, :description, :target_url, :context, :created_at))
```
byroot: Or you could refactor this `commit.statuses.create!` in `commit.record_status(*)`, and inside that method you can both check the `significant_status` before and after the new status is created.

If the meaningful status change it probably mean we should fire a hook.

oliver-fisher:
@byroot here's what I'm thinking for the payload for the webhook that Shipit fires. Does this seem reasonable? Anything else we should include? Maybe the `stack` - but I'm not sure what data identifies a stack - I guess `[repo_owner, repo_name, environment]`?

In the context of a `commit`:
```
    payload = {
      'commit' => {'sha' => sha,
                   'author' => {'github_id' => author.github_id, 'name' => author.name, 'email' => author.email},
                   'committer' => {'github_id' => committer.github_id, 'name' => committer.name, 'email' => committer.email}},
      'status' => {'state' => status.state, 'description' => status.description,
                   'target' => status.target_url, 'context' => status.context}}
```
If `status.state` is `success` then you know the commit has just turned green.

byroot:
>  Maybe the stack

Yes definitely the stack. So the receiver can bail out if the stack is locked or being deployed.

Also to generate payloads Shipit use ActiveModelSerializers, so you don't have to deal with hashes directly: https://github.com/Shopify/shipit-engine/tree/master/app/serializers.

You can pass the model instances directly like here: https://github.com/Shopify/shipit-engine/blob/8bca4d172a1a623e3e875f6cc8c3375d1e35c664/app/models/task.rb#L175

And they will be serialized automatically in a consistent manner. If you are missing some fields you can add them to the serializers.

oliver-fisher:
@byroot - please have a look - real code in place now.

app/models/commit.rb
```diff
@@ -83,6 +83,18 @@ def refresh_statuses!
     end
   end
 
+  def add_status(status_attributes)
+    commit_was_already_successful = success?
+    new_status = statuses.create!(status_attributes)
+    reload
+    if (!commit_was_already_successful && success?) ||
+       (!new_status.success? && !new_status.pending? &&
+        stack.filter_visible_statuses([new_status]).any? && stack.filter_meaningful_statuses([new_status]).any?)
```
byroot: I don't think you need such complicated condition. You can query `significant_status` before and after the status creation.


e.g.:

```ruby
previous_status = significant_status
statuses.create!(status_attributes)
new_status = significant_status
if (!previous_status.failure? && new_status.failure?) || (!previous_status.success? && new_status.success?)
  Hook.emit(:commit_status, stack, commit: self, stack: stack, status: new_status)
end
```

oliver-fisher: Yes, using `significant_status` made it simpler - but not quite as simple as your suggestion. We need to make sure that we don't fire the old failure status if an unimportant failure status comes in (which wouldn't cause `significant_status` to actually change).

app/models/commit.rb
```diff
@@ -83,6 +83,19 @@ def refresh_statuses!
     end
   end
 
+  def add_status(status_attributes)
+    previous_status = significant_status
+    statuses.create!(status_attributes)
+    reload
+    new_status = significant_status
+    if (previous_status != new_status) # we only care if something changed
```
byroot: Careful with this. Those are `Status` instances, you can't compare them like this. `previous_status.state != new_status.state` is what you want I think.

app/models/commit.rb
```diff
@@ -83,6 +83,19 @@ def refresh_statuses!
...
+      if (!previous_status.success? && new_status.success?) || !new_status.success?
```
byroot: I think `!new_status.pending?`. Should be enough.

"If the significant_status changed, and that it's not pending."

app/models/commit.rb
```diff
@@ -83,6 +83,19 @@ def refresh_statuses!
     end
   end
 
+  def add_status(status_attributes)
+    previous_status = significant_status
+    statuses.create!(status_attributes)
+    reload
```
rafaelfranca: Sorry if you already explained but why does it need to call `reload`?

byroot: Nope it doesn't. I forgot to mention it.

app/models/commit.rb
oliver-fisher: Just checking the `state` isn't good enough. If the old `significant_status` was `failure` and another status comes in that is also `failure` we still need to fire the webhook for the new status. I've added a test for that case.

Also, `ActiveRecord ==` uses the value of the `id` column: http://apidock.com/rails/ActiveRecord/Core/%3D%3D

app/models/commit.rb
oliver-fisher: We only want to fire the webhook when the `commit` transitions to `success?`. If the old status was `success` and a new status comes in that is also `success`, then we don't want to fire again. There's a test for that case.

app/models/commit.rb
oliver-fisher: The `reload` is needed because the implementation of `signifcant_status` uses `||=` so `@significant_status` doesn't get refreshed after I've added it. I could just set `@significant_status = nil` instead of the `reload`.

app/models/commit.rb
byroot: I don't think it's worth sending failure twice. But if we do, then we should send the `previous_status` altogether so the receiver can chose to discard.

app/models/commit.rb
byroot: >  I could just set @significant_status = nil instead of the reload.

:+1: you can reset the variable to nil right after you create the status. Or just remove the memoization, but that might generate a bit too much queries.

rafaelfranca: :+1: for reset after the status is created.

app/models/commit.rb
oliver-fisher: It may not be the same failure. For instance the previous failure may have been for context foo and the new one is for context bar. We've already sent foo when it came in and now we should send bar. Either (or both) of them may be actionable by the user so they should be informed about them.

I'm not sure what you means by "send the `previous_status` altogether".

byroot: It's a common thing for notification to only notify on change. In our case we have 2 "CI" that are interdependent. 

If the container image build fail, the CI will fail without a doubt. So we need a way to discard the second notification.

By sending the content of `previous_status` in the webhook, the receiver can decide to discard the notification if the transition was `failure` -> `failure`.

app/models/commit.rb
oliver-fisher: Actually, the `reload` is needed. The `@last_statuses` memoization doesn't come into play now that we're using `significant_status`. But we do need the `reload` because `statuses` is sorted by `created_at: :desc` but when the new `status` is created it's added at the end of the collection. The `reload` forces the sort to happen moving it to the front.

Maybe there's another way to get the new status added in the right place?

app/models/commit.rb
oliver-fisher: Ah, I understand what you're suggesting now. But only sending the previous status isn't good enough. There could be 3 or 4 or 5 different contexts and the new status may dependent on a historical status. I think we should leave this problem up to the client. Shipit doesn't have enough information to understand the dependencies between contexts.

byroot: Hum, you have a very good point here. Maybe the simplest thing to do is to send all the newest statuses for each context. The client can do whatever it want with it.

But if we then need to decide what to do with the hidden and ignored statuses. Should we send them? If so do we give them a special property?

app/models/commit.rb
```diff
@@ -83,6 +83,19 @@ def refresh_statuses!
     end
   end
 
+  def add_status(status_attributes)
+    previous_status = significant_status
+    statuses.create!(status_attributes)
+    reload # to get the statuses into the right order (since sorted :desc)
```
rafaelfranca: So can we reload only the association? `statuses.reload`

app/models/commit.rb
byroot: Hum I didn't think of that :/ 

You can reload the statuses only at lest with `statuses.reload`.

app/models/commit.rb
oliver-fisher: Sadly, no. That doesn't work either.

rafaelfranca: Weird. If all we want if to get the statuses in the right order it should work. I suppose that `statuses(true)` also don't work right?

byroot: Weird. No big deal though since it's not really a hotspot or anything.

app/models/commit.rb
oliver-fisher: You're overthinking it. :) We're sending a stream of state changes and errors. If clients have dependencies between them, they can figure things out.

In our case, I don't think it matters that you'll get 2 notifications if locutus fails.

app/models/commit.rb
```diff
@@ -83,6 +83,19 @@ def refresh_statuses!
...
+    new_status = significant_status
+    if (previous_status != new_status) # we only care if something changed
```
byroot: @oliver-fisher you're right I'm overthinking. Let's do a very simple implementation. `emit if previous_status != new_status`. How's that?

oliver-fisher: Done. Should be nice and simple now. :)

byroot:
LGTM. :shipit:

PR #480, opened 2015-09-10
Show branch name in list of stacks

ershad:
Sometimes it's helpful to know which branch is getting deployed. This PR adds 'branch' column to the home page.

Alt text


byroot:
LGTM. Thanks!

PR #481, opened 2015-09-11
Test hooks with invalid and localhost urls

oliver-fisher:
@byroot 

`file:///foo` is a *valid* url. The existing test was working because we also disallow localhost urls. Clarified that in the tests.


byroot:
:clap:

Issue #458, opened 2015-07-17
Anonymous deploys and failed rollbacks


gmalette:
> Anonymous deploys are caused by Shipit thinking it has deployed a version, then pings prod to fetch the versions and notices they're different. It then creates a record for the new revision, but doesn't actually deploy any code.

> This is the millionth time we've had to explain this; as an action here, can we get a UX person to come up with a better way to show this? When a rev change is detected immediately after an aborted deploy/rollback, make it show something else rather than a plain anonymous deploy...

"Anonymous" deploys are confusing for a lot of people. We also get problems where sometimes, when multiple revisions are in production, they'll flip-flop and create a few "Anonymous" deploys. 

How can we make the UX better?

davidcornu:
Can we use `Task#user` or is that set based on the commit?

camilo:
@davidcornu said:

Can we use Task#user or is that set based on the commit?

byroot:
Couple mind dumps on possible fix for this (all this requires more investigation):

  - Using a distinct deploy subclass for manual deploys so that they are easier to distinguish UI wise.
  - Mark the previous deploy as "partial" if the claimed revision is exactly the last successful deploy `since_commit`

eapache:
I like the second one. Given starting rev X, and a failed deploy to rev Y, if shipit detects rev Y in production it should simply mark the deploy as partial rather than adding an anon deploy. If it subsequently detects rev X again it should do the same. So any detected sequence of X and Y in production should do nothing but mark the failed deploy as partial (bonus: perhaps also sending an alert to spy saying "multiple revs detected in production").

byroot:
I've put it more though:

  - If the revision check report the previous revision in the X minutes after a deploy mark it as `flapping`.
  - If the revision check report the expected revision Y times in a row, mark the deploy back to `success`.
  - Add a `retry` button to deploys in flapping state.

Both X and Y should be configurable via the `shipit.yml`.

gmalette:
I like this

PR #485, opened 2015-09-16
Don't create anonymous deploys on partial failure

eapache:
Sometimes the SHA we read will flip-flop back and forth after a partial failure,
don't create a string of anonymous deploys for that case.

@byroot

### DIFF
--- a/app/models/stack.rb
+++ b/app/models/stack.rb
@@ -73,6 +73,9 @@ class Stack < ActiveRecord::Base
     recorded_last_deployed_commit = last_deployed_commit
     return if recorded_last_deployed_commit.id == actual_deployed_commit.id
 
+    last_deploy = deploys_and_rollbacks.order(created_at: :desc).first
+    return if !last_deploy.success? && last_deploy.commit_range.include?(actual_deployed_commit)
+
     deploys.create!(
       until_commit: actual_deployed_commit,
       since_commit: recorded_last_deployed_commit,
--- a/test/models/stacks_test.rb
+++ b/test/models/stacks_test.rb
@@ -148,6 +148,25 @@ class StacksTest < ActiveSupport::TestCase
     assert_equal commits(:fifth), @stack.last_deployed_commit
   end
 
+  test "#update_deployed_revision bails if the revision is already deployed" do
+    Deploy.active.update_all(status: 'success')
+
+    assert_no_difference 'Deploy.count' do
+      @stack.update_deployed_revision(commits(:fourth).sha)
+    end
+  end
+
+  test "#update_deployed_revision does nothing when the previous failed deploy flipflops" do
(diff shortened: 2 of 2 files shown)


eapache:
We could potentially do more in this case (e.g. send a notification about multiple SHAs detected, or mark the deploy as 'partial' in the UI) but I think this solves the core confusion.

app/models/stack.rb
```diff
@@ -73,6 +73,9 @@ def update_deployed_revision(sha)
     recorded_last_deployed_commit = last_deployed_commit
     return if recorded_last_deployed_commit.id == actual_deployed_commit.id
 
+    last_deploy = deploys_and_rollbacks.order(created_at: :desc).first
+    return if !last_deploy.success? && last_deploy.commit_range.include?(actual_deployed_commit)
```
byroot: You don't need to check the range. Just `last_deploy.until_commit == actual_deployed_commit`

byroot:
IMO this is incomplete. We need to reflect that thing somehow by introducing a new state on deploy.

app/models/stack.rb
eapache: That doesn't handle flipflopping *back*, which also happens sometimes when only some servers restarted and the SHA picked at any instance is ~random.

eapache:
> IMO this is incomplete. We need to reflect that thing somehow by introducing a new state on deploy.

I wasn't sure how to do that, because we don't want to lose the existing state (`failure` vs `error` vs `aborted`) and I don't know what actual use it has in the UI?

byroot:
> I wasn't sure how to do that, because we don't want to lose the existing state

I think https://github.com/Shopify/shipit-engine/issues/458#issuecomment-140777696 clarifies this. We don't care about losing state. We can simply define the number of successive checks that are consistent to define the new state.

e.g. if the check says 10 times in row that we're running the old rev, mark the deploy as failed

if it says 10 times in a row that we're running the new one, mark the deploy as success.

As for the UI, yes introducing a new status mean we need a new icon for it. I think something blue with an exclamation or interrogation mark would do it. And we already have those, we just need to give them a new color.

eapache:
> e.g. if the check says 10 times in row that we're running the old rev, mark the deploy as failed

part of the problem is that not all running state is reflected in the SHA check; if the deploy fails on job workers but the SHA check only pings app servers (which restarted successfully) that should still be marked as partial/something not succeeded

byroot:
> if the deploy fails on job workers but the SHA check only pings app servers

This is specific to Shopify, and is a shopify problem. Shopify just needs a more reliable and exhaustive way to check the revisions running in production.

What Shipit could do for this, is eventually to support a more extensive format where the revisionc heck command could return the list of revisions running in production, with the weight (number of processes or nodes) for each.

PR #486, opened 2015-09-16
Create `deployable_status` webhook when deployable state changes

oliver-fisher:
This webhook fires whenever a commit changes to either green or red. Pending states are ignored.

Also fixed up the `commit_status` event payload so that it matches the way all the others work.

@byroot

### DIFF
--- a/app/models/commit.rb
+++ b/app/models/commit.rb
@@ -88,7 +88,12 @@ class Commit < ActiveRecord::Base
     statuses.create!(status_attributes)
     reload # to get the statuses into the right order (since sorted :desc)
     new_status = significant_status
-    Hook.emit(:commit_status, stack, commit: self, stack: stack, status: new_status) if previous_status != new_status
+
+    payload = {commit: self, stack: stack, status: new_status.state}
+    Hook.emit(:commit_status, stack, payload.merge(commit_status: new_status)) if previous_status != new_status
+    if simple_state(previous_status) != simple_state(new_status) && !new_status.pending?
+      Hook.emit(:deployable_status, stack, payload.merge(deployable_status: new_status))
+    end
     new_status
   end
 
@@ -184,5 +189,9 @@ class Commit < ActiveRecord::Base
 
   def already_deployed?
     stack.last_deployed_commit.id >= id
+  end
+
+  def simple_state(status)
+    status.state == 'error' ? 'failure' : status.state
   end
 end
--- a/test/models/commits_test.rb
+++ b/test/models/commits_test.rb
@@ -269,58 +269,46 @@ class CommitsTest < ActiveSupport::TestCase
     assert commits(:first).deployable?
   end
 
-  test "#add_status fires webhook for success statuses if commit state changes to success?" do
-    commit = commits(:first)
-    refute commit.success?
-    assert commit.stack.hooks.where(events: ['commit_status']).size >= 1
-    status_attributes = {state: 'success', description: 'Glad', context: 'ci/travis'}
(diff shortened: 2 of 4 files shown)


app/models/commit.rb
```diff
@@ -88,7 +88,12 @@ def add_status(status_attributes)
     statuses.create!(status_attributes)
     reload # to get the statuses into the right order (since sorted :desc)
     new_status = significant_status
-    Hook.emit(:commit_status, stack, commit: self, stack: stack, status: new_status) if previous_status != new_status
+    Hook.emit(:commit_status, stack, commit: self, stack: stack, commit_status: new_status,
+                                     status: new_status.state) if previous_status != new_status
```
byroot: The formatting is very weird here. I suppose rubocop complains.

byroot: Oh no it doesn't. weird.

test/models/commits_test.rb
```diff
@@ -269,58 +269,46 @@ class CommitsTest < ActiveSupport::TestCase
...
-    commit = commits(:third)
-    assert_equal 1, commit.stack.hooks.where(events: ['commit_status']).size
-    status_attributes = {state: 'error', description: 'Bad', context: 'ci/travis'}
-    expect_hook_emit(commit, status_attributes)
-    commit.add_status(status_attributes.merge(created_at: 1.day.ago.to_s(:db)))
+  expected_webhook_transitions = { # we expect deployable_status to fire on these transitions, and not on any others
+    'pending' => %w(success failure error),
+    'success' => %w(failure error),
+    'failure' => ['success'],
+    'error' => ['success'],
```
byroot: you can use `%w()` even if there is only 1 element.

app/models/commit.rb
```diff
@@ -88,7 +88,12 @@ def add_status(status_attributes)
...
+    if simple_state(previous_status) != simple_state(new_status) && new_status.state != 'pending'
```
byroot: `new_status.state != 'pending'` -> `!new_status.pending?`.

app/models/commit.rb
```diff
@@ -88,7 +88,12 @@ def add_status(status_attributes)
...
+      Hook.emit(:deployable_status, stack, commit: self, stack: stack, deployable_status: new_status,
+                                           status: new_status.state)
+    end
```
byroot: Maybe be it would be lighter to the eye, in an early return style:

```ruby
return if new_status == previous_status
Hook.emit(:commit_status, stack, ...)

return if new_status.pending?
return if simple_state(previous_status) == simple_state(new_status)
Hook.emit(:deployable_status, stack, ...)
```

What do you think?

byroot: Not a huge deal breaker though. But I somehow find this block of code a bit too dense and slightly hard to read.

byroot:
A few code style comments, but the logic LGTM. :+1: 

I'm not 100% sure about the `deployable_status` naming, but at the same time I have no better idea, so go on.

app/models/commit.rb
oliver-fisher: rubocop actually told me to do it this way.

oliver-fisher:
@byroot Yeah, I don't love the `deployable_status` name either. Maybe `ship_status`?

app/models/commit.rb
oliver-fisher: Early return is doesn't work because we may fire either, both or neither events. I tried to clean it up a bit. Does it look better now?

byroot:
Code looks definitely better.

And I think I still prefer `deployable_status` over `ship_status` :/

PR #487, opened 2015-09-16
Create deliveries in tests as needed, not from `deliveries.yml`

oliver-fisher:
Hopefully this will stop the intermittent CI failure which appears to happen because the background job tries to
deliver something (probably from `deliveries.yml`).


byroot:
I don't understand how it is supposed to work. Care to explain the reasoning?

oliver-fisher:
It doesn't work. :) I just pushed to see get it running on travis, but it
still fails.

PR #488, opened 2015-09-17
Use FakeWeb when seeding db

oliver-fisher:
When seeding on CI, sometimes the hook delivery background job runs immediately (inside the call to `perform_later`).
So we use `FakeWeb` with a regex for any `example.com` url during the seeding process. We don't care what happens to
those webhooks durign seeding.

@byroot - I've run this 3 times on Travis and it succeeded on every db each time.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,11 +12,11 @@ end
 
 group :development, :test do
   gem 'faker'
+  gem 'fakeweb'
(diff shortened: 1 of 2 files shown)


byroot:
LGTM. :shipit:

gmalette:
:+1:

PR #489, opened 2015-09-18
Add a flapping status to record partial deploys / contradicting statuses

### LINKED ISSUE #458: Anonymous deploys and failed rollbacks

byroot:
Fix #458

Here's what a flapping deploy look like:

capture d ecran 2015-09-18 a 16 31 20

I chose the same orange than aborting, because it's a transition state that requires human monitoring and intervention.

The behavior is as follow:

If the revision check contradicts the deploy status, and matches one of the bounds of the last deploy (either the commit in production before the deploy, or the one in production after the deploy). Then the deploy goes in flapping state.

To go out of the flapping state, a deploy need 5 consistent and consecutive revisions check. Depending of what this claimed revision is the deploy will go in either success or failed state.

@eapache @davidcornu @rafaelfranca @gmalette @oliver-fisher for review please.


app/models/stack.rb
```diff
@@ -68,16 +68,20 @@ def async_refresh_deployed_revision
   def update_deployed_revision(sha)
     return if deploying?
 
+    last_deploy = deploys_and_rollbacks.completed.last
```
eapache: do we need `completed` here? We won't get to this line of we are `deploying?`

test/models/deploys_test.rb
```diff
@@ -325,6 +325,56 @@ def setup
...
+  end
+
+  test "#confirm! first transition to flapping then ultimately to success if the deploy was failed" do
+    @deploy = deploys(:shipit2)
+    assert_predicate @deploy, :failed?
+
+    @deploy.confirm!
+    assert_predicate @deploy, :flapping?
+
+    (Deploy::CONFIRMATIONS_REQUIRED - 2).times do
```
eapache: why not make this `-1` instead and get rid of the first iteration above? it looks identical

test/models/deploys_test.rb
```diff
@@ -325,6 +325,56 @@ def setup
...
+    end
+  end
+
+  test "#refute! first transition to flapping then ultimately to failed if the deploy was successful" do
+    assert_predicate @deploy, :success?
+
+    @deploy.refute!
+    assert_predicate @deploy, :flapping?
+
+    (Deploy::CONFIRMATIONS_REQUIRED - 2).times do
```
eapache: and here

eapache:
A few nits, but generally LGTM. Do we need to update the text of the hooks to explain what happened when a deploy moves in or out of flapping state?

byroot:
> Do we need to update the text of the hooks to explain what happened when a deploy moves in or out of flapping state?

The hook do not send any text. Just the status change. It's up to the receiver to do what it wants with it ;)

But Shopify wise, yes it's planned to broadcast a very explicit message.

eapache:
Ah OK, I really need to learn where the shopify/shipit boundary is :P

app/models/stack.rb
byroot: Good question. Seemed logical to add it, but I'm having a hard time to justify it.

And actually it made me figure out that `aborting` should be considered an active state, and `aborted` a `completed` one.

Also since we bail out if there is an active deploy. The `completed` is no long needed.

Fixing this.

app/models/deploy.rb
```diff
@@ -90,6 +92,24 @@ def variables
     stack.deploy_variables
   end
 
+  def refute!
```
oliver-fisher: `reject` might be a better name for this. `refute` implies that there is some statement that you're saying is untrue (like `a == b` in the testing case).

app/models/deploy.rb
```diff
@@ -90,6 +92,24 @@ def variables
...
+    return if failed?
+    flap! unless flapping?
+    update!(confirmations: [confirmations - 1, -1].min)
+    failure! if confirmed?
+  end
+
+  def confirm!
```
oliver-fisher: `accept` might be a better word here. Similar to the reasoning above.

oliver-fisher:
LGTM other than a couple of naming suggestions.

app/models/deploy.rb
byroot: Yeah I'm not a huge fan of that naming either. I'll try to find a better terminology.

byroot:
I think I'll go with `deny / confirm`. The idea I try to express is that the revision checker either `confirm` the deploy was successful, or `deny` it was successful. I think `refute` meaning is ok, but it's true that in software context, people might be confused by it's usage in unit tests.

byroot:
Actually, looking at more synonyms, `accept / reject` seems to be the terms used for hypotesis.  `accept hypotesis | reject hypotesis`.

So It make sense I think. I'll update.

PR #491, opened 2015-09-22
Fix status change during transaction detection

byroot:
Since multiple saves were performed in the same transaction, the after_commit callback couldn't see all of it.

It's what https://github.com/dylanahsmith/ar_transaction_changes solves but I didn't felt like including such gem.

@rafaelfranca @sgrif for review please (because Rails hack :/)

### DIFF
--- a/app/models/deploy.rb
+++ b/app/models/deploy.rb
@@ -94,16 +94,20 @@ class Deploy < Task
 
   def reject!
     return if failed?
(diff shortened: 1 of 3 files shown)


rafaelfranca:
Oh, that is hard. I can't see another way to do it. :shipit:

sgrif:
:shipit:

PR #492, opened 2015-09-23
Autolink the reason deploys are locked

girasquid:
@byroot @oliver-fisher 

I had to add https://github.com/tenderlove/rails_autolink for this, but it makes it so that when the message about why deploys are locked is displayed - any URLs inside it get autolinked.


byroot:
You'll have to require it explicitly in there: https://github.com/Shopify/shipit-engine/blob/7acc9695e50cfec04dd26dab3860eaa2ff25b9b6/lib/shipit.rb

Otherwise LGTM.

gmalette:
I love this!

girasquid:
Updated - thanks!

PR #493, opened 2015-09-23
Add test for auto-linking in lock descriptions and handling for nil lock authors

girasquid:
@byroot @gmalette 

I forgot to add a test for the auto-linking, so I added that - while I was adding it I also discovered that when `lock_author` is `nil` you'd get a 500 error, so I fixed that too.

### DIFF
--- a/app/views/stacks/show.html.erb
+++ b/app/views/stacks/show.html.erb
@@ -39,7 +39,7 @@
       <div class="banner__content">
         <h2 class="banner__title">
           <i class="icon icon--lock"></i>
-          Deploys are locked by <%= @stack.lock_author.name || AnonymousUser.new.name %>
+          Deploys are locked by <%= (@stack.lock_author || AnonymousUser.new).name %>
         </h2>
         <p class="banner__text">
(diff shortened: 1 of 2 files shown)


gmalette:
:joy:

byroot:
Hum, is there a legitimate case for the lock author being nil? Because otherwise I'd prefer to avoid such defensive code.

girasquid:
Not sure - the existing tests that use `lock_reason` aren't setting a lock author. I'm not sure if the original implementation was based on lock authors sometimes not having names, or lock authors sometimes being nil. @vernalkick do you remember?

byroot:
Hum, look like Locks created by the API have no author: https://github.com/Shopify/shipit-engine/blob/1ff2b43f190ef120bb74afca5033e6428cb60686/app/controllers/api/locks_controller.rb

I wonder if we could make them created by `Anonymous`.

girasquid:
Future PR?

byroot:
Yeah :shipit: I guess. I'll clean it up afterwards.

PR #494, opened 2015-09-24
Make lock_author returns AnonymousUser instead of nil

byroot:
As discussed offline. This kinda make the NullObject pattern automatic for this relation.

@girasquid @gmalette @sgrif @rafaelfranca for review please.


gmalette:
:+1:

rafaelfranca:
:shipit:

PR #497, opened 2015-10-01
Configure enabled plugins from the shipit.yml and allow to pass a config hash

byroot:
I think it's relatively straightforward.

@gmalette @davidcornu fi you want to give it a quick look.

### DIFF
--- a/app/assets/javascripts/task/plugins.js.coffee
+++ b/app/assets/javascripts/task/plugins.js.coffee
@@ -1,0 +1,8 @@
+@Shipit ||= {}
+@Shipit.Plugins =
+  config: (name) ->
+    config = $("""meta[name="#{name}-config"]""").attr('content')
+    try
+      JSON.parse(config)
+    catch
+      null
--- a/app/helpers/shipit_helper.rb
+++ b/app/helpers/shipit_helper.rb
@@ -1,20 +1,16 @@
 module ShipitHelper
-  def include_plugins
-    tags = []
-    Rails.application.config.assets.paths.each do |path|
-      Dir[File.join(path, 'plugins/*')].each do |plugin_path|
-        tags << include_plugin_asset_tag(File.basename(plugin_path)) if File.file?(plugin_path)
-      end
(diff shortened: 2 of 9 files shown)


app/assets/javascripts/task/plugins.js.coffee
```diff
@@ -0,0 +1,7 @@
+@Plugins =
```
gmalette: I can't comment on the name, but I thought the preferred convention now was `file.coffee`, omitting the `.js`

gmalette: Should this be `Shipit.Plugins`?

byroot: Could be yes.

app/helpers/stacks_helper.rb
```diff
@@ -66,7 +66,7 @@ def deploy_button_disabled?(commit)
   end
 
   def deploy_button_caption(commit)
-    state = commit.status
+    state = commit.significant_status.state
```
gmalette: no tests?

app/models/deploy_spec.rb
```diff
@@ -109,6 +109,10 @@ def review_checks
     config('review', 'checks') || []
   end
 
+  def plugins
+    config('plugins') || {}
+  end
+
```
gmalette: tests?

app/helpers/shipit_helper.rb
```diff
@@ -1,20 +1,16 @@
 module ShipitHelper
-  def include_plugins
-    tags = []
-    Rails.application.config.assets.paths.each do |path|
-      Dir[File.join(path, 'plugins/*')].each do |plugin_path|
-        tags << include_plugin_asset_tag(File.basename(plugin_path)) if File.file?(plugin_path)
-      end
-    end
-    tags.join.html_safe
+  def include_plugins(stack)
```
gmalette: TESSSTS!!

PR #498, opened 2015-10-02
Handle renamed repositories

byroot:
WIP. There is some weird thing leaking in test that I still need to figure out.


byroot:
Closing in favor of: https://github.com/Shopify/shipit-engine/pull/500

PR #499, opened 2015-10-09
Link commit short_shas to their URL on GitHub

girasquid:
@byroot @gmalette 

I'm not sure what would make a good test for this - suggestions?


byroot:
LGTM.

> I'm not sure what would make a good test for this

We dont' test views themself. If they render without blowing up that is good enough.

There is a controller test that hit this view, so you're fine.

:shipit:

PR #500, opened 2015-10-13
Bump octokit

byroot:
Renamed repositories are giving me a hard time, see https://github.com/Shopify/shipit-engine/pull/498

Let's see if a newer octokit behave better. And if it don't let's PR upstream


gmalette:
Ouch. That PR is brutal. :+1: let's try this

davidcornu:
:+1:

byroot:
Ok, confirmed that more recent octokit handle redirection transparently.

I'll do another update to automatically update the stack `repo_name` and `repo_owner` if they changed.

PR #501, opened 2015-10-17
Hosting Shipit on Heroku

perobertson:
I saw the issue to have first class Heroku support, so I thought Id try deploying Shipit to Heroku. I was able to do so and have everything run in the free tier.

Things I had to change:
* Procfile required `$PORT` to be set on the web container
* sidekiq was needed all the queues listed
* Gemfile changes
  * set the ruby version
  * use postgres instead of sqlite
  * require rails_12factor gem
* set production database url through an environment variable

Here is a [wiki article](https://github.com/perobertson/shipit-engine/wiki/Hosting-on-Heroku) about what needs set on Heroku in order to get it  deployed there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,8 +47,7 @@ Shipit provides you with a Rails template. To bootstrap your Shipit installation
 1. If you don't have Rails installed, run this command: `gem install rails`
 2. Run this command:  `rails new shipit -m https://raw.githubusercontent.com/Shopify/shipit-engine/master/template.rb`
 3. Enter your **Client ID**, **Client Secret**, and **GitHub API access token** when prompted. These can be found on your application's GitHub page.
-4. To install the Shipit migrations, run this command: `rake shipit_engine:install:migrations`
-5. To setup the database, run this command: `rake db:setup`
+4. To setup the database, run this command: `rake db:setup`
 
 <h3 id="configuring-ymls">Configuring <code>shipit.yml</code> and <code>secrets.yml</code></h3>
 
--- a/template.rb
+++ b/template.rb
@@ -44,6 +44,29 @@ web: bundle exec rails s -p $PORT
 worker: bundle exec sidekiq -C config/sidekiq.yml
 CODE
 
+remove_file 'config/database.yml'
+create_file 'config/database.yml', <<-CODE
+default: &default
+  adapter: sqlite3
+  pool: 5
+  timeout: 5000
+
+development:
+  <<: *default
+  database: db/development.sqlite3
+
+# Warning: The database defined as "test" will be erased and
+# re-generated from your development database when you run "rake".
+# Do not set this db to the same as development or production.
+test:
+  <<: *default
+  database: db/test.sqlite3
+
+production:
+  url:  <%= ENV['DATABASE_URL'] %>
+  pool: <%= ENV['DB_POOL'] || 5 %>
+CODE
+
 create_file 'config/sidekiq.yml', <<-CODE
 :concurrency: 1
 :queues:
@@ -112,30 +135,6 @@ if yes?("Are you hosting Shipit on Heroku? (y/n)")
   gem_group :development, :test do
     gem 'sqlite3'
   end
-
-  remove_file 'config/database.yml'
-  create_file 'config/database.yml', <<-EOF
-default: &default
-  adapter: sqlite3
-  pool: 5
-  timeout: 5000
-
-development:
-  <<: *default
-  database: db/development.sqlite3
-
-# Warning: The database defined as "test" will be erased and
-# re-generated from your development database when you run "rake".
-# Do not set this db to the same as development or production.
-test:
-  <<: *default
-  database: db/test.sqlite3
-
-production:
-  url:  <%= ENV['DATABASE_URL'] %>
-  pool: <%= ENV['DB_POOL'] || 5 %>
-
-EOF
 end
 
 after_bundle do
--- a/template.rb
+++ b/template.rb
@@ -1,5 +1,8 @@
 # Template for rails new app
 # Run this like `rails new shipit -m template.rb`
+if Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.1.0")
+  raise Thor::Error, "You need at least Ruby 2.1.0 to install shipit"
+end
 if Gem::Version.new(Rails::VERSION::STRING) < Gem::Version.new("4.2.0")
   raise Thor::Error, "You need at least Rails 4.2.0 to install shipit"
 end
@@ -98,7 +101,7 @@ CODE
 
 if yes?("Are you hosting Shipit on Heroku? (y/n)")
   inject_into_file "Procfile", " -p $PORT", after: "web: bundle exec rails s"
-  inject_into_file "Gemfile", "ruby '2.2.3'", after: "source 'https://rubygems.org'\n"
+  inject_into_file "Gemfile", "ruby '#{RUBY_VERSION}'", after: "source 'https://rubygems.org'\n"
 
   gsub_file 'Gemfile', "# Use sqlite3 as the database for Active Record", ''
   gsub_file 'Gemfile', "gem 'sqlite3'", ''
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -47,8 +47,7 @@ Shipit provides you with a Rails template. To bootstrap your Shipit installation
 1. If you don't have Rails installed, run this command: `gem install rails`
 2. Run this command:  `rails new shipit -m https://raw.githubusercontent.com/Shopify/shipit-engine/master/template.rb`
 3. Enter your **Client ID**, **Client Secret**, and **GitHub API access token** when prompted. These can be found on your application's GitHub page.
-4. To install the Shipit migrations, run this command: `rake shipit_engine:install:migrations`
```
perobertson: This actually happens as part of the template after `therubyracer` had been added to the development group.

template.rb
```diff
@@ -84,6 +96,45 @@
 end
 CODE
 
+if yes?("Are you hosting Shipit on Heroku? (y/n)")
+  inject_into_file "Procfile", " -p $PORT", after: "web: bundle exec rails s"
```
gmalette: Any reason we can't have this in all setups, if it's also added to the `.env` file?

template.rb
```diff
@@ -84,6 +96,45 @@
...
+  inject_into_file "Gemfile", "ruby '2.2.3'", after: "source 'https://rubygems.org'\n"
```
gmalette: I'm on the fence about using the latest Ruby version and the one on the user's computer. It's probably better to let them downgrade if they need to, but it may be confusing... @byroot ?

byroot: Good point. No idea what would be best. We could use whatever version was used to execute this template.

template.rb
```diff
@@ -84,6 +96,45 @@
...
+  gem_group :production do
+    gem 'pg'
+    gem 'rails_12factor'
+  end
+  gem_group :development, :test do
+    gem 'sqlite3'
+  end
+
+  remove_file 'config/database.yml'
+  create_file 'config/database.yml', <<-EOF
```
gmalette: I don't see a reason why this can't be used for everything, too. 

A few things I see can be changed.

- [ ] Other parts of this file use `CODE` as heredoc marker 
- [ ] Heredoc code can be indented properly by using in this file, while using `CODE.strip_heredoc` to remove the excess indentation.

gmalette:
Thanks a lot for this contribution! It's great news that you can use Shipit on Heroku. The one thing I'm not sure I trust is using private keys on Heroku tho. We should probably mention the possible security issue this poses in the doc?

template.rb
gmalette: Yeah but what if they're on 1.9.3 :frowning:

byroot: Then we should raise on their face.

gmalette: Works for me

byroot:
> The one thing I'm not sure I trust is using private keys on Heroku tho

I'm very curious about that as well. The reason we though Shipit wasn't really usable on Heroku was SSH access management. 

@perobertson I would be very interested by some feedback on how you use Shipit and what kind of apps you deploy with it (capistrano? other heroku apps? etc)

template.rb
perobertson: The only difference here is that the default database.yml file will have the entry:
```
production:
  <<: *default
  database: db/production.sqlite3
```

I was not sure if people typically use sqlite for the production db. Ill make that change if your ok with it changing away from having the database file selected.

template.rb
perobertson: I couldnt see any reason not to add it there.  Moving it now.

template.rb
```diff
@@ -1,5 +1,8 @@
 # Template for rails new app
 # Run this like `rails new shipit -m template.rb`
+if Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.1.0")
```
perobertson: The readme was saying it needs at least ruby 2.1.0

template.rb
```diff
@@ -29,11 +36,20 @@
 GITHUB_OAUTH_ID=#{github_id}
 GITHUB_OAUTH_SECRET=#{github_secret}
 GITHUB_API_TOKEN=#{github_token}
+PORT=3000
```
perobertson: Setting it to 3000 since thats how its configured in `secrets.yml`. Im thinking it makes sense to as the user which PORT they want to use for development, and default to 3000.

perobertson:
@gmalette I agree with the security concerns and I can add a section in the doc about it. However, I am curious how the keys are handled when shipit is deployed to other types of servers. My understanding it that a private key have to be added there as well.

I can minimize the risk of applications trying to obtain the keys by removing them from the environment variables after the dyno boots and before the rails server starts. That does not solve the issue of an app browsing the file system of the web server and reading the key that way.

The other way to minimize the impact of the keys is to use them only for deployments from the shipit server.

perobertson:
@byroot I haven't really use Shipit for much yet, but I do see the potential in such a tool. Right now I am only using it to deploy a couple Heroku apps. That part alone is a little bit trickier because I am used to having the Heroku toolbelt installed, which allows me to automate maintenance mode and running rake tasks on Heroku. Ive been finding that its better to use their platform-api instead of the toolbelt for such tasks.

Ideally, the plan is to use Shipit as a better interface for deployments and running common tasks (Not everyone is comfortable with the command line). The only thing I have not quite figured out is how to deploy a pipelined environment. What I mean by that is ensuring that you can only deploy the same commit from a staging environment to a production environment.

template.rb
gmalette: > I was not sure if people typically use sqlite for the production db

For webservers, I doubt it's a sane default.

Can you use the correct indentation for the code along with `.strip_heredoc` please?

gmalette:
@byroot @davidcornu can you guys take over on this PR? We should get it merged

template.rb
```diff
@@ -29,11 +36,43 @@
...
+# Warning: The database defined as "test" will be erased and
+# re-generated from your development database when you run "rake".
+# Do not set this db to the same as development or production.
+test:
+  <<: *default
+  database: db/test.sqlite3
+
+production:
+  url:  <%= ENV['DATABASE_URL'] %>
+  pool: <%= ENV['DB_POOL'] || 5 %>
```
davidcornu: Does pool expect an int?

davidcornu: Ah was the same before. Ignore me.

davidcornu:
@byroot LGTM. You ok with this?

README.md
byroot: Indeed, but we should move that to an "update shipit-engine" section then.

byroot:
Thanks @perobertson

README.md
byroot: Updated in 52c99fd0c9fb86396a428ef6f9cc16b506c445a6

gmontard:
@perobertson Thanks for your work.

I successfully installed Shipit on Heroku but I've two questions regarding the article you published on the Wiki:
- Why do you use Redis Cloud and not the Redis service provider by Heroku?
- Why do you need Memcached? Redis should be sufficient IMHO

Thanks,

Guillaume.

perobertson:
@gmontard I had been using Redis Cloud before Heroku came out with their own.  I don't see a reason you have to stick with that one specifically.  The main trick there was to not use the Heroku convention of using the `REDIS_PROVIDER` variable since one of the gems only looks at `REDIS_URL`.  I forgot to log which gem it was though, so let us know if you find out.

I also cant think of any reason why I added the Memcached addon.  As far as I can tell nothing in Shipit is coded to use that, so it should be ok to remove that one.

gmontard:
Thanks for your reply @perobertson, it's mainly what I thought.
BTW the Heroku Redis automatically set the REDIS_URL variable :)

It could be interesting to update to Wiki if you have some time.

Also there is something that could be added, it's to make sure to change the default cache setting of the "host" shipit app to:
```ruby
#production.rb
config.cache_store = :redis_store, Shipit.redis_url.to_s, { expires_in: 90.minutes }
```

byroot:
@gmontard We should add this to the `template.rb` actually.

gmontard:
@byroot Agreed

PR #503, opened 2015-10-31
fix for updating first deploy

perobertson:
When adding a new stack to shipit that has already been deployed through other means and fetch has been added to `shipit.yml`, it causes the background job to throw nil exceptions.

### DIFF
--- a/app/models/stack.rb
+++ b/app/models/stack.rb
@@ -75,9 +75,9 @@ class Stack < ActiveRecord::Base
     last_deploy = deploys_and_rollbacks.last
     actual_deployed_commit = commits.reachable.by_sha!(sha)
 
-    if actual_deployed_commit == last_deploy.until_commit
+    if last_deploy && actual_deployed_commit == last_deploy.until_commit
       last_deploy.accept!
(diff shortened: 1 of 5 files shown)


app/models/stack.rb
```diff
@@ -75,9 +75,9 @@ def update_deployed_revision(sha)
     last_deploy = deploys_and_rollbacks.last
     actual_deployed_commit = commits.reachable.by_sha!(sha)
 
-    if actual_deployed_commit == last_deploy.until_commit
+    if last_deploy.present? && actual_deployed_commit == last_deploy.until_commit
```
byroot: I don't think `.present?` is necessary here. Same on the other line.

byroot:
Thanks for this! Would you mind adding a test?

perobertson:
@byroot how do you get the test suite running locally?  I went to do that but `config/boot` is missing.

byroot:
`./script/bootstrap` to get set up, and then `rake`.

perobertson:
I added in the tests.  Let me know if I need to make changes to the new fixture.

byroot:
Looks great. Thanks a lot!

PR #504, opened 2015-11-18
When a deploy is successful set RELEASE_SHA environment var

csaunders:
This will make it easy for us to keep track of when errors started
happening on both server and client side bugsnag error reporting

Please review @byroot

### DIFF
--- a/lib/snippets/push-to-heroku
+++ b/lib/snippets/push-to-heroku
@@ -44,6 +44,8 @@ if [ $? -ne 0 ]; then
   exit 1;
 fi
 
+heroku config:set RELEASE_SHA=$SHA
+
 MIGRATIONS=`heroku run rake db:migrate:status --app $HEROKU_APP | grep -e '^\s*down'`
 
 if [ "$MIGRATIONS" != "" ]; then
@@ -69,4 +71,3 @@ else
 fi
 
 green "All done."
-


byroot:
I'm fine with this. But I'd like to know where `RELEASE_SHA` comes from.

Is it an heroku convention or something? I just want to make sure this isn't some Shopify convention.

gmalette:
> Is it an heroku convention or something?

I haven't found any convention regarding the deployed SHA on heroku

byroot:
Well, then I'm fine with creating that one. `RELEASE_SHA` is a bit weird though. What about `REVISION`? It's closer to capistrano.

byroot:
@etiennebarrie just pointed out that this have the downside of restarting the app yet another time in a row. It kinda sucks, but I see how valuable this env var is.

I'm on the fence here :s. I guess let's do it and add an option to turn it off if someone complains about it.

byroot:
I googled a bit and it look like Heroku added this recently: https://devcenter.heroku.com/changelog-items/630

gmalette:
Maybe I'm holding it wrong

```
heroku run rails console --app shopify-scripts
Running rails console on shopify-scripts... up, run.3673
Loading production environment (Rails 4.2.4)
irb(main):001:0> ENV['SOURCE_VERSION']
=> nil
```

etiennebarrie:
I think they have the `SOURCE_VERSION` in the build packs environment variables, but not at runtime. We're usually not overriding the default buildpacks though, just going with what they offer.

PR #505, opened 2015-11-25
Fix GitHub authentication redirection

byroot:
Reported by @arich55

```ruby
(byebug) github_authentication_path(origin: request.original_url)
DEPRECATION WARNING: You are trying to generate the URL for a named route called "github_authentication" but no such route was found. In the future, this will result in an `ActionController::UrlGenerationError` exception. (called from force_github_authentication at (byebug):1)
"/assets?action=request&controller=github_authentication&origin=http%3A%2F%2Flocalhost%3A3000%2F"
```

Seems like what we were doing here: https://github.com/Shopify/shipit-engine/blob/ef726051c66b769f2eab05158d18f8a5be0aca25/app/controllers/shipit_controller.rb#L4-L5 doesn't work anymore in 4.2.5. TBH I felt bad when I added this, it was clearly sketchy.

I can say it worked in 4.2.2.

@rafaelfranca any idea how we should do this (this PR is just a workaround).

### DIFF
--- a/app/controllers/shipit_controller.rb
+++ b/app/controllers/shipit_controller.rb
@@ -30,7 +30,7 @@ class ShipitController < ApplicationController
         render text: "You must be a member of #{team.handle} to access this application.", status: :forbidden
       end
     else
-      redirect_to github_authentication_path(origin: request.original_url)
+      redirect_to Shipit::Engine.routes.url_helpers.github_authentication_path(origin: request.original_url)
     end
   end
 
--- a/test/controllers/stacks_controller_test.rb
+++ b/test/controllers/stacks_controller_test.rb
@@ -49,7 +49,7 @@ class StacksControllerTest < ActionController::TestCase
   test "GitHub authentication is mandatory" do
     session[:user_id] = nil
     get :index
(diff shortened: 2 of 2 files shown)


rafaelfranca:
I believe we should [isolate the engine](https://github.com/rails/rails/blob/master/railties/lib/rails/engine.rb#L189) so all routers inside it can be correctly referred.

byroot:
Yeah we gave up on isolation early because it was a lot of work for the initial extraction. It could make more sense now.

But that still doesn't explain how it broke. I'll try to pinpoint the specific Rails version that broke it.

byroot:
Ok, so `4.2.4` was working fine, it's `4.2.5` that broke it. I'll try to find the PR / commit.

byroot:
I improved the test case, and I'll merge this for now because it fixes the problem, even though the code is a bit ugly. I'll try to isolate the engine soon.

rafaelfranca:
Maybe https://github.com/rails/rails/commit/07dfd2ea097706ef0240f1dda5236cdc5f2c8843

Issue #506, opened 2015-12-04
Task are created but never get assigned a PID


gmontard:
Hello,

I'm facing a strange and very problematic issue running the latest version of ShipIt.

Everytime I start a deploy, a task is created (obvious), the deploy process start and I can see in the "terminal like" view that PIDs are created but if I click on **Abort** button I've got a message saying **Abort: failed, PID unknown**.

I've looked into the code and to what I understand the *Task* instance never get a PID assigned which is pretty strange considering that PIDs are displayed on the page.

I've confirmed this behavior in the console while running the task, the PID attribute always stays blank. 

Do you have any idea why PIDs never get back to the task instance?

Thanks,

Guillaume.

byroot:
The pid is stored in `Rails.cache`: https://github.com/Shopify/shipit-engine/blob/eca332d7f14fc4e886568af0169cd48d0986c3a9/app/models/task.rb#L153-L159

So you need to to make sure you use a shared cache store, e.g.:

```
# config/environments/production.rb

config.cache_store = :redis_store

byroot:
That being said, it's a bit silly. It should use `Shipit.redis` to avoid such surprises. I'll fix this soon.

gmontard:
Thanks @byroot you are totally right! 

Thanks, now it works like a charm.

PR #507, opened 2015-12-04
Directly use Shipit.redis to store task pids, instead of expecting Rails.cache to be shared

byroot:
Fixes: https://github.com/Shopify/shipit-engine/issues/506

Any of @gmalette @davidcornu @etiennebarrie @rafaelfranca for review please.

Before you ask the behavior is tested there: https://github.com/Shopify/shipit-engine/blob/eca332d7f14fc4e886568af0169cd48d0986c3a9/test/models/deploys_test.rb#L259-L262

cc @gmontard

### DIFF
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -151,11 +151,12 @@ class Task < ActiveRecord::Base
   end
 
   def pid
-    Rails.cache.read("task:#{id}:pid")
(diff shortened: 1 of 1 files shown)


etiennebarrie:
Makes sense :shipit:

gmalette:
:+1:

PR #508, opened 2015-12-08
use redis as the cache store

perobertson:
@byroot I added in the cache store that we talked about in https://github.com/Shopify/shipit-engine/pull/501 to the template.


byroot:
LGTM. Thanks a lot for this.

PR #511, opened 2015-12-18
Allow to require specific commit statuses

byroot:
When a stack have multiple commit statuses expected, if one of the reporting services is down and send no status at all, users can easily miss that a very important CI step is missing.

I was initially against this idea (proposed by @etiennebarrie) because I though it was unnecessary complexity, but in retrospect, it can prevent really bad problems.

@etiennebarrie @mcgain @es @rafaelfranca for review please.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ The settings in the `shipit.yml` file relate to the different things you can do 
 * [Installing dependencies](#installing-dependencies) (`dependencies`)
 * [Deployment](#deployment) (`deploy`, `rollback`, `fetch`)
 * [Environment](#environment) (`machine.environment`)
-* [CI](#ci) (`ci.hide`, `ci.allow_failures`)
+* [CI](#ci) (`ci.require`, `ci.hide`, `ci.allow_failures`)
 * [Custom tasks](#custom-tasks) (`restart`, `unlock`)
 * [Review Process](#review-process) (`monitor`, `checklist`, `checks`)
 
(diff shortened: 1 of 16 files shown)


rafaelfranca:
LGTM

gmalette:
:+1:! 
Should the readme explain these?

byroot:
Oh right. I'll update the doc. :facepalm:

app/models/commit.rb
```diff
@@ -173,6 +173,12 @@ def last_statuses
     @last_statuses ||= statuses.to_a.uniq(&:context).sort_by(&:context).presence || [UnknownStatus.new(self)]
   end
 
+  def status
+    visibles = visible_statuses
+    status = visibles.size > 1 ? StatusGroup.new(significant_status, visibles) : visibles.first
+    missing_statuses.empty? ? status : MissingStatus.new(status, missing_statuses)
+  end
```
etiennebarrie: Should this be handled in StatusGroup instead of right here?

etiennebarrie:
I'm a bit unclear on the responsibilities of the different classes but LGTM.

README.md
```diff
@@ -119,7 +119,7 @@ The settings in the `shipit.yml` file relate to the different things you can do
 * [Installing dependencies](#installing-dependencies) (`dependencies`)
 * [Deployment](#deployment) (`deploy`, `rollback`, `fetch`)
 * [Environment](#environment) (`machine.environment`)
-* [CI](#ci) (`ci.hide`, `ci.allow_failures`)
+* [CI](#ci) (`ci.require`, `ci.hide`, `ci.allow_failures`)
```
byroot: Writing the doc made me realize the previous ones were in the imperative form, so I renamed `ci.required` in `ci.require` for consistency

app/models/commit.rb
byroot: What do you mean by `this`?

etiennebarrie: The logic that makes the final status not successful if some status is missing.

byroot: Then no. Because if you have a single status you don't go through the group path.

Issue #512, opened 2015-12-21
After creating new deploy, redirected back to root url


dwradcliffe:
After I create a new deploy, I'm redirected back to the root url, with lots of query params. The deploy is still happening properly in the background, I just need to navigate to find it again.

Example: `/?action=show&controller=deploys&id=381&stack_id=rubygems%2Fshipit%2Fproduction`

byroot:
Hum, weird. That's very similar to https://github.com/Shopify/shipit-engine/pull/505.

I'll try to figure that out.

cc @rafaelfranca

byroot:
SO I confirm I can repro locally, and that it is like #505 due to Rails 4.2.5. Until we figure out the root cause you can lock Rails at 4.2.4 as a workaround.

byroot:
Fixed by #513

PR #513, opened 2016-01-04
Convert to an isolated engine for Rails 4.2.5 compat

byroot:
Rails 4.2.5 include a regression that messes with URL generation from the engine (#512, #505).
 
According to @rafaelfranca, isolating the namespace would help fix the issue. It was initially the plan but was reported because of the sheer amount of renaming and file moving to apply :/ 

This PR finally does it.

@rafaelfranca mind giving a quick look? I don't expect a deep review given the amount of changes.


app/controllers/shipit/api/base_controller.rb
```diff
@@ -0,0 +1,68 @@
+module Shipit
+  module Api
+    class BaseController < ActionController::Base
+      include Shipit::Engine.routes.url_helpers
```
rafaelfranca: Why this is needed? Should not the controller have access to all url helpers of the engine?

byroot: Indeed. It's already on master, I added it a while ago trying to debug something. I'll remove it.

PR #517, opened 2016-01-17
update readme to reference new namespace

perobertson:
After converting to the isolated engine in https://github.com/Shopify/shipit-engine/pull/513 the install migrations command stopped working.


byroot:
Nice catch, thanks!

PR #518, opened 2016-01-27
Multiple GH teams

baldowl:
Rationale: sometimes a single team is not enough.

I tried to be as backward compatible as possible and thus a simple `team` with a single value still works. While I was there, I made Rubocop happy again.

### DIFF
--- a/app/controllers/shipit/shipit_controller.rb
+++ b/app/controllers/shipit/shipit_controller.rb
@@ -33,7 +33,7 @@ module Shipit
     def force_github_authentication
       if current_user.logged_in?
         teams = Shipit.github_teams
-        if teams && teams.none? { |team| team.in?(current_user.teams) }
+        unless teams.empty? || current_user.teams.where(id: teams).exists?
           team_list = teams.map(&:handle).to_sentence(two_words_connector: ' or ', last_word_connector: ', or ')
           render text: "You must be a member of #{team_list} to access this application.", status: :forbidden
         end
--- a/app/controllers/shipit/shipit_controller.rb
+++ b/app/controllers/shipit/shipit_controller.rb
@@ -32,9 +32,10 @@ module Shipit
 
     def force_github_authentication
       if current_user.logged_in?
-        team = Shipit.github_team
-        if team && !current_user.in?(team.members)
-          render text: "You must be a member of #{team.handle} to access this application.", status: :forbidden
+        teams = Shipit.github_teams
+        if teams && teams.none? { |team| team.in?(current_user.teams) }
+          team_list = teams.map(&:handle).to_sentence(two_words_connector: ' or ', last_word_connector: ', or ')
+          render text: "You must be a member of #{team_list} to access this application.", status: :forbidden
         end
       else
         redirect_to Shipit::Engine.routes.url_helpers.github_authentication_path(origin: request.original_url)
--- a/lib/shipit.rb
+++ b/lib/shipit.rb
@@ -103,8 +103,9 @@ module Shipit
     secrets.host.presence
   end
 
-  def github_team
-    @github_team ||= github_oauth_credentials['team'] && Team.find_or_create_by_handle(github_oauth_credentials['team'])
+  def github_teams
+    teams = (Array(github_oauth_credentials['team']) + Array(github_oauth_credentials['teams'])).sort.uniq
+    @github_teams ||= teams.any? && teams.map { |t| Team.find_or_create_by_handle(t) }
   end
 
   def github_oauth_id
--- a/lib/shipit.rb
+++ b/lib/shipit.rb
@@ -105,7 +105,7 @@ module Shipit
 
   def github_teams
     teams = (Array(github_oauth_credentials['team']) + Array(github_oauth_credentials['teams'])).sort.uniq
-    @github_teams ||= teams.any? && teams.map { |t| Team.find_or_create_by_handle(t) }
+    @github_teams ||= teams.map { |t| Team.find_or_create_by_handle(t) }
   end
 
   def github_oauth_id
--- a/test/controllers/stacks_controller_test.rb
+++ b/test/controllers/stacks_controller_test.rb
@@ -54,11 +54,12 @@ module Shipit
       assert_redirected_to '/github/auth/github?origin=http%3A%2F%2Ftest.host%2F'
     end
 
-    test "current_user must be a member of Shipit.github_team" do
-      Shipit.stubs(:github_team).returns(shipit_teams(:cyclimse_cooks))
+    test "current_user must be a member of at least a Shipit.github_teams" do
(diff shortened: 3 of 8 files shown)


db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
```
pallan: Why are there so many schema changes?

.rubocop.yml
```diff
@@ -35,7 +35,7 @@ Style/DoubleNegation:
   Enabled: false
 
 Metrics/LineLength:
-  Max: 120
+  Max: 127
```
gmalette: Instead of this, you can split the culprit line into multiple lines

lib/shipit.rb
```diff
@@ -103,8 +103,9 @@ def host
     secrets.host.presence
   end
 
-  def github_team
-    @github_team ||= github_oauth_credentials['team'] && Team.find_or_create_by_handle(github_oauth_credentials['team'])
+  def github_teams
+    teams = (Array(github_oauth_credentials['team']) + Array(github_oauth_credentials['teams'])).sort.uniq
+    @github_teams ||= teams.any? && teams.map { |t| Team.find_or_create_by_handle(t) }
```
gmalette: The `any?` here means that `@github_teams` can be `false`

test/controllers/stacks_controller_test.rb
```diff
@@ -54,11 +54,12 @@ class StacksControllerTest < ActionController::TestCase
...
 
-    test "current_user must be a member of Shipit.github_team" do
-      Shipit.stubs(:github_team).returns(shipit_teams(:cyclimse_cooks))
+    test "current_user must be a member of at least a Shipit.github_teams" do
+      session[:user_id] = shipit_users(:bob).id
+      Shipit.stubs(:github_teams).returns([shipit_teams(:shopify_developers), shipit_teams(:cyclimse_cooks)])
       get :index
       assert_response :forbidden
-      assert_equal 'You must be a member of cyclimse/cooks to access this application.', response.body
+      assert_equal 'You must be a member of an authorized team to access this application.', response.body
```
gmalette: Can you keep a test for `team` only, if we're going to support it?

app/controllers/shipit/shipit_controller.rb
```diff
@@ -32,9 +32,9 @@ def ensure_required_settings
 
     def force_github_authentication
       if current_user.logged_in?
-        team = Shipit.github_team
-        if team && !current_user.in?(team.members)
-          render text: "You must be a member of #{team.handle} to access this application.", status: :forbidden
+        teams = Shipit.github_teams
+        if teams && teams.none? { |t| current_user.in?(t.members) }
+          render text: "You must be a member of an authorized team to access this application.", status: :forbidden
```
gmalette: Why not list the authorized teams?

app/controllers/shipit/shipit_controller.rb
```diff
@@ -32,9 +32,9 @@ def ensure_required_settings
 
     def force_github_authentication
       if current_user.logged_in?
-        team = Shipit.github_team
-        if team && !current_user.in?(team.members)
-          render text: "You must be a member of #{team.handle} to access this application.", status: :forbidden
+        teams = Shipit.github_teams
+        if teams && teams.none? { |t| current_user.in?(t.members) }
```
gmalette: I don't think `t.members` are preloaded and so this will result in N queries for N teams. I think this query should be reversed

db/schema.rb
byroot: That's probably my fault. @baldowl can you discard the schema.rb changes from your PR?

baldowl: Ok.

.rubocop.yml
baldowl: I think that

```
      test "#create do not allow to set protected attributes" do
        post :create, delivery_url: 'https://example.com/hook',
                      events: %w(deploy rollback),
                      created_at: 2.months.ago.to_s(:db)
        Hook.last.created_at > 2.seconds.ago
      end
```

is way uglier than a long line, but I'll split it anyway.

lib/shipit.rb
baldowl: Yes: if there's no team, either we return false here or check for `github_teams` having returned an empty Array in `ShipitController#force_github_authentication`. I preferred the current code.

test/controllers/stacks_controller_test.rb
baldowl: I'll add explicit tests for `Shipit.github_teams` to check for no team, `team` and both `team` and `teams`.

app/controllers/shipit/shipit_controller.rb
baldowl: I didn't feel the need to expose internal configuration, but if you prefer the list of authorized teams, I'll re-add it.

app/controllers/shipit/shipit_controller.rb
baldowl: Fine. I'll add the missing association to `User`, but I don't think we should preload teams for users, though.

baldowl:
Almost all comments have been addressed. Rubocop is broken again.

.rubocop.yml
gmalette: you can put parens

```ruby
post(
  :create,
  delivery_url: https://example.com/hook,
  ...
)
```
or extract the params to a hash 
```ruby
params = {
  delivery_url: 'https://example.com/hook'
  ....
}
post :create, params
```

lib/shipit.rb
gmalette: I'd rather it be a single type. @byroot ?

app/controllers/shipit/shipit_controller.rb
```diff
@@ -32,9 +32,10 @@ def ensure_required_settings
 
     def force_github_authentication
       if current_user.logged_in?
-        team = Shipit.github_team
-        if team && !current_user.in?(team.members)
-          render text: "You must be a member of #{team.handle} to access this application.", status: :forbidden
+        teams = Shipit.github_teams
+        if teams && teams.none? { |team| team.in?(current_user.teams) }
```
gmalette: This still performs N queries. How about: 

```ruby
current_user.teams.where(id: teams).exists?
```

byroot: Sounds good

lib/shipit.rb
byroot: Yeah, let's return an empty array instead of false, combined with the other comment:

```ruby
>> Shipit::Team.where(id: []).exists?
  Shipit::Team Exists (0.4ms)  SELECT  1 AS one FROM `teams` WHERE 1=0 LIMIT 1
```

So that sounds good. (Yes we could avoid the query, but I prefer simpler code)

gmalette:
:ship:

PR #522, opened 2016-02-05
Concurrent deploys now need to be acknowledged with a checkbox

byroot:
### Context

If you attempt to deploy while another deploy is already ongoing (either because of a race condition, or because you bypassed the UI), you are warned by a banner:

capture d ecran 2016-02-05 a 15 55 57

### Problem

If the review step become a bit too lengthly, e.g multiple checklist items plus review graphs and a long list of commits, the banner can easily be outside of the screen because of the scrolling.

### Solution

The banner is repeated at the bottom, and the second one require you to acknowledges the risks of a concurrent deploy before proceeding. Not sure if the top banner should be removed or not.

capture d ecran 2016-02-05 a 16 00 47

cc @aaslamin 

@davidcornu @etiennebarrie @tjoyal for review please.

### DIFF
--- a/app/assets/javascripts/shipit/checklist.js.coffee
+++ b/app/assets/javascripts/shipit/checklist.js.coffee
@@ -1,9 +1,9 @@
 $document = $(document)
 
 toggleDeployButton = ->
-  $('.trigger-deploy').toggleClass('disabled btn--disabled', !!$(':checkbox[name=checklist]:not(:checked)').length)
+  $('.trigger-deploy').toggleClass('disabled btn--disabled', !!$(':checkbox.required:not(:checked)').length)
 
-$document.on('change', ':checkbox[name=checklist]', toggleDeployButton)
+$document.on('change', ':checkbox.required', toggleDeployButton)
 
 jQuery ($) ->
   toggleDeployButton()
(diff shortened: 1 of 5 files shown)


etiennebarrie:
I guess only at the bottom with the checkbox would make sense. Code LGTM.

The issue I see with the feature (not the PR) is that it's not reloading live, so it doesn't actually prevent concurrent deploys if you've already opened the new deploy view, and someone else starts a deploy after that.

byroot:
> The issue I see with the feature (not the PR) is that it's not reloading live, so it doesn't actually prevent concurrent deploys if you've already opened the new deploy view, and someone else starts a deploy after that.

It does: https://github.com/Shopify/shipit-engine/blob/558a8c9aa6399f44b5bbc2f9d8d1b37a9db114cf/test/controllers/deploys_controller_test.rb#L61-L68

etiennebarrie:
Oh cool it does prevent concurrent in that way. :ok_hand:
It's not live though, even the commit statuses don't update on that page :crying_cat_face:.

byroot:
It's not live indeed. But if someone triggered a deploy while you were reviewing yours, you'll be taken back to review and will have to acknowledge it. I don't think live update will be really worth the effort. This is not something that happens that much.

gmalette:
I think this makes perfect sense

tjoyal:
I don't have much context but the changes make sense.

No longer hidden "force" flag seems way better! :+1:

byroot:
I dropped the top banner to keep only the bottom one.

aaslamin:
:+1:

PR #523, opened 2016-02-10
Prevent running concurrent tasks unless they are explictly allowed to

byroot:
By design tasks were allowed to be triggered even though a deploy or another task was already running.

But in practice it's not always true, especially for `restart` which is the one we advertise in the README.

So this PR prevent concurrent tasks, just like concurrent deploys are prevented, and it can be bypassed by clicking a checkbox like for deploy.

You can also give your task the `allow_concurrency: true` attribute to both allow to trigger the task during a deploy, and to trigger a deploy while this task is running.

@rafaelfranca @etiennebarrie for review please.

cc @eapache since IIRC you complained about this behavior a couple times.

### DIFF
--- a/app/controllers/shipit/tasks_controller.rb
+++ b/app/controllers/shipit/tasks_controller.rb
@@ -24,7 +24,7 @@ module Shipit
     def create
       @definition = stack.find_task_definition(params[:definition_id])
 
-      if @definition.allow_concurrency? || params[:force] || !@stack.deploying?
+      if @definition.allow_concurrency? || params[:force] || !@stack.active_task?
         @task = stack.trigger_task(params[:definition_id], current_user)
         redirect_to [stack, @task]
       else
--- a/app/controllers/shipit/tasks_controller.rb
+++ b/app/controllers/shipit/tasks_controller.rb
@@ -22,8 +22,14 @@ module Shipit
     end
 
     def create
-      @task = stack.trigger_task(params[:definition_id], current_user)
-      redirect_to [stack, @task]
+      @definition = stack.find_task_definition(params[:definition_id])
+
+      if @definition.allow_concurrency? || params[:force] || !@stack.deploying?
+        @task = stack.trigger_task(params[:definition_id], current_user)
+        redirect_to [stack, @task]
+      else
+        redirect_to new_stack_tasks_path(stack, @definition)
+      end
     end
 
     def abort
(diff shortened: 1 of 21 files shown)


gmalette:
Should tasks not be concurrent one-another, or is it just against deploys?

byroot:
> Should tasks not be concurrent one-another, or is it just against deploys?

Sorry I wasn't clear. `Deploy` and `Rollback` are simply `Task` specializations in Shipit terms. So yes a running task prevent both a new deploy and a new task. And a running deploy prevent both a new task and a new deploy.

app/controllers/shipit/tasks_controller.rb
```diff
@@ -22,8 +22,14 @@ def show
     end
 
     def create
-      @task = stack.trigger_task(params[:definition_id], current_user)
-      redirect_to [stack, @task]
+      @definition = stack.find_task_definition(params[:definition_id])
+
+      if @definition.allow_concurrency? || params[:force] || !@stack.deploying?
```
gmalette: I was asking the previous question about tasks exclusion because `Stack#deploying?` is true only when a `Deploy` or a `Rollback` is running, not other tasks.

gmalette: Oh :facepalm: I just saw the change of definition below. Maybe worth renaming to `active_task` then?

byroot: Yeah I wondered about it. My though was that `deploying?` while not 100% accurate is relatively good. However I agree `active_deploy` should be renamed in `active_task`. Will do.

byroot: Done.

gmalette: I'm pretty sure `deploying?` will cause misconception problems to someone someday... just saying. Someone will assume it actually means deploying...

byroot: Ok, if you insist: d5f8540e2d4f998014ee0d0c7556953623d4376a

etiennebarrie:
Renaming is a breaking change though. Should we add aliases to keep backwards compat or bump the major when releasing?

byroot:
> bump the major when releasing?

It's not a public API, you can use it if you want but at your own risks.

Also we're in `0.x` which allow for breaking changes according to semantic versioning. Oh and also we never claimed we were following semantic versioning.

etiennebarrie:
:shipit:

eapache:
:+1: I like to complain ;)

fw42:
this is awesome :heart:

PR #524, opened 2016-02-16
Add variables to the task definitions

hkdsun:
As a first step towards being able to set environment variables when running tasks through shipit's API, task definitions need to be able to hold a whitelist of environment variables that a user defines in their `shipit.yml` file.

Later on, we will use this whitelist to reject requests that don't conform to the defined configuration.

cc @byroot @fw42

### DIFF
--- a/app/models/shipit/task_definition.rb
+++ b/app/models/shipit/task_definition.rb
@@ -15,7 +15,7 @@ module Shipit
       end
     end
 
-    attr_reader :id, :action, :description, :steps, :checklist
+    attr_reader :id, :action, :description, :steps, :checklist, :variables
     alias_method :to_param, :id
 
     def initialize(id, config)
@@ -23,6 +23,7 @@ module Shipit
       @action = config['action']
       @description = config['description'] || ''
       @steps = config['steps'] || []
+      @variables = task_variables(config['variables'] || [])
       @checklist = config['checklist'] || []
       @allow_concurrency = config['allow_concurrency'] || false
     end
@@ -37,9 +38,16 @@ module Shipit
         action: action,
         description: description,
         steps: steps,
+        variables: variables.map(&:to_h),
         checklist: checklist,
         allow_concurrency: allow_concurrency?,
       }
+    end
+
+    private
+
+    def task_variables(config_variables)
+      config_variables.map(&VariableDefinition.method(:new))
     end
   end
 end
--- a/test/dummy/db/seeds.rb
+++ b/test/dummy/db/seeds.rb
@@ -51,6 +51,9 @@ module Shipit
               "steps": [
                 "bundle exec cap $ENVIRONMENT deploy:restart"
               ],
+              "variables": [
+                {"name": "SAFETY_DISABLED", "title": "Set to 1 to do stuff", "default": "0"}
+              ],
               "checklist": [
                 "Hold on your butts",
                 "Eat some chips"
--- a/test/models/task_definitions_test.rb
+++ b/test/models/task_definitions_test.rb
@@ -9,6 +9,8 @@ module Shipit
         'description' => 'Restart app and job servers',
         'steps' => ['touch tmp/restart'],
         'allow_concurrency' => true,
+        'variables' => [{'name' => 'FOO', 'title' => 'Set to 0 to foo', 'default' => 1},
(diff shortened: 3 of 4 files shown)


byroot:
Look like the branch is outdated. Could you rebase please?

test/dummy/db/seeds.rb
```diff
@@ -51,6 +51,10 @@ module Shipit
               "steps": [
                 "bundle exec cap $ENVIRONMENT deploy:restart"
               ],
+              "variables": [
+                {"name": "SAFETY_DISABLED", "title": "Set to 1 to do stuff", "default": "0"},
+                {"name": "SAFETY_DISABLED", "title": "Set to 1 to do stuff", "default": "0"}
```
byroot: Why do you define the same variable twice?

test/models/task_definitions_test.rb
```diff
@@ -1,4 +1,5 @@
 require 'test_helper'
+require 'byebug'
```
byroot: :scissors:

test/unit/deploy_spec_test.rb
```diff
@@ -249,6 +249,34 @@ class DeploySpecTest < ActiveSupport::TestCase
       assert_equal ['bundle exec foo'], definition.steps
     end
 
+    test "task definitions returns an array of VariableDefinition instances" do
+      @spec.expects(:load_config).returns('tasks' =>
+        {'restart' =>
```
byroot: The indentation is a bit inconsistent here. Also rubocop is complaining about it.

test/models/task_definitions_test.rb
```diff
@@ -29,6 +31,8 @@ class TaskDefinitionsTest < ActiveSupport::TestCase
         steps: ['touch tmp/restart'],
         checklist: [],
         allow_concurrency: true,
+        variables: [{'name' => 'FOO', 'title' => 'Set to 0 to foo', 'default' => 1, 'value'=>nil},
+          {'name' => 'BAR', 'title' => 'Set to 1 to bar', 'default' => 0, 'value'=>nil}]
```
gmalette: spacing inconsistent here too

app/models/shipit/task_definition.rb
```diff
@@ -37,9 +38,14 @@ def as_json
...
         description: description,
         steps: steps,
+        variables: variables.map(&:to_h),
         checklist: checklist,
         allow_concurrency: allow_concurrency?,
       }
     end
+
+    def task_variables(config_variables)
+      config_variables.map(&VariableDefinition.method(:new))
```
gmalette: I personally don't like using the lambda conversion style because it's harder to understand, and longer than `config_variables.map{ |v| VariableDefinition.new(v) }` but...

byroot: I kinda like it, and @hkdsun probably copied from my code somewhere else: https://github.com/Shopify/shipit-engine/blob/6e10e148c328f9c27a77c3b653234149f5e9a0ca/app/models/shipit/deploy_spec.rb#L66

hkdsun: yup that's what I did

test/unit/deploy_spec_test.rb
hkdsun: I think it's looking better now

hkdsun:
@fw42 is on vacation so thanks @gmalette for also reviewing this :+1:

gmalette:
Is there value in merging this without the 2nd part? Code LGTM

hkdsun:
I would like to wait till both parts are cleared before merging them since they're dependent on each other.

@byroot and I discussed this and decided that having two PRs makes more sense

byroot:
Yeah, we agreed on the outcome already so no problem for shipping small independent changes even though they don't change anything on their own. :shipit:

app/models/shipit/task_definition.rb
```diff
@@ -37,9 +38,14 @@ def as_json
         action: action,
         description: description,
         steps: steps,
+        variables: variables.map(&:to_h),
         checklist: checklist,
         allow_concurrency: allow_concurrency?,
       }
     end
+
+    def task_variables(config_variables)
```
hkdsun: I'm going to make this method private if there are no objections

hkdsun:
@byroot cool. I'll :ship: this in that case

PR #525, opened 2016-02-16
Allow accepting environment variables through Shipit's API

hkdsun:
This is a continuation of #524 in an effort to allow setting environment variables through shipit's API endpoints.

I achieved this by 
* having a whitelist of allowed variables per #524 
* ensuring all variables being set are checked before we trigger a deploy/rollback/task
* creating a small helper library for sanitizing environment variables similar to Rails' strong parameters

cc @fw42 @byroot @gmalette

### DIFF
--- a/app/controllers/shipit/api/base_controller.rb
+++ b/app/controllers/shipit/api/base_controller.rb
@@ -7,6 +7,7 @@ module Shipit
       include Paginable
 
       rescue_from ApiClient::InsufficientPermission, with: :insufficient_permission
+      rescue_from EnvironmentVariables::NotPermitted, with: :validation_error
       rescue_from TaskDefinition::NotFound, with: :not_found
 
       class << self
@@ -58,6 +59,10 @@ module Shipit
 
       def insufficient_permission(error)
         render status: :forbidden, json: {message: error.message}
+      end
+
+      def validation_error(error)
+        render status: :unprocessable_entity, json: {message: error.message}
       end
 
       def not_found(_error)
--- a/app/controllers/shipit/tasks_controller.rb
+++ b/app/controllers/shipit/tasks_controller.rb
@@ -25,7 +25,7 @@ module Shipit
       @definition = stack.find_task_definition(params[:definition_id])
 
       if @definition.allow_concurrency? || params[:force] || !@stack.active_task?
-        @task = stack.trigger_task(params[:definition_id], current_user)
+        @task = stack.trigger_task(params[:definition_id], current_user, env: task_params[:env])
         redirect_to [stack, @task]
       else
         redirect_to new_stack_tasks_path(stack, @definition)
@@ -49,6 +49,12 @@ module Shipit
 
     def stack
       @stack ||= Stack.from_param!(params[:stack_id])
+    end
+
+    def task_params
+      return {} unless params[:task]
+      @definition = stack.find_task_definition(params[:definition_id])
+      @task_params ||= params.require(:task).permit(env: @definition.variables.map(&:name))
     end
   end
 end
--- a/app/models/shipit/deploy.rb
+++ b/app/models/shipit/deploy.rb
@@ -13,7 +13,7 @@ module Shipit
     before_create :denormalize_commit_stats
     after_commit :broadcast_update
 
-    delegate :broadcast_update, to: :stack
+    delegate :broadcast_update, :filter_deploy_envs, to: :stack
 
     def build_rollback(user = nil, env: nil)
       Rollback.new(
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -55,13 +55,14 @@ module Shipit
       undeployed_commits_count > 0
     end
 
-    def trigger_task(definition_id, user)
+    def trigger_task(definition_id, user, env: nil)
       commit = last_deployed_commit
       task = tasks.create(
         user_id: user.id,
         definition: find_task_definition(definition_id),
         until_commit_id: commit.id,
         since_commit_id: commit.id,
+        env: filter_task_envs(definition_id, (env || {})),
       )
       task.enqueue
       task
@@ -74,7 +75,7 @@ module Shipit
         user_id: user.id,
         until_commit: until_commit,
         since_commit: since_commit,
-        env: env || {},
+        env: filter_deploy_envs(env || {}),
       )
       deploy.enqueue
       deploy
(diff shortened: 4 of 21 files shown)


app/controllers/shipit/api/base_controller.rb
```diff
@@ -7,6 +7,7 @@ class BaseController < ActionController::Base
       include Paginable
 
       rescue_from ApiClient::InsufficientPermission, with: :insufficient_permission
+      rescue_from EnvironmentVariables::NotPermitted, with: :insufficient_permission
```
byroot: Hum, IMO non whitelisted env var should return a `422`. It's basically a validation error.

lib/shipit/environment_variables.rb
```diff
@@ -0,0 +1,36 @@
...
+    private
+
+    def initialize(env)
+      @env = env
+    end
+
+    def sanitize_env_vars(variable_definitions)
+      allowed_variables = variable_definitions.map(&:name)
+
+      all_envs = @env.group_by { |k, _| allowed_variables.include?(k) }
```
byroot: `Enumerable#partition` is probably a better method here: http://ruby-doc.org/core-2.3.0/Enumerable.html#method-i-partition

app/models/shipit/deploy.rb
```diff
@@ -22,7 +22,7 @@ def build_rollback(user = nil, env: nil)
         parent_id: id,
         since_commit: stack.last_deployed_commit,
         until_commit: until_commit,
-        env: env || {},
+        env: filter_deploy_envs(env || {}),
```
byroot: Let's keep rollbacks out for now. I need to think about it. There is a case for rollbacks re-using the env of the deploy they rollback to. But not 100% sure yet.

app/models/shipit/stack.rb
```diff
@@ -74,7 +75,7 @@ def trigger_deploy(until_commit, user, env: nil)
         user_id: user.id,
         until_commit: until_commit,
         since_commit: since_commit,
-        env: env || {},
+        env: filter_deploy_envs(env || {}),
```
byroot: :+1: for env being filtered inside `trigger_deploy` etc.

lib/shipit/environment_variables.rb
```diff
@@ -0,0 +1,36 @@
...
+      allowed = all_envs[true].to_h
+      disallowed = all_envs[false].to_h
+
+      error_message = "Variables #{disallowed.keys} have not been whitelisted"
```
byroot: Might want to use `to_sentence` here: http://api.rubyonrails.org/classes/Array.html#method-i-to_sentence

app/controllers/shipit/api/base_controller.rb
hkdsun: Fixed. I wasn't entirely sure about this.

app/models/shipit/stack.rb
hkdsun: I spent a long time contemplating where the filter should go and I'm still open to discussion on it. 

I could have filtered it in [`TaskCommands`](https://github.com/Shopify/shipit-engine/blob/add_env_params_api/lib/shipit/task_commands.rb#L30) but that would have restricted any sort of divergence from the whitelist

hkdsun:
Pushed some fixes

lib/shipit/environment_variables.rb
```diff
@@ -0,0 +1,36 @@
...
+    def initialize(env)
+      @env = env
+    end
+
+    def sanitize_env_vars(variable_definitions)
+      allowed_variables = variable_definitions.map(&:name)
+
+      all_envs = @env.partition { |k, _| allowed_variables.include?(k) }
+      allowed = all_envs[0].to_h
+      disallowed = all_envs[1].to_h
```
byroot:
```ruby
allowed, disallowed = @env.partition ...
```

byroot:
One last nitpick. After that :+1: from me.

fw42:
LGTM. Can you squash your commits?

app/views/shipit/tasks/new.html.erb
```diff
@@ -13,13 +13,32 @@
 
   <%= render_checklist @task.checklist %>
 
-  <section class="submit-section">
-    <%= form_for @task, url: stack_tasks_path(@stack, definition_id: @task.definition.id) do |f| %>
+  <%= form_for @task, url: stack_tasks_path(@stack, definition_id: @task.definition.id) do |f| %>
```
hkdsun: I've added some changes after discovering a bug while doing functional tests locally but I squashed my commits by mistake so I'll comment on what they are.

I've added a form in the UI here so that the task can be triggered from shipit

byroot: Nice catch.

app/views/shipit/tasks/new.html.erb
```diff
@@ -13,13 +13,32 @@
...
+        <%= f.fields_for :env do |env_fields| %>
+          <% @task.definition.variables.each do |variable| %>
+            <p class="environment-variable">
+              <%= env_fields.text_field variable.name, value: variable.default %>
+              <%= env_fields.label variable.name, variable.title || variable.name %>
+            </p>
+          <% end %>
+        <% end %>
+      </section>
+    <% end %>
```
byroot: Could this section be factored out? The one in `deploys/new` must be very similar no?

app/controllers/shipit/tasks_controller.rb
```diff
@@ -25,7 +25,7 @@ def create
       @definition = stack.find_task_definition(params[:definition_id])
 
       if @definition.allow_concurrency? || params[:force] || !@stack.active_task?
-        @task = stack.trigger_task(params[:definition_id], current_user)
+        @task = stack.trigger_task(params[:definition_id], current_user, env: task_params[:env])
```
hkdsun: I had missed this embarrassingly enough, so the app controller wasn't respecting the variables through the UI :panda_face:

app/controllers/shipit/tasks_controller.rb
```diff
@@ -50,5 +50,11 @@ def task
     def stack
       @stack ||= Stack.from_param!(params[:stack_id])
     end
+
+    def task_params
```
hkdsun: ditto

app/views/shipit/tasks/new.html.erb
hkdsun: I could. I didn't because I couldn't see what else could use it in the future. But I guess a form renderer for an array of `VariableDefinition`s could be handy

hkdsun: `form_for_variables(variables, header)`

app/views/shipit/_variables.html.erb
```diff
@@ -0,0 +1,15 @@
+<% unless variables.empty? %>
```
hkdsun: done

byroot: Thanks!

hkdsun:
:ship:?

byroot:
Yes. :sheep: :it:

PR #528, opened 2016-02-17
Prevent race condition when continuous delivery is enabled

byroot:
### Context

CircleCI is a bit drunk and sometimes send the same status twice in a very short timeframe:

```ruby
=> [{:url=>
  "https://api.github.com/repos/Shopify/starscream/statuses/3c28bfdcc48fae50d0192aa7ce64742c9c89aedc",
 :id=>453068457,
 :state=>"success",
 :description=>"Your tests passed on CircleCI!",
 :target_url=>"https://circleci.com/gh/Shopify/starscream/43380",
 :context=>"ci/circleci",
 :created_at=>2016-02-16 19:19:18 UTC,
 :updated_at=>2016-02-16 19:19:18 UTC,
 ...
, {:url=>
  "https://api.github.com/repos/Shopify/starscream/statuses/3c28bfdcc48fae50d0192aa7ce64742c9c89aedc4742c9c89aedc",
 :id=>453068453,
 :state=>"success",
 :description=>"Your tests passed on CircleCI!",
 :target_url=>"https://circleci.com/gh/Shopify/starscream/43380",
 :context=>"ci/circleci",
 :created_at=>2016-02-16 19:19:18 UTC,
 :updated_at=>2016-02-16 19:19:18 UTC,
 ...
```

In those cases Shipit is subject to a race condition and can trigger two deploys for the same commit.
While there is clearly a problem on CircleCI side, Shipit should not be subject to such race condition.

### Solution

This PR move the code responsible for triggering automatic deploys in a background job, and `BackgroundJob::Unique` prevent any kind of concurrency.

@angelini @rafaelfranca @etiennebarrie @gmalette for review please.

### DIFF
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -80,6 +80,14 @@ module Shipit
       deploy
     end
 
+    def trigger_continuous_deploy
+      return unless deployable?
+      if commit = last_deployable_commit
+        return if commit.deployed?
+        trigger_deploy(commit, commit.committer)
+      end
+    end
+
     def async_refresh_deployed_revision
       FetchDeployedRevisionJob.perform_later(self)
     end
@@ -122,6 +130,10 @@ module Shipit
       else
         commits.first
       end
+    end
+
+    def last_deployable_commit
+      commits.reachable.preload(:statuses).find_each(batch_size: 10).find(&:deployable?)
     end
 
     def filter_visible_statuses(statuses)
--- a/test/models/commits_test.rb
+++ b/test/models/commits_test.rb
@@ -69,7 +69,10 @@ module Shipit
       @stack.deploys.destroy_all
 
       assert_difference "Deploy.count" do
-        @stack.commits.last.statuses.create!(state: 'success', context: 'ci/travis')
+        assert_enqueued_with(job: ContinuousDeliveryJob, args: [@stack]) do
+          @stack.commits.last.statuses.create!(state: 'success', context: 'ci/travis')
+        end
+        ContinuousDeliveryJob.new.perform(@stack)
       end
     end
 
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -94,3 +94,6 @@ Style/TrivialAccessors:
 
 Style/ExtraSpacing:
   AllowForAlignment: false
+
+Style/GuardClause:
+  Enabled: false
--- a/app/jobs/shipit/continuous_delivery_job.rb
+++ b/app/jobs/shipit/continuous_delivery_job.rb
@@ -1,0 +1,11 @@
+module Shipit
+  class ContinuousDeliveryJob < BackgroundJob
+    include BackgroundJob::Unique
+
+    queue_as :default
+
+    def perform(stack)
+      stack.trigger_continuous_deploy
+    end
+  end
+end
--- a/app/models/shipit/commit.rb
+++ b/app/models/shipit/commit.rb
@@ -141,10 +141,8 @@ module Shipit
     end
 
     def schedule_continuous_delivery
-      return unless state == 'success' && stack.continuous_deployment?
-      return unless stack.deployable?
-      return if already_deployed?
-      stack.trigger_deploy(self, committer)
+      return unless state == 'success' && stack.continuous_deployment? && stack.deployable?
+      ContinuousDeliveryJob.perform_later(stack)
     end
 
     def github_commit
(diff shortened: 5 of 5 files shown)


app/models/shipit/stack.rb
```diff
@@ -80,6 +80,14 @@ def trigger_deploy(until_commit, user, env: nil)
       deploy
     end
 
+    def trigger_continuous_deploy!
+      return unless deployable?
+      if commit = last_deployable_commit
+        return if commit.deployed?
+        trigger_deploy(commit, commit.committer)
```
angelini: I'm not super familiar with Ruby style, but if this method has a `!` because it modifies internal state, I would assume `trigger_deploy` would also need to be suffixed with a `!`.

byroot: Yeah right.

angelini: Why did you remove the `!` instead of adding it to `trigger_deploy`?

byroot: To stay consistent with the existing code. If I want to change `trigger_deploy` to `trigger_deploy!` I'll do it in another PR / commit.

etiennebarrie: From our [Ruby style guide](http://shopify.github.io/ruby-style-guide/):
> Never use a ! at the end of a method name if you have no equivalent method without the bang. Methods are expected to change internal object state; you don't need a bang for that. Bangs are to mark a more dangerous version of a method, e.g. save returns a bool in ActiveRecord, whereas save! will throw an exception on failure.

test/models/commits_test.rb
```diff
@@ -65,11 +65,15 @@ class CommitsTest < ActiveSupport::TestCase
     end
 
     test "updating state to success triggers new deploy when stack has continuous deployment" do
+      Hook.destroy_all
```
etiennebarrie: Why is this necessary?

app/models/shipit/status_group.rb
```diff
@@ -9,7 +9,7 @@ def initialize(significant_status, visible_statuses)
       @statuses = visible_statuses
     end
 
-    delegate :success?, :state, to: :significant_status
+    delegate :pending?, :success?, :error?, :failure?, :state, to: :significant_status
```
etiennebarrie: I don't see new calls to these methods, was it necessary to add the delegation? I'm not against it though.

etiennebarrie:
Would it be useful to add a test for the specific issue that PR solves? Like creating two statuses and checking only one job is enqueued?

app/models/shipit/stack.rb
byroot: > From our Ruby style guide:

That guideline is :hankey:, and it's for internal apps anyway so meh.

app/models/shipit/status_group.rb
byroot: Yeah, I figured they were missing during my interactive tests. That shouldn't be part of this PR. Removing.

test/models/commits_test.rb
byroot: It's not anymore. It was to prevent hooks from firing and generating network calls. But since then I changed the test to only run `ContinuousDeliveryJob`, so I'll :fire: that too.

byroot:
> Would it be useful to add a test for the specific issue that PR solves? Like creating two statuses and checking only one job is enqueued?

No 2 would be enqueued, the fix is based on the fact that those jobs are serialized via `BackgroundJob::Unique`. Concurrency tests are very hard and a PITA to maintain (see Semian). So nope, the current tests are enough IMO.

gmalette:
With this kind of check tho, it means it's possible (although very unlikely) that a race condition will happen as a result of a manual deploy + CI checks, given manual deploys are checked in the controller and CI using serialization. Although I don't think it's worth taking that case too seriously, should we instead serialize all deploy creation in the same manner to avoid this?

byroot:
> it's possible (although very unlikely) that a race condition will happen as a result of a manual deploy + CI checks

Indeed.

>  should we instead serialize all deploy creation in the same manner to avoid this?

I though about it. My personal conclusion is:
  -  if continuous deployment is enabled you are not very likely to trigger a deploy manually, and that the window is very short.
  - If we create the deploy record from background, we can't redirect you immediately to the deploy logs page, we have to poll or something for a while.

So IMO it's lots of added complexity for fixing a very unlikely problem. I'll definitely reconsider if it ever happens though.

rafaelfranca:
LGTM

byroot:
Any other concerns or should I :sheep: ?

angelini:
LGTM. We hit this problem again this morning:

double_deploy

Issue #23, opened 2014-03-01
Support new deployment API from GitHub


byroot:
See: http://developer.github.com/v3/repos/deployments/

/cc @gmalette

PR #543, opened 2016-03-03
Add support for GitHub's deployment API

byroot:
Fix: https://github.com/Shopify/shipit-engine/issues/23

capture d ecran 2016-03-03 a 14 55 30
See: https://github.com/byroot/junk/pull/9

So we're not really using it as it is intended to be, but it does work good enough to have the little squirrels in the PRs that link to the deploys and make it easy to know when a feature was actually shipped.

@rafaelfranca @gmalette @davidcornu @etiennebarrie mind taking a quick :eyes: ?

### DIFF
--- a/app/models/shipit/commit_deployment.rb
+++ b/app/models/shipit/commit_deployment.rb
@@ -1,0 +1,40 @@
+module Shipit
+  class CommitDeployment < ActiveRecord::Base
+    belongs_to :commit
+    belongs_to :task
+    has_many :statuses, class_name: 'CommitDeploymentStatus'
+
+    after_commit :schedule_create_on_github, on: :create
+
+    delegate :stack, to: :task
+
+    def create_on_github!
+      return unless commit.pull_request?
+
+      create_deployment_on_github!
+      statuses.order(id: :asc).each(&:create_on_github!)
+    end
+
+    def create_deployment_on_github!
+      return if github_id?
+
+      response = Shipit.github_api.create_deployment(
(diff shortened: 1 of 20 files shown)


dwradcliffe:
Even just having the squirrels is awesome, nice work!

byroot:
Forgot to mention something: The user that is referenced is the owner of the github token, which mean in our case it will be @shopify-shipit and not the actual deploy author.

I think we need to start to perform some GitHub actions on behalf of the users. I'll look into it but I believe it's better suited for a followup.

rafaelfranca:
Awesome!

app/models/shipit/commit_deployment.rb
```diff
@@ -0,0 +1,58 @@
...
+    has_many :statuses, class_name: 'CommitDeploymentStatus'
+
+    after_commit :schedule_create_on_github, on: :create
+
+    delegate :stack, to: :task
+
+    class << self
+      def append_status(task_status)
+        if github_status = GITHUB_STATUSES[task_status]
+          all.each do |deployment|
```
rafaelfranca: Took me a while to understand that this `all` would be scoped by the task id. Maybe we should add this method as an association extension to make clear that we expect it to be used only in the `Deploy` model?

byroot: Right. I'll do that :+1:

byroot: Fixed :ok_hand:

PR #545, opened 2016-03-03
Store users tokens

byroot:
They are stored encrypted, so that database backups stay safe.

This doesn't do anything just yet, but it will allow us to create the GitHub `Deployment` records on behalf of the deploy author so that it's his name that will appear on the PR.

@etiennebarrie @rafaelfranca @gmalette @davidcornu for review please.


app/models/shipit/user.rb
```diff
@@ -8,6 +8,8 @@ class User < ActiveRecord::Base
     has_many :commits, foreign_key: :committer_id, inverse_of: :committer
     has_many :tasks
 
+    attr_encrypted :access_token, key: Shipit.user_access_tokens_key
```
davidcornu: Should this be `github_access_token`?

app/controllers/shipit/github_authentication_controller.rb
```diff
@@ -20,8 +22,9 @@ def logout
     private
 
     def sign_in_github(auth)
-      user = Shipit.github_api.user(auth[:info][:nickname])
-      User.find_or_create_from_github(user).id
+      user = User.find_or_create_from_github(auth.extra.raw_info)
```
davidcornu: `raw_info` is the same as `auth[:info][:nickname]`?

app/models/shipit/user.rb
byroot: I don't it's necessary to prefix with `github_` here because the `User` model is literally github users.

app/controllers/shipit/github_authentication_controller.rb
byroot: Nope, it's the user resource from the GitHub api: https://github.com/intridea/omniauth-github/blob/45f2fc73d6d06f30863adac0e6aa112bcaaadf67/lib/omniauth/strategies/github.rb#L47

app/models/shipit/user.rb
davidcornu: Right but we could add a shipit-related `access_token` for CCMenu or something

byroot: Good point. I'll change it.

app/controllers/shipit/github_authentication_controller.rb
davidcornu: Oh so the output of `Shipit.github_api.user(auth[:info][:nickname])` is basically `auth.extra.raw_info`. Got it.

davidcornu:
Minor point on the column name. Otherwise LGTM.

app/controllers/shipit/github_authentication_controller.rb
byroot: Yup. This was actually a mistake, we were doing a useless extra request :/. So this will be slightly more webscale.

app/models/shipit/user.rb
byroot: Done!

rafaelfranca:
:+1:

PR #546, opened 2016-03-03
Stack merge status hooks

berkcaputcu:
@byroot @gmalette 

This PR adds a hook called `merge_status` that gets emitted after stack is updated. `merge_status` contains the stack parameters and last significant commit status as `merge_status` attribute.

`merge_status` can be any of these: `['locked', 'success', 'failure', 'pending']`
it falls back to the last deployed commit status when all of the undeployed commits are pending or when there is no undeployed commits.

One thing to mention is that it fires this hook after each commit to `stack`. An improvement could be that it only emits this hook after only if merge_status changes.

### DIFF
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -121,11 +121,8 @@ module Shipit
       if locked?
         'locked'
       else
-        significant_commits = commits.reachable.where('id >= ?', last_deployed_commit).order(id: :desc)
-        significant_statuses = significant_commits.map(&:significant_status)
-        return 'pending' if significant_statuses.empty? || significant_statuses.all?(&:pending?)
-
-        significant_commits.map(&:significant_status).detect { |status| !status.pending? }.simple_state
+        significant_commits = commits.reachable.newer_than(last_deployed_commit).order(id: :desc)
+        significant_commits.map(&:significant_status).find { |s| !s.pending? }.try!(:simple_state) || 'pending'
       end
     end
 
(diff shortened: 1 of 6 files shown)


app/models/shipit/stack.rb
```diff
@@ -116,6 +117,18 @@ def head
       commits.reachable.first.try!(:sha)
     end
 
+    def merge_status
+      if locked?
+        'locked'
+      else
+        significant_commits = commits.reachable.where('id >= ?', last_deployed_commit).order(id: :desc)
```
byroot: I'm not sure the last deployed commit actually matter. `id > ?` is what I'd use. So `...reacheable.newer_than(last_deployed_commit)...`

app/models/shipit/stack.rb
```diff
@@ -116,6 +117,18 @@ def head
...
+        significant_statuses = significant_commits.map(&:significant_status)
+        return 'pending' if significant_statuses.empty? || significant_statuses.all?(&:pending?)
+
+        significant_commits.map(&:significant_status).detect { |status| !status.pending? }.simple_state
```
byroot: I think the code would be simple without the guard clauses:

```ruby
significant_statuses.find { |s| !s.pending? }.try!(:simple_state) || 'pending'
```

byroot:
A couple very minor code style concerns. Otherwise LGTM.

berkcaputcu:
I updated this with better tests and it now considers `UnknownStatus` as well.

byroot:
:+1:

PR #547, opened 2016-03-03
Create Deployment and DeploymentStatus with the task author

byroot:
This is the completion of #545 and #543. 

Now Shipit attempt to create deployments and deployments statuses on behalf of the user that triggered the deploy.

any of @rafaelfranca @etiennebarrie @davidcornu @gmalette for review please.

### DIFF
--- a/app/models/shipit/anonymous_user.rb
+++ b/app/models/shipit/anonymous_user.rb
@@ -39,5 +39,9 @@ module Shipit
     def read_attribute_for_serialization(attr)
       public_send(attr)
     end
+
+    def github_api
+      Shipit.github_api
+    end
   end
 end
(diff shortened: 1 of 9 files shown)


byroot:
Ok, so I just tested it in production and it works as intended.

One quite big downside though: You can't make the difference between a deploy and a rollback unless you put your mouse on it and wait for the browser tooltip to appear. See https://github.com/byroot/junk/pull/10 (first one is a deploy, second one a rollback)

So I wonder if it wouldn't be best not to send the events for rollbacks. Thoughts?

rafaelfranca:
Yeah, I think we can skip the rollback events. LGTM.

byroot:
Done.

PR #549, opened 2016-03-04
Adds support for capitalization in repo_name.

JeremyAshton:
@byroot @baldowl

### DIFF
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -150,11 +150,11 @@ module Shipit
     end
 
     def repo_name=(name)
-      super(name.try(:downcase))
+      super(name.try!(:downcase))
     end
 
     def repo_owner=(name)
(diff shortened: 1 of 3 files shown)


byroot:
Hum. `repo_name` and `repo_owner` are case insensitive on the GitHub side, so we always store them in lower case to make matching simpler: https://github.com/Shopify/shipit-engine/blob/354ec755843839ce1e00fb84d73de19b1aad06a2/app/models/shipit/stack.rb#L226-L227

The best here would probably be:

```ruby
def repo_onwer=(owner)
  super(owner.try!(:downcase))
end
```

As for environment, we control it entirely so we could accept capitalization, but from my point of view it would only add confusion.

Could you provide a use case with this PR?

byroot:
I added a bit more test coverage. Thanks!

Issue #516, opened 2016-01-12
Navigating back to the commit/deploy list is annoying


eapache:
Minor thing that I have now hit frequently enough to bother filing a ticket over. I open shipit, and start a deploy which automatically happens in the same tab. There are then two cases:
- frequently while the deploy is still ongoing I'll want to check what other commits are waiting
- when the deploy is done I want to go back to the main page to e.g. start another deploy

In both cases the navigation is annoying: I have to scroll all the way back to the top to click on the tab, and the tab UI doesn't make a whole lot of sense because I'm not *on* any of the existing tabs when on the deploy screen.

Suggestion (caveat: I'm not in UX): float the three tabs into the header when scrolling on the deploy page. When a deploy is in-progress, add that as a fourth tab to all screens displaying the tab bar.

PR #550, opened 2016-03-04
Add a link to the stack in the task output header

byroot:
Fix: https://github.com/Shopify/shipit-engine/issues/516

When the banner isn't sticky yet the link is not displayed so it's not redondant with the page header:
capture d ecran 2016-03-04 a 13 27 43

Once you are scrolled the link appear:
capture d ecran 2016-03-04 a 13 27 32

@eapache hope it suits your need.


eapache:
:+1:

Issue #466, opened 2015-07-31
Differentiate ignored builds


davidcornu:
<img width="666" alt="screen shot 2015-07-31 at 10 31 19 am" src="https://cloud.githubusercontent.com/assets/325821/9009469/665f2690-376f-11e5-978c-da9f7a6d9e57.png">

Not sure what's going on here

gmalette:
Can you elaborate?

davidcornu:
If all the builds aren't complete, the build status should not be a green checkbox

gmalette:
[`ci/circleci`](https://github.com/Shopify/shopify/blob/f6a4aea84f7f5873656ee4410470b3864171ee39/shipit.yml#L3) is ignored from the build results, so we're left with a single green status

davidcornu:
> `ci/circleci` is ignored from the build results, so we're left with a single green status

Ah. So we should probably show that visually to avoid confusion. Maybe reduce the opacity of ignored builds or add `(ignored)` after the service name.

gmalette:
Good idea

PR #551, opened 2016-03-04
Reduce opacity of ignored statuses

byroot:
Fix: https://github.com/Shopify/shipit-engine/issues/466

capture d ecran 2016-03-04 a 13 48 02

@davidcornu @gmalette what do you think of my designer skillz?

### DIFF
--- a/app/views/shipit/statuses/_group.html.erb
+++ b/app/views/shipit/statuses/_group.html.erb
@@ -7,7 +7,7 @@
       <strong class="status-item__service"><%= group.description %></strong>
     </div>
     <% group.statuses.each do |status| %>
-      <div class="status-item status-item--<%= status.state %>">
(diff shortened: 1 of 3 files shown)


app/views/shipit/statuses/_group.html.erb
```diff
@@ -7,7 +7,7 @@
       <strong class="status-item__service"><%= group.description %></strong>
     </div>
     <% group.statuses.each do |status| %>
-      <div class="status-item status-item--<%= status.state %>">
+      <div class="status-item status-item--<%= status.state %> <%= :ignored if status.ignored? %>">
```
davidcornu: `'ignored'`?

davidcornu:
:shipit:

app/views/shipit/statuses/_group.html.erb
byroot: Meh, same thing.

PR #555, opened 2016-03-11
downcase organization name for incoming webhook to find the correct hook

perobertson:
We were having trouble with the teams webhooks and it turns out that its because of the capital letters in our organization name. When the `Shipit::Team` is created it downcases the organization, so when the organization webhook came in, it could not be found.

Here is the relevant part of the webhook payload
```
  "organization": {
    "login": "YounilityInc",
    ...
  }
```


byroot:
:facepalm: Thanks for the fix!

byroot:
I just released `0.8.5` with your fix.

perobertson:
excellent, thanks!

Issue #448, opened 2015-07-03
new stack UI needs a little love


dwradcliffe:
The repo field on the new stack page has line breaks that are not pretty:

<img alt="shipit 2015-07-03 11-10-42" src="https://cloud.githubusercontent.com/assets/118850/8501521/4a4fefb8-2174-11e5-8448-f79e28d0a99d.png">

davidcornu:
<img width="652" alt="screen shot 2015-07-06 at 9 36 34 am" src="https://cloud.githubusercontent.com/assets/325821/8523300/87a6e498-23c2-11e5-9879-e5041c5b4ddc.png">

The rest of the form needs some love too. IMHO the help text looks like a misplaced label and the checkbox for skipping CI should be on the other side.

dwradcliffe:
This was fixed a while ago.

PR #556, opened 2016-03-12
Clean up new stack UI

perobertson:
I cleaned up the new stack UI a little bit to address some of the comments in #448 

I did not come up with anything better for the help text of the deploy url, but its bothering me too so I may figure out something yet.

Heres what it looks like:

screenshot from 2016-03-12 09 39 38

@dwradcliffe @davidcornu


byroot:
Thanks a lot!

The CI is red because of the new rake version. I fixed that on master.

dwradcliffe:
Thanks!

PR #559, opened 2016-03-25
Use system sed absolutely in Mac OS X

takatoshiono:
hi

I use gnu-sed installed with homebrew on Max OS X. So when I deploy, the sed command is in error.

```
$ sed -i '' '/^ruby\s/d' Gemfile
pid: 36177
sed: can't read /^ruby\s/d: No such file or directory
```

I thought it is better to specify the absolute path.


dwradcliffe:
Doesn't that mean your PATH is wrong?

byroot:
> Doesn't that mean your PATH is wrong?

No, he's PATH properly point to the GNU sed in priority. What is wrong is the premise that `if /darwin/ =~ RUBY_PLATFORM` sed will be BSD sed.

So Yeah :+1: to this PR. The OS X path is only intended for development anyway.

takatoshiono:
Thank you for merge :smile:

PR #560, opened 2016-03-30
Integrate Clusterize.js to improve web page performance on huge log pages

byroot:
### Why?

Our current way of displaying logs is quite naive and works ok. But for big applications with lots of logs it tend to make the browser use tons of RAM and CPU. Just the sheer amount of DOM nodes make all the reflows / repaint etc terribly slow and expensive.

### How

This PR integrates https://github.com/NeXTs/Clusterize.js so that only the logs that are visible are actually rendered in the DOM.

One of the big downsides of it is that it breaks the native `cmd + F` browser functionality. As a result I had to also implement a search bar to compensate:

shipit-filter


byroot:
@davidcornu @rafaelfranca @gmalette care to take a look at this?

cc @garymardell

rafaelfranca:
:+1

app/assets/javascripts/task/search_bar.js.coffee
```diff
@@ -0,0 +1,39 @@
...
+
+  addEventListener: (type, handler) ->
+    @listeners(type).push(handler)
+
+  listeners: (type) ->
+    @eventListeners[type] ||= []
+
+  updateQuery: =>
+    @query = @$input.val()
+    @broadcastQueryChange()
```
davidcornu: Should we only broadcast if there was an actual change?

davidcornu:
:+1:

app/assets/javascripts/task/search_bar.js.coffee
byroot: Done.

davidcornu:
:clap:

garymardell:
Thanks @byroot

PR #561, opened 2016-04-01
Display tasks durations in UI and revamp commit UI for consistency

byroot:
### Deploys

Before:
capture d ecran 2016-04-01 a 14 37 51

After: 
capture d ecran 2016-04-01 a 14 26 59

### Commits

Before:
capture d ecran 2016-04-01 a 14 39 40

After:
capture d ecran 2016-04-01 a 14 42 22

NB: The full UTC timestamp is still visible if you go on the deploy timeline:

capture d ecran 2016-04-01 a 14 42 57

@davidcornu @etiennebarrie @gmalette @rafaelfranca thoughts on this?

### DIFF
--- a/app/models/shipit/duration.rb
+++ b/app/models/shipit/duration.rb
@@ -1,0 +1,28 @@
+module Shipit
+  class Duration
+    UNITS = {
+      's' => :seconds,
+      'm' => :minutes,
+      'h' => :hours,
+      'd' => :days,
+    }.freeze
+
+    def initialize(seconds)
+      @seconds = seconds
+    end
+
+    def to_i
+      @seconds.to_i
+    end
+
+    def to_s
+      seconds = to_i
+      days, seconds = seconds.divmod(1.day.to_i)
+      if days > 0
+        "#{days}d#{Time.at(seconds).utc.strftime('%Hh%Mm%Ss')}"
+      else
+        Time.at(seconds).utc.strftime('%Hh%Mm%Ss')[/[^0a-z]\w+/] || '0s'
+      end
+    end
+  end
+end
--- a/app/views/shipit/commits/_commit.html.erb
+++ b/app/views/shipit/commits/_commit.html.erb
@@ -5,12 +5,13 @@
     <span class="commit-title"><%= render_commit_message_with_link commit %></span>
     <p class="commit-meta">
       <span class="sha"><%= render_commit_id_link(commit) %></span>
-      <%= timeago_tag(commit.committed_at, force: true) %>
-      <span class="utc-timecode">at <%= commit.committed_at.strftime('%H:%M:%S') %> UTC</span>
       <% if commit.additions? && commit.deletions? %>
         <span class="code-additions">+<%= commit.additions %></span>
         <span class="code-deletions">-<%= commit.deletions %></span>
       <% end %>
+    </p>
+    <p class="commit-meta">
+      <%= timeago_tag(commit.committed_at, force: true) %>
     </p>
   </div>
   <div class="commit-actions">
--- a/app/assets/stylesheets/_pages/_commits.scss
+++ b/app/assets/stylesheets/_pages/_commits.scss
@@ -12,7 +12,7 @@
 }
 
 .commit, .task {
-  padding: 1rem 0;
+  padding: .75rem 0;
   display: flex;
   flex-direction: column;
 
--- a/app/models/shipit/task.rb
+++ b/app/models/shipit/task.rb
@@ -107,7 +107,7 @@ module Shipit
     end
 
     def duration
-      ended_at - started_at if duration?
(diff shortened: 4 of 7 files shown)


app/models/shipit/duration.rb
```diff
@@ -0,0 +1,31 @@
...
+      minutes, seconds = seconds.divmod(60)
+      hours, minutes = minutes.divmod(60)
+      days, hours = hours.divmod(24)
+
+      components = [days, hours, minutes, seconds].drop_while(&:zero?)
+      return "0s" if components.empty?
+
+      first, *rest = components.reverse.zip(UNITS.keys).reverse
+      "#{first.first}#{first.last}" + rest.map { |value, unit| "#{value.to_s.rjust(2, '0')}#{unit}" }.join
+    end
```
davidcornu: TIL there isn't a `exact_distance_of_time_in_words` method in `ActionView::Helpers::DateHelper`

byroot: Yeah... But even if there was it would be `about 2 minutes` instead of `2m12s` like I want it to.

app/models/shipit/duration.rb
```diff
@@ -0,0 +1,31 @@
...
+    def to_s
+      seconds = to_i
+      minutes, seconds = seconds.divmod(60)
+      hours, minutes = minutes.divmod(60)
+      days, hours = hours.divmod(24)
+
+      components = [days, hours, minutes, seconds].drop_while(&:zero?)
+      return "0s" if components.empty?
+
+      first, *rest = components.reverse.zip(UNITS.keys).reverse
```
davidcornu: Double reverse is a bit confusing

app/views/shipit/commits/_commit.html.erb
```diff
@@ -5,13 +5,14 @@
...
       <span class="sha"><%= render_commit_id_link(commit) %></span>
-      <%= timeago_tag(commit.committed_at, force: true) %>
-      <span class="utc-timecode">at <%= commit.committed_at.strftime('%H:%M:%S') %> UTC</span>
       <% if commit.additions? && commit.deletions? %>
         <span class="code-additions">+<%= commit.additions %></span>
         <span class="code-deletions">-<%= commit.deletions %></span>
       <% end %>
     </p>
+    <p class="commit-meta">
+      <%= timeago_tag(commit.committed_at, force: true) %>
```
davidcornu: `force: true`?

byroot: It was there before, see: https://github.com/jgraichen/rails-timeago/blob/542cdfb91b1a866a86457a8a44b92595021c865c/lib/rails-timeago/helper.rb#L24-L30

app/models/shipit/duration.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    def initialize(seconds)
+      @seconds = seconds
+    end
+
+    def to_i
+      @seconds.to_i
+    end
+
+    def to_s
```
davidcornu: Can we add some basic test coverage for this?

davidcornu:
Minor stuff. Otherwise LGTM

app/models/shipit/duration.rb
```diff
@@ -0,0 +1,31 @@
...
+      seconds = to_i
+      minutes, seconds = seconds.divmod(60)
+      hours, minutes = minutes.divmod(60)
+      days, hours = hours.divmod(24)
+
+      components = [days, hours, minutes, seconds].drop_while(&:zero?)
+      return "0s" if components.empty?
+
+      first, *rest = components.reverse.zip(UNITS.keys).reverse
+      "#{first.first}#{first.last}" + rest.map { |value, unit| "#{value.to_s.rjust(2, '0')}#{unit}" }.join
```
gmalette: If we don't expect days, we can use `strftime` instead

```ruby
Time.at(seconds).utc.strftime("%Hh%Mm%Ss")[/[^0a-z]\w+/]
```

Or you can use the same `divmod` for days only... just a suggestion but I think it's easier to read

gmalette:
I like this

app/models/shipit/duration.rb
etiennebarrie: We could avoid that by doing the `drop_while` after zipping. But I think tests covering it should be enough.

app/models/shipit/duration.rb
byroot: Test added.

app/models/shipit/duration.rb
byroot: Wow that's a very neat hack! I definitely don't expect duration bigger than a day, but somehow it bother me not to support it. I'm on the fence about this :s

byroot: Ok, I went with an hybrid solution!

PR #563, opened 2016-04-04
Raw output feature for deploys and tasks

as3richa:
Addresses concerns raised by @grollest et al. @byroot for review. To quickly test this in a working instance of Shipit, add `gem 'shipit-engine', git: 'git@github.com:Shopify/shipit-engine', branch: 'raw-output'` to the Gemfile and `bundle update`.

Because of #560, searching through Shipit deploy/task logs got a little clumsy for some use cases. This PR addresses that by adding a link in the deploy/task header to a plain text version of the logs.

For instance (note the link in the header):
screen shot 2016-04-04 at 3 45 05 pm

The link leads to the same page, but with a `.txt` extension. The entire log file is rendered verbatim as `text/plain`.
screen shot 2016-04-04 at 3 45 41 pm

This also applies to tasks:
screen shot 2016-04-04 at 3 46 07 pm

The rendering also behaves sanely when the task or deploy is still producing output.
screen shot 2016-04-04 at 3 46 14 pm

### DIFF
--- a/app/controllers/shipit/deploys_controller.rb
+++ b/app/controllers/shipit/deploys_controller.rb
@@ -13,6 +13,10 @@ module Shipit
     end
 
     def show
+      respond_to do |format|
+        format.html
+        format.text { render plain: @deploy.chunk_output }
+      end
     end
 
     def create
--- a/app/controllers/shipit/tasks_controller.rb
+++ b/app/controllers/shipit/tasks_controller.rb
@@ -19,6 +19,10 @@ module Shipit
 
     def show
       task
+      respond_to do |format|
+        format.html
+        format.text { render plain: @task.chunk_output }
+      end
     end
 
     def create
--- a/app/views/shipit/deploys/show.html.erb
+++ b/app/views/shipit/deploys/show.html.erb
@@ -5,7 +5,7 @@
     deploying
   <% end %>
   <span class="short-sha"><%= render_commit_id_link(@deploy.until_commit) %></span>
-  <%= timeago_tag(@deploy.created_at, force: true) %> at <%= @deploy.created_at %>
+  <%= timeago_tag(@deploy.created_at, force: true) %>
 <% end %>
 
 <%= render partial: 'shipit/tasks/task_output', locals: {task: @deploy} %>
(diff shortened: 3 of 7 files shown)


byroot:
Code LGTM. One very small concern is the space on the banner. Does it fit even with abort and abort + rollback buttons? Not sure what the smallest width we want to support :/

as3richa:
Entire screen width on a MBP:
screen shot 2016-04-04 at 4 30 58 pm

But if we shrink the window so the buttons can't sit next to the text:
screen shot 2016-04-04 at 4 31 27 pm

And if we shrink it enough that the text needs two lines:
screen shot 2016-04-04 at 4 31 35 pm

Not great, but IMO small screens were a little problematic in the first place - the buttons just kind of disappear :grin: 

Replace `(view raw output)` with `(raw)`, maybe?

byroot:
I think we can drop the full UTC timestamp. That would reclaim enough space.

Once it's done :ship:

as3richa:
Actually, do we need `height: 4rem` on `.deploy-banner`? Removing it makes only a tiny visual difference in a wide screen view, and it prevents the buttons from disappearing on smaller widths

byroot:
Hum, we might not need it. TBH I struggled on that layout and I some point I though I had to give those elements a fixed height: https://github.com/Shopify/shipit-engine/blob/c61dab151b4f1227b4403cb93310dde8c5a90d1e/app/assets/stylesheets/_pages/_deploy.scss#L57-L59

I might be wrong though and there might be a way to have this working without a fixed height.

as3richa:
What was the issue without it? I can't reproduce any visual flaw. Some screenshots:

Full width:
screen shot 2016-04-04 at 4 58 06 pm

Smaller:
screen shot 2016-04-04 at 4 58 14 pm

Really small:
screen shot 2016-04-04 at 4 58 22 pm

compared to the layout eating the buttons before.

If we don't want to think about CSS crap right now, removing the UTC timestamp does buy us a fair bit of space :+1: so it's ready to ship regardless

gmalette:
can you add tests please :heart:

as3richa:
Tests added. Merging this, but my suggestion about the CSS thing still stands if we want to revisit it at some point :+1:

PR #564, opened 2016-04-11
Added a js runtime

perobertson:
I needed to add a js runtime in order to run the tests locally on a Fedora box. It was unclear to me if this should actually be in the `gemspec` as a development dependency instead of in the `Gemfile`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,7 @@ gemspec
 gem 'mysql2'
 gem 'pg'
 gem 'sqlite3'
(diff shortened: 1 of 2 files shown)


byroot:
> It was unclear to me if this should actually be in the gemspec as a development dependency instead of in the Gemfile.

IMO it should be in the Gemfile. I don't want to force an ExecJS backend onto the host application. But I'm fine with forcing it for development.

Thanks!

Issue #558, opened 2016-03-24
[Question] Should the UI be changed to use a different css framework and/or templating engine?


perobertson:
Ive seen some discussion about using [bootstrap](http://getbootstrap.com/) and [primer](http://primercss.io/). Is this something that is still wanted?

There has also been some discussion about switching to [slim](http://slim-lang.com/) over erb. Is this something that is still wanted?

See #469 for some of the discussion that has occurred.

byroot:
> Ive seen some discussion about using bootstrap and primer. Is this something that is still wanted?

Yes. Design work keep being an issue in the project and being able to implement clean UI with little to no CSS would be a huge benefit.

> There has also been some discussion about switching to slim over erb

I'm in favor of it but it's a relatively more controversial change, so probably not.

perobertson:
> with little to no CSS
So you do not want to go down the bootstrap route or did you just mean no custom css being made?

byroot:
> So you do not want to go down the bootstrap route or did you just mean no custom css being made?

Sorry I realize it wasn't very clear.

We're quite happy with the current design, and we'd like to keep it.
But since it's homemade CSS it requires skills that I and most Shipit contributors don't have to add new UI.

So the idea behind #469 was to develop a custom Bootstrap theme so that we can keep the current look and feel of Shipit, but make it easy to implement new interfaces.

perobertson:
Gotcha. Ive been meaning to do some work on the UI side and just wanted to figure out the direction you were going with that.

So the next question is should a CDN be used instead of having shipit host the css?

byroot:
>  Ive been meaning to do some work on the UI side and just wanted to figure out the direction you were going with that.

And it's very much appreciated :) #448 was such a shame, thanks again for fixing it.

So if you want to do more UI work on Shipit:

  - If you think you have **a lot** of time to commit you can attempt the Bootstrap conversion, but beware it's not simple nor fast.
  - If you just have a few hours there and there, I have no problem merging PRs going forward with the current CSS.

> So the next question is should a CDN be used instead of having shipit host the css?

Since Shipit is a rails engine it's up to the host application to decide such thing. There is many ways to serve assets, and Sprockets let you chose the one you prefer.

byroot:
Closing this. Feel free to re-open if you have more questions.

PR #566, opened 2016-04-13
Bootstrap 4 conversion

perobertson:
@byroot after the discussion in #558 I decided to try out the bootstrap conversion and it does appear to be a decent chunk of work. The way I plan to tackle this is to use 2 different layouts, similar to the work in #469, with the addition of only using this layout if the feature toggle has been enabled in the `secrets.yml`. This way we can get individual pages working and merged in without a massive PR.

The feature toggle in the `secrets.yml` looks like this:
```yml
development:
  features:
    - bootstrap
```

What are your thoughts on this approach? If we go with this approach, Ill open up an issue that can keep track of which pages still need to get converted and tested.

Here is what the new stack page looks like so far:
screenshot from 2016-04-12 22-30-29


README.md
```diff
@@ -514,6 +514,17 @@ development:
...
 
+**`features`** is the list of development features you want to enable. These are typically not production ready.
+
+For example:
+
+```yml
+development:
+  features:
+    - bootstrap
+
```
byroot: I wouldn't bother documenting it since it's aimed at shipit developers only.

lib/shipit.rb
```diff
@@ -173,6 +173,10 @@ def revision
     end
   end
 
+  def feature_bootstrap?
+    secrets.features.include?('bootstrap')
```
byroot: The tests break here because `features` is `nil`. Might want to do `secrets.features.try!(:include?, 'bootstrap')`.

byroot:
The markup look sane and the feature toogle is a great idea. I'll try it first thing tomorrow.

Thanks a lot for this 😻

app/controllers/shipit/stacks_controller.rb
```diff
@@ -4,6 +4,7 @@ class StacksController < ShipitController
 
     def new
       @stack = Stack.new
+      render 'new_bs', layout: 'shipit_bs' if Shipit.feature_bootstrap? || params[:layout] == 'bootstrap'
```
byroot: I think we could factorize this with [`prepend_view_path`](http://api.rubyonrails.org/classes/ActionView/ViewPaths/ClassMethods.html#method-i-prepend_view_path).

It would allow to make new versions of the view without having to add these conditionals in every actions. I'll take a stab at this tomorrow.

byroot:
So I PRed the automatic view and layout selection: https://github.com/perobertson/shipit-engine/pull/1

byroot:
Thanks again!

PR #569, opened 2016-04-18
First commit for deploy needs to be a commit, not a module

jules2689:
What this does
---
- First time you deploy the since commit cannot be a module.

cc @byroot


byroot:
I fixed it a bit differently in https://github.com/Shopify/shipit-engine/commit/b1042d120e2103388941b3983c6fbd7322fa166d, but thanks a lot for the report.

Issue #562, opened 2016-04-01
shipit meltdown


teu:
Hi,

we experienced a total shipit meltdown today, for no apparent reason (servers runs only shipit) it stopped deploying anything.

One of our deployments is hanging in a constant state of running, when we tried to abort, we are getting:
```
Abort: failed, PID unknown
```

Tried:
- restarting sidekiq
- flushing redis
- restarting passenger app (runs apache/passenger)
- restarting apache

Another stack hangs completely on a commit saying "pending", none of the above also helped. Tried to manually refresh statuses, also no change.

Aside removing the stacks and redoing them, is there any procedure you could provide for such occasions?

Thanks

byroot:
What likely happened here is that sidekiq died without marking those deploys as completed, so now Shipit won't trigger any other deploy because it thinks there is one on the way.

It happened to us once where I had to `docker kill` a background job. It so rare that I never added anything in the UI to fix that state, but we likely should.

In the meantime what you can do to fix the state:

```
$ bundle exec rails console production
> Shipit::Task.active.where('created_at < ?', 1.hour.ago).each(&:error!)
```

teu:
Hi byroot,

excellent, thanks. That seems to have helped with the running one, but not with the pending commit, is that another type of background job?

byroot:
Oh sorry I missed the part about the pending commit. For that there is 2 possiblities:

  - Either that commit is still marked as `pending` on the GitHub side. You can check this via their API: https://developer.github.com/v3/repos/statuses/#list-statuses-for-a-specific-ref
  - If it's not pending on GitHub, then Shipit must fail to fetch the new status, in that case check the sidekiq UI for failed jobs (`RefreshStatusJob` / `GitHubSyncJob`).

teu:
Might sounds silly, but what sidekiq UI ?

teu:
Actually it helped, the second problem was also a deployment job, however wasn't an hour old when I first ran the job. Great support as always! Thanks @byroot

Issue #520, opened 2016-01-29
Commands timing out can sometimes leave tasks in running state


byroot:
That exception is supposed to be rescued and the task marked as errored

```ruby
Errno::EIO: Input/output error @ io_getpartial - /dev/pts/0
  /artifacts/ruby/2.1.0/bundler/gems/shipit-engine-8007264631ba/lib/shipit/command.rb:117:in `readpartial'
  /artifacts/ruby/2.1.0/bundler/gems/shipit-engine-8007264631ba/lib/shipit/command.rb:117:in `block (2 levels) in read_stream'
  /usr/lib/shopify-ruby/2.1.6-shopify2/lib/ruby/2.1.0/timeout.rb:91:in `block in timeout'
  /usr/lib/shopify-ruby/2.1.6-shopify2/lib/ruby/2.1.0/timeout.rb:35:in `block in catch'
  /usr/lib/shopify-ruby/2.1.6-shopify2/lib/ruby/2.1.0/timeout.rb:35:in `catch'
  /usr/lib/shopify-ruby/2.1.6-shopify2/lib/ruby/2.1.0/timeout.rb:35:in `catch'
  /usr/lib/shopify-ruby/2.1.6-shopify2/lib/ruby/2.1.0/timeout.rb:106:in `timeout'
  /artifacts/ruby/2.1.0/bundler/gems/shipit-engine-8007264631ba/lib/shipit/command.rb:126:in `with_timeout'
  /artifacts/ruby/2.1.0/bundler/gems/shipit-engine-8007264631ba/lib/shipit/command.rb:116:in `block in read_stream'
  /artifacts/ruby/2.1.0/bundler/gems/shipit-engine-8007264631ba/lib/shipit/command.rb:115:in `loop'
  /artifacts/ruby/2.1.0/bundler/gems/shipit-engine-8007264631ba/lib/shipit/command.rb:115:in `read_stream'
  /artifacts/ruby/2.1.0/bundler/gems/shipit-engine-8007264631ba/lib/shipit/command.rb:147:in `block in kill_and_wait'
  /usr/lib/shopify-ruby/2.1.6-shopify2/lib/ruby/2.1.0/timeout.rb:91:in `block in timeout'
  /usr/lib/shopify-ruby/2.1.6-shopify2/lib/ruby/2.1.0/timeout.rb:35:in `block in catch'
[...]

Issue #572, opened 2016-04-19
Fix test suite against Redis 3.3.0


byroot:
The redis client now uses `Timeout.timeout` and it breaks a big part of the test suite because we are stubbing it in a few places to test for timeout behavior. 

https://github.com/redis/redis-rb/compare/v3.2.2...v3.3.0#diff-20ca4377c0eb8c090a4e21893b08e8f9R85

For now I locked the Redis client we use in test to `3.2.2`, but I don't like it.

PR #575, opened 2016-04-27
Improve the abort process

byroot:
Fix: #562 
Fix: #520
Ref: #572 (we use `Timeout.timeout` less)

### Problems

- Until now `Task#abort!` assumed all the shipit processes (web and job) were on the same host. Web processed had to be able to send signal to processes spawned by the jobs.
- If for some reason a `PerformTaskJob` was lost (e.g. sidekiq restart timeout), the task be stuck in the `running` state without any proper way to be marked as failed, see https://github.com/Shopify/shipit-engine/issues/562 

### Solution

- `Shipit::Command` (the class responsible for spawning an monitoring sub processes) now use non-blocking IOs and allows to perform actions at regular intervals.
- That new ability is used by `PerformTaskJob` to report it's still alive back into Redis and to check if the deploy must be aborted (in another redis key).
- The TERM and KILL signals are now sent by the thread which spawned the subprocess, so Shipit an be deployed on multiple hosts or inside containers.
- If a `PerformTaskJob` haven't reported back in the last 15 seconds, `Task#abort!` will assume it's lost and mark the task as errored. While this should theorically only happen in extreme cases, it at least allow to recover from such bad state.

### ETA

I'll test this further on our production deployment of Shipit. I'll merge when I'm confident enough with that new code.

cc @angelini @jules2689 @teu @sgringwe because you suffered from those bugs recently.

### DIFF
--- a/app/models/shipit/task.rb
+++ b/app/models/shipit/task.rb
@@ -1,7 +1,10 @@
 module Shipit
   class Task < ActiveRecord::Base
+    PRESENCE_CHECK_TIMEOUT = 15
     ACTIVE_STATUSES = %w(pending running aborting).freeze
     COMPLETED_STATUSES = %w(success error failed flapping aborted).freeze
+
+    attr_accessor :pid
 
     belongs_to :deploy, foreign_key: :parent_id, required: false # required for fixtures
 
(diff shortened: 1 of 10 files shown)


app/models/shipit/task.rb
```diff
@@ -177,27 +181,45 @@ def finished?
...
-      Shipit.redis.set("task:#{id}:pid", pid, ex: 24.hour.to_i)
+    def alive?
+      Shipit.redis.get(status_key) == 'alive'
     end
 
-    def abort!(rollback_once_aborted: false)
-      target_pid = pid
-      return write("\nAbort: failed, PID unknown\n") unless target_pid.present?
+    def report_dead!
+      write("ERROR: Background job haven't reported back in #{PRESENCE_CHECK_TIMEOUT} seconds.")
```
jules2689: > Background job hasn't

app/models/shipit/task.rb
```diff
@@ -1,8 +1,12 @@
 module Shipit
   class Task < ActiveRecord::Base
+    PRESENCE_CHECK_INTERVAL = 5
```
jules2689: Why have 2 constants?

app/models/shipit/task.rb
```diff
@@ -177,27 +181,45 @@ def finished?
...
 
-    def abort!(rollback_once_aborted: false)
-      target_pid = pid
-      return write("\nAbort: failed, PID unknown\n") unless target_pid.present?
+    def report_dead!
+      write("ERROR: Background job haven't reported back in #{PRESENCE_CHECK_TIMEOUT} seconds.")
+      error!
+    end
+
+    def should_abort?
```
jules2689: I'm not following the logic here, can you explain? Can't this just be `Shipit.redis.get(abort_key).to_i > 0`?

app/models/shipit/task.rb
byroot: Thx.

app/models/shipit/task.rb
byroot: The idea if that it yield once every time you hit the abort button. Many command line utilities do a clean shutdown on first SIGTERM, and a qucik & dirty one on the second. I want every click of the abort button to be translated into a signal to the process.

Also passed the third click, we send SIGKILL instead of SIGTERM.

app/models/shipit/task.rb
byroot: It's a relic of a previous attempt at this. 🔥  thx.

byroot:
Worked fine both on my dev environment and in production.

PR #576, opened 2016-05-05
dev.yml for internal contributors

casperispoop:
@burke for review please. It's not as performant as it could be as I don't want to duplicate the logic from `bin/bootstrap`, but IMO it's good enough.

For external contributors the setup stays the same (`bin/bootstrap`) but they must have sqlite3 and redis locally.

Also I moved `mysql2` and `pg` in `ci` group and ignore it by default since they are really only useful on CI unless you are debugging a DB specific bug.

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -1,0 +1,35 @@
+# This file is for Shopify employees development environment.
+# If you are an external contributor you don't have to bother with it.
+name: shipit-engine
+
+up:
+  - homebrew:
+    - openssl
+    - sqlite
+  - ruby:
+      version: 2.1.8
+  - railgun
+  - bundler:
+      without: ci
+
+commands:
+  bootstrap: bin/bootstrap shopify
+  console: test/dummy/bin/rails console
+  server: test/dummy/bin/rails server -b 192.168.64.1 -p 55330
+  test: bin/rake
+
+open:
+  shipit: http://shipit-engine.localhost
+
+railgun:
+  image: dev:railgun-common-services-0.2.x
+  services:
+    redis: 6379
+    nginx: 80
+  ip_address: 192.168.64.85
+  memory: 2G
+  cores: 2
+  disk: 2G
+  proxy:
+    shipit-engine.localhost: 55330
+
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ tmp/*
 shared/*
 config/database.yml
 config/secrets.yml
+config/secrets.development.yml
 data/*
 coverage/*
 public/assets/*
@@ -28,6 +29,5 @@ test/dummy/log/*.log
 test/dummy/tmp/*
 test/dummy/data/*
 test/dummy/.sass-cache
-test/dummy/config/secrets.yml
 
 .byebug_history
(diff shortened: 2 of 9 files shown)


dev.yml
```diff
@@ -0,0 +1,35 @@
...
+      without: ci
+
+commands:
+  bootstrap: env REDIS_URL=redis://shipit-engine.railgun:6379 bin/bootstrap
+  console: pushd test/dummy; env REDIS_URL=redis://shipit-engine.railgun:6379 bin/rails console; popd
+  server: pushd test/dummy; env REDIS_URL=redis://shipit-engine.railgun:6379 bin/server -b 192.168.64.1 -p 55330; popd
+  test: env REDIS_URL=redis://shipit-engine.railgun:6379 bundle exec rake
+
+open:
+  shipit: https://shipit-engine.myshopify.io
```
dwradcliffe: I'd suggest using `shipit.dev` since this is straying quite far from myshopify.io intentions.

casperispoop: `.dev` is not recommended these days for dev env because it's a real TLD now. I think `shipit-engine.localhost` would be good. Not sure if it's possible with `dev`. @burke ?

burke:
Seems ok, I'd prefer `dev up` to automatically deal with db bootstrap/migrations and so on, but it doesn't support sqlite right now, and this a good enough most-of-the way solution :+1:

dev.yml
```diff
@@ -0,0 +1,35 @@
...
+
+railgun:
+  image: dev:railgun-common-services-0.2.x
+  services:
+    redis: 6379
+    nginx: 80
+  ip_address: 192.168.64.85
+  memory: 512M
+  cores: 1
```
burke: nginx won't like this, for some reason nginx-routing-modules fails to initialize unless there are at least 2 cores

burke: cc @pushrax in case that's a super easy fix ^

casperispoop: oh :/ Seemed totally overkill to give 2 cores and 2G to the VM. Especially since my new laptop is only 2 cores.

dev.yml
burke: OS X tries to resolve `.localhost` via Bonjour, I think, and it can cause mysterious delays. It's a really sad state of affairs, or maybe I just don't understand what's happening. Give it a shot though, I guess if it works then :shipit:

dev.yml
burke: Yeah, it's relatively low overhead per core, and the memory mostly stays paged out so it's not as crazy as it sounds.

dev.yml
casperispoop: > OS X tries to resolve .localhost via Bonjour, I think

Bonjour is `.local`. `.localhost` is unaffected AFAIK.

burke: I seem to remember having the same problem with `.localhost` but I could totally just be misremembering. Maybe I didn't want to use `.localhost` because it so strongly implied that it would resolve to `127.0.0.1`. Dunno.

dev.yml
pushrax: I'm not really sure why nginx-routing-modules would require more than 1 core – it runs within individual worker processes which are dictated by nginx itself. Is there an error message I can take a look at?

PR #579, opened 2016-05-17
Display link to sha range instead of last sha

berkcaputcu:
On the deploy/show page we only show a link to the last commit that's getting deployed.

I think instead we can show the sha range that's getting deployed.

@byroot


casperisfine:
Make sense. Thanks!

PR #581, opened 2016-05-24
Peek to the last deployed commit on merge_status

berkcaputcu:
Currently `merge_status` only looks at the undeployed commits. `merge_status` doesn't send a `success` webhook when there are no undeployed commits and shipit gets unlocked. 

In those cases, I think it's logical to look at the last deployed commit's status in order to see if the stack is mergeable or not.

/cc @Amos47 @casperisfine

### DIFF
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -144,6 +144,8 @@ module Shipit
         'locked'
       else
         significant_statuses = undeployed_commits.map(&:significant_status)
+        significant_statuses << last_deployed_commit.significant_status unless last_deployed_commit.blank?
+
         last_finalized_status = significant_statuses.reject { |s| %w(pending unknown).include?(s.state) }.first
         last_finalized_status.try!(:simple_state) || 'pending'
       end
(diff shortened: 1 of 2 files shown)


app/models/shipit/stack.rb
```diff
@@ -144,6 +144,8 @@ def merge_status
         'locked'
       else
         significant_statuses = undeployed_commits.map(&:significant_status)
+        significant_statuses << last_deployed_commit.significant_status unless last_deployed_commit.blank?
```
casperisfine: Hum, unless I'm mistaken `undeployed_commits` is in reverse order (most recent first). So you probably want to `unshift` here.

casperisfine: Nevermind. What I said is totally stupid.

casperisfine:
LGTM. :shipit:

app/models/shipit/stack.rb
berkcaputcu: That confused me as well at first. But since `<<` adds the element to the end of the array, we're fine.

PR #583, opened 2016-05-30
Fix "Enable notifications" not showing up

Darep:
Fixes the "Enable Notifications" thingie not showing up for new users.

The CSS class names in HTML & JS were different – line 24 in [shipit.js.coffee](https://github.com/Shopify/shipit-engine/blob/223ea28/app/assets/javascripts/shipit.js.coffee#L24) tries to find `.notifications` while the [HTML layout](https://github.com/Shopify/shipit-engine/blob/b248d57/app/views/layouts/shipit.html.erb#L16) declares the div with CSS class `.notification` (missing letter `s` at the end).

This PR also makes the CSS class name more verbose so as to prevent any mishaps with using `.notification` or `.notifications` somewhere else later in development.

### DIFF
--- a/app/assets/javascripts/shipit.js.coffee
+++ b/app/assets/javascripts/shipit.js.coffee
@@ -21,7 +21,7 @@ $(document).on 'click', '.disabled, .btn--disabled', (event) ->
   event.preventDefault()
 
 jQuery ->
-  $notificationNotice = $('.notifications')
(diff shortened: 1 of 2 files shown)


gmalette:
Oh nice! Thanks!

byroot:
👏

Darep:
👍

PR #586, opened 2016-06-09
Make desktop notifications banner dismissable

girasquid:
This adds a dismiss option to the desktop notifications banner:
screen shot 2016-06-09 at 4 40 32 pm

Clicking it hides the banner until the next time you visit; if we like this I can look into cookieing or something so that it stays out of your way for a while.

@casperisfine @pseudomuto


casperisfine:
> cookieing or something

LocalStorage should do.

LGTM :shipit:

girasquid:
I didn't think of that! Added, let me know what you think.

casperisfine:
👏

PR #591, opened 2016-07-07
update readme to specify rails version to use

perobertson:
@byroot I went to set up a new shipit instance not to long after rails 5 came out and ran into a couple bumps. First of all I had rails 5 installed, so the application got generated differently and the shipit-engine gem has a dependency on `~> 4.2.0`, so bundler failed. Additionally I have the rake 11 gem, so the after bundle step failed.


byroot:
Thanks.

PR #593, opened 2016-07-18
Upgrade to Rails 5

casperisfine:
This is a work in progress:

TODO and blockers:
  - Dependencies
    - [ ] `pubsubstub` (as well as sidekiq web) depends on `sinatra` and [Sinatra haven't published any Rack 2.0 compatible release yet](https://github.com/sinatra/sinatra/issues/1135). In the meantime master works.
    - [x] `explicit-parameters` is locked on Rails `4.x`, the upgrade is relatively easy.
    - [X] `test_after_commit` must be removed.
    - [x] `attr_encrypted` must be bumped to `3.0`, but [there is a major breaking change](https://github.com/attr-encrypted/attr_encrypted#upgrading-from-attr_encrypted-v2x-to-v3x)
  - Deprecations
    - [ ] The big one is `ActionController::TestCase HTTP request methods will accept only
keyword arguments in future Rails versions.`. We need to cleanup pretty much all the controller tests.
    - [ ] few minor ones
  - Failures
    - [x] Something about `AR#serialize` and `DeploySpec` and `AM::Dirty`, need to be investigated.


casperisfine:
@rafaelfranca @sgrif I'm having an issue with Rails 5 that seems specific to sqlite3: https://travis-ci.org/Shopify/shipit-engine/builds/146099643

When I run `db:schema:load` for the test environment, it crashes complaining that there is pending migrations :/ See: https://github.com/Shopify/shipit-engine/blob/5c38aebbbdd0715243dca397328e87438c5f9e31/script/cibuild#L8

And I can't reproduce it locally, so I'm not sure how to investigate this. Would that ring a bell to one of you ?

rafaelfranca:
It seems it is failing when running the tests in the automatic test schema maintenance but it was successful when loading the schema.

Have you tried to run the tests locally after the `db:schema:load` with the `development.sqlite3` file not present?

sgrif:
Doesn't ring any bells, but just as a sanity check I might try moving `rake test` to it's own command.

casperisfine:
So none of those worked. I also tried running `db:migrate` after schema load, and then it fails because the migrations are already applied. So it's as if the version table wasn't recorded properly.

I think I'll just allow sqlite3 failures for now :/

casperisfine:
Ok, the failure is ignored for now. Shipit can now work with both 4.2 and 5. When Sinatra does release a rack 2 compatible version, I'll make rails 5 the only supported version and cleanup the deprecations.

rafaelfranca:
If you can't reproduce locally it may be a version problem with sqlite3. I believe the version that runs in travis is really old.

casperisfine:
It was totally that, thank you @rafaelfranca. https://github.com/rails/rails/issues/24288#issuecomment-206011385 solves the issue.

I suppose `db:schema:load` does a mass insert in the `schema_versions` table which silently fail.

zzak:
Hello bros: http://zzak.io/log/2016-08-22-sinatra-two-point-oh.html

casperisfine:
@zzak ❤️

gmalette:
Haha I was just PR-ing https://github.com/sinatra/sinatra/commit/b115767785dc77ca7be52b02a0eeeecaf94b77a5

PR #596, opened 2016-07-28
Run review checks from ContinuousDeliveryJob and do not deploy if they fail

casperisfine:
As discussed briefly with @es @camilo and @csfrancis, it would be nice if continuous delivery would automatically back down when it's not a perfectly appropriate time to deploy.

Trying to implement such logic into Shipit would inevitably end with some very bloated and inflexible feature, but if we agree to delegate this intelligence to a script, then we kina already have the feature via `review.checks`, but it was limited to manual deploy.

This PR simply extend this feature to continuous delivery so that you can do something like:
```
review:
  checks:
    - curl --fail --silent https://monitoring.example.com/can_i_haz_deployz
```

My only concern with this, is the the end user UX. If the check fail and Shipit decide to delay the deploy, how does the end user is supposed to know and to understand?


gmalette:
@casperisfine what if we used the lock banner-ish-style (perhaps in a blue shade) to explain? perhaps the script could return the message to display or something similar?

csfrancis:
Would the review check only prevent continuous delivery from deploying? I'm assuming we can still force deploy even if the review check fails?

I think a banner would be fine.

casperisfine:
> Would the review check only prevent continuous delivery from deploying?

For manual deploys, users are warned but they can bypass it.

As for the banner, I'll see what I can do.

casperisfine:
So here's the banner:
capture d ecran 2016-08-02 a 14 43 03

And if you follow the link, it will display the commands output:
capture d ecran 2016-08-02 a 14 43 12

Looks good to you?

csfrancis:
That works for me.

PR #600, opened 2016-08-22
add locked_since column to stacks table

jpittis:
This adds a `locked_since` datetime column to the stacks table.
This will enable us to build analytics around how often certain stacks are locked.

I'm not sure if I'm doing all this migration stuff properly. Also made some small refactors.

@byroot

### DIFF
--- a/app/controllers/shipit/api/locks_controller.rb
+++ b/app/controllers/shipit/api/locks_controller.rb
@@ -10,7 +10,7 @@ module Shipit
         if stack.locked?
           render json: {message: 'Already locked'}, status: :conflict
         else
-          stack.update(lock_reason: params.reason, lock_author: current_user)
+          stack.lock(params.reason, current_user)
           render_resource stack
         end
       end
@@ -19,12 +19,12 @@ module Shipit
         requires :reason, String, presence: true
       end
       def update
-        stack.update(lock_reason: params.reason, lock_author: current_user)
+        stack.lock(params.reason, current_user)
         render_resource stack
       end
 
       def destroy
-        stack.update(lock_reason: nil, lock_author: nil)
+        stack.unlock
         render_resource stack
       end
     end
--- a/app/controllers/shipit/stacks_controller.rb
+++ b/app/controllers/shipit/stacks_controller.rb
@@ -53,6 +53,14 @@ module Shipit
       unless @stack.update(update_params)
         options = {flash: {warning: @stack.errors.full_messages.to_sentence}}
       end
+
+      reason = params[:stack][:lock_reason]
+      if reason.present?
+        @stack.lock(reason, current_user)
+      elsif @stack.locked?
+        @stack.unlock
+      end
+
       redirect_to(params[:return_to].presence || stack_settings_path(@stack), options)
     end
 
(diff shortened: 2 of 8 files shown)


app/controllers/shipit/api/locks_controller.rb
```diff
@@ -19,14 +19,29 @@ def create
...
 
       def destroy
-        stack.update(lock_reason: nil, lock_author: nil)
+        unlock
         render_resource stack
       end
+
+      private
+
+      def lock(reason, user)
```
casperisfine: This method should be in the model.

app/controllers/shipit/api/locks_controller.rb
```diff
@@ -19,14 +19,29 @@ def create
...
+        stack.update(lock_reason: reason, lock_author: user,
+                     locked_since: Time.current)
```
casperisfine: We shouldn't override the `locked_since` value unless it's `nil`.

app/controllers/shipit/api/locks_controller.rb
casperisfine: Also this isn't the only controller that can lock the stack. Humans can set a lock via StacksController#update: https://github.com/Shopify/shipit-engine/blob/83d46c8ecb5cfc6466ff9bba95e2cabcc3861ad0/app/controllers/shipit/stacks_controller.rb#L51-L57

app/controllers/shipit/api/locks_controller.rb
casperisfine: Oh, I see you handle it in the controller. It would be better if lock would handle this transparently.

jpittis:
I moved the lock / unlock logic to the model. Added 3 model tests to make sure it works.

stack_controller update still does not update the lock time. Should we have update set time.current or should update call in to `lock`? I prefer the later but it's kinda ugly too.

app/models/shipit/stack.rb
```diff
@@ -317,6 +317,16 @@ def locked?
       lock_reason.present?
     end
 
+    def lock(reason, user)
+      params = {lock_reason: reason, lock_author: user}
+      params[:locked_since] = Time.current if locked_since.nil?
+      update(params)
```
casperisfine: I'd use `update!` here, so that if raises if the lock can't be recorded.

app/models/shipit/stack.rb
```diff
@@ -317,6 +317,16 @@ def locked?
...
+    end
+
+    def unlock
+      update(lock_reason: nil, lock_author: nil, locked_since: nil)
```
casperisfine: Same here.

casperisfine:
> stack_controller update still does not update the lock time. Should we have update set time.current or should update call in to lock? I prefer the later but it's kinda ugly too.

Yeah, I'm not sure there is an optimal solution. I do prefer the later too. I think the best would be to remove `lock_reason` from `update_params` https://github.com/Shopify/shipit-engine/blob/83d46c8ecb5cfc6466ff9bba95e2cabcc3861ad0/app/controllers/shipit/stacks_controller.rb#L80 and to check for it separately and sent it to lock if it's present.

I think after that this PR will be good to ship.

jpittis:
Switched the updates to `update!` and #update now uses [lock and unlock](https://github.com/Shopify/shipit-engine/pull/600/files#diff-f1a38892b030c28340444e0f9555ef71R57).

app/controllers/shipit/stacks_controller.rb
```diff
@@ -53,6 +53,14 @@ def update
       unless @stack.update(update_params)
         options = {flash: {warning: @stack.errors.full_messages.to_sentence}}
       end
+
+      reason = params['stack']['lock_reason']
```
casperisfine: It's more idiomatic to use symbols as keys for params /nitpick.

casperisfine:
One last nitpick. Once fixed feel free to merge on green.

Issue #297, opened 2015-01-20
Treat Heroku as first-class citizen


gmalette:
Making Heroku first-class has many advantages. We could support a default set of tasks, checklist, fetch script, visualizations, etc. We could have interfaces to set the config values.

The current `shipit.yml` for Heroku projects looks like:

``` yml
dependencies:
  override: []

deploy:
  override:
    - push-to-heroku my-awesome-app
```

I suggest

``` yml
heroku:
  name: my-awesome-app

# heroku: my-awesome-app
```

PR #601, opened 2016-08-23
Add option to skip post-migration restart

SebastianSzturo:
For Reference: https://github.com/Shopify/service-disruptions/issues/497

Post-migration restarts for apps with preboot enabled can lead to problems when removing columns, because there could be more than 2 versions of an app running at the same time.

This PR adds a way to skip post-migration restarts by setting `POST_MIGRATION_RESTART=false`.

**I have no idea about bash, please make sure my code is right.**

--

@davidcornu @EiNSTeiN- @Hammadk @kevinhughes27 @shawnfrench

### DIFF
--- a/lib/snippets/push-to-heroku
+++ b/lib/snippets/push-to-heroku
@@ -13,7 +13,7 @@ function yellow() {
 HEROKU_APP=$1
 HEROKU_RELEASES=`heroku releases --app $HEROKU_APP 2>&1`
 GIT_COMMAND=${GIT_COMMAND:="git push git@heroku.com:$HEROKU_APP.git $SHA:master $GIT_FLAGS"}
-POST_MIGRATION_DEPLOY="${POST_MIGRATION_DEPLOY:=true}"
+POST_MIGRATION_DEPLOY=${POST_MIGRATION_DEPLOY:="true"}
 
 if [[ -z "${HEROKU_APP}" ]]; then
   red "Usage: $0 <app-name>"
@@ -33,7 +33,7 @@ if [[ $? -ne 0 ]]; then
   exit 1;
 fi
 
-HEROKU_LAST_GOOD_RELEASE=`echo "${HEROKU_RELEASES}" | head -n2 | tail -n1 | sed -e 's/^\(v[0-9]\+\).*/\1/'`
+HEROKU_LAST_GOOD_RELEASE=`echo "$HEROKU_RELEASES" | head -n2 | tail -n1 | sed -e 's/^\(v[0-9]\+\).*/\1/'`
 green "Current heroku release: ${HEROKU_LAST_GOOD_RELEASE}"
 
 green "Pushing to heroku."
@@ -63,7 +63,7 @@ if [[ "${MIGRATIONS}" != "" ]]; then
     exit 1
   fi
 
-  if [[ "${POST_MIGRATION_DEPLOY}" ]]; then
+  if [[ "${POST_MIGRATION_DEPLOY}" == "true" ]]; then
     green "Post-migration dyno restart."
     heroku restart --app $HEROKU_APP
   fi
--- a/lib/snippets/push-to-heroku
+++ b/lib/snippets/push-to-heroku
@@ -13,8 +13,9 @@ function yellow() {
 HEROKU_APP=$1
 HEROKU_RELEASES=`heroku releases --app $HEROKU_APP 2>&1`
 GIT_COMMAND=${GIT_COMMAND:="git push git@heroku.com:$HEROKU_APP.git $SHA:master $GIT_FLAGS"}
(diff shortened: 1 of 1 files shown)


lib/snippets/push-to-heroku
```diff
@@ -62,11 +62,12 @@ if [ "$MIGRATIONS" != "" ]; then
     exit 1
   fi
 
-  green "Post-migration dyno restart."
-  heroku restart --app $HEROKU_APP
+  if [ -z $NO_POST_MIGRATION_RESTART ]; then
```
casperisfine: Google's Shell styleguide is a good reference to write robust scripts. Notably in your case:
  - use `if [[ cond ]]` instead of `if [ cond ]`: https://google.github.io/styleguide/shell.xml?showone=Test,_%5B_and_%5B%5B#Test,_%5B_and_%5B%5B
  - and always interpolating variables `"${BLAH}"` instead of `$BLAH`: https://google.github.io/styleguide/shell.xml?showone=Testing_Strings#Testing_Strings

casperisfine:
One comment, appart from that 👍

lib/snippets/push-to-heroku
Hammadk: I like avoiding negated variables -- go with `if [[ ${POST_MIGRATION_DEPLOY} ]]` with a default truthy value set for `POST_MIGRATION_DEPLOY` in line 16

kevinhughes27:
I think this PR might be a good place to have a discussion on whether we should make not restarting after migrations the default or not. 

Technically we should be shipping migrations in a way that a restart is not required but we tend to be lazy with Heroku apps since we can often get away with it.

Forcing everyone to change might be hard but having more ways that our apps are configured is also not the best. What does everybody else think?

casperisfine:
> we should make not restarting after migrations the default or not.

This change would be backward incompatible (Shopify isn't the sole user of Shipit) so ideally we wouldn't change it.

Now this snippet is basically just an example of the kind of script you can embed in Shipit. So we can totally embed another `push-to-heroku` script in our own install.

Improving heroku support in Shipit is long overdue (https://github.com/Shopify/shipit-engine/issues/297) but it's not really on anyone roadmap. If someone want to go down that rabbit hole, I'd be happy to provide all the support needed.

davidcornu:
> This change would be backward incompatible (Shopify isn't the sole user of Shipit) so ideally we wouldn't change it.

Opt-out is probably fine for now

SebastianSzturo:
✅  Cleaned up the script according to comments.

@casperisfine @Hammadk: Could you please take another close  👀

lib/snippets/push-to-heroku
```diff
@@ -24,49 +25,50 @@ fi
...
-    red "The app $HEROKU_APP will now be reverted to its last working version $HEROKU_LAST_GOOD_RELEASE."
+    red "The app ${HEROKU_APP} will now be reverted to its last working version ${HEROKU_LAST_GOOD_RELEASE}."
     heroku releases:rollback $HEROKU_LAST_GOOD_RELEASE --app $HEROKU_APP
     red "Abort."
     exit 1
   fi
 
-  green "Post-migration dyno restart."
-  heroku restart --app $HEROKU_APP
+  if [[ "${POST_MIGRATION_DEPLOY}" ]]; then
```
casperisfine: This will always be true. Even if you pass `POST_MIGRATION_DEPLOY=false` or `0`, it will evaluate to a string, and a string unless empty is always truthy.

SebastianSzturo: Woops. So I will remove the string interpolation here, because we changed it from `NO_ POST_MIGRATION_RESTART` to `POST_MIGRATION_RESTART` with a default.

lib/snippets/push-to-heroku
```diff
@@ -63,7 +63,7 @@ if [[ "${MIGRATIONS}" != "" ]]; then
     exit 1
   fi
 
-  if [[ "${POST_MIGRATION_DEPLOY}" ]]; then
+  if [[ $POST_MIGRATION_DEPLOY ]]; then
```
casperisfine: Sorry but it still won't work. The best IMO is:

```bash
if [[ "${POST_MIGRATION_DEPLOY}" == "true" ]]
```

Hammadk:
👍

SebastianSzturo:
Will 🎩  on staging of one of the stores and merge.

SebastianSzturo:
🎩 'ed on app-store staging. Everything works as expected.

Is it okay to ignore the `travis-ci/push` failure for `Ruby: 2.2.3 Gemfile: gemfiles/Gemfile-rails5`?

casperisfine:
> Is it okay to ignore the travis-ci/push failure for Ruby: 2.2.3 Gemfile: gemfiles/Gemfile-rails5?

Yes, it seems like a flaky test. I'll look into it later. 🚢

PR #603, opened 2016-08-25
Allowing stack creation from API

lxfontes:
Seems like permission `write:stack` wasn't being used anywhere, so I used it instead of creating `create:stack`.

### DIFF
--- a/app/controllers/shipit/api/stacks_controller.rb
+++ b/app/controllers/shipit/api/stacks_controller.rb
@@ -17,9 +17,7 @@ module Shipit
         accepts :ignore_ci, Boolean
       end
       def create
-        @stack = Stack.new(params)
-        @stack.save!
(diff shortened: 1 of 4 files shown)


lxfontes:
cc @byroot @wfarr

camilo:
👍

app/controllers/shipit/api/stacks_controller.rb
```diff
@@ -16,6 +23,10 @@ def show
       def stack
         @stack ||= stacks.from_param!(params[:id])
       end
+
+      def create_params
```
byroot: For the API shipit uses explicit-parameters. You can look at other API controllers or webhook controller.

Also for creation parame shouldn't be nested under :stack

casperisfine:
Thx!

PR #604, opened 2016-09-19
fix xss in deploy url

lalithr95:
This PR fixes XSS in stack deploy_url parameter by validating the link in the model and sanitize in the view.

Review: @EiNSTeiN- @byroot


app/views/shipit/stacks/_header.html.erb
```diff
@@ -36,7 +36,7 @@
     </li>
     <% if stack.deploy_url.present? %>
       <li class="nav__list__item">
-        <%= link_to 'View website', stack.deploy_url, :target => '_blank', :class => 'deploy-url' %>
+        <%= sanitize link_to 'View website', stack.deploy_url, :target => '_blank', :class => 'deploy-url' %>
```
rafaelfranca: Why is this needs? The `link_to` should already escape the second parameter, unless we explicitly marked it as html safe.

test/controllers/stacks_controller_test.rb
```diff
@@ -166,5 +166,20 @@ class StacksControllerTest < ActionController::TestCase
       get :lookup, id: @stack.id
       assert_redirected_to stack_path(@stack)
     end
+
+    test "#create should not create stack with invalid deploy_url" do
```
gmalette: please don't use `should` in test names. be imperative about what it does. `#create does not create ...`

app/views/shipit/stacks/_header.html.erb
lalithr95: `" > < '` gets escaped but if we use a javascript directive, something like `Javascript:prompt(1)` will be rendered in the same way.

rafaelfranca: :+1: so can we just call `sanitize` in the argument not in the whole `link_to` method? That will make clear what is dangerous.

test/models/stacks_test.rb
```diff
@@ -591,5 +591,12 @@ def setup
       assert_equal new_reason, @stack.lock_reason
       assert_equal old_time, @stack.locked_since
     end
+
+    test "stack should contains valid deploy_url" do
```
gmalette: `s/should //`

app/views/shipit/stacks/_header.html.erb
lalithr95: Sanitize works on html tags, I tried to sanitize argument it didn't escape, 
http://stackoverflow.com/a/20413947

rafaelfranca: that is true.

test/models/stacks_test.rb
```diff
@@ -591,5 +591,12 @@ def setup
...
+      @stack.deploy_url = "Javascript:alert(0);//"
+      assert_not @stack.valid?
+      @stack.deploy_url = "https://shopify.com"
+      assert @stack.valid?
```
rafaelfranca: `assert_predicate`

test/models/stacks_test.rb
```diff
@@ -591,5 +591,12 @@ def setup
       assert_equal new_reason, @stack.lock_reason
       assert_equal old_time, @stack.locked_since
     end
+
+    test "stack should contains valid deploy_url" do
+      @stack.deploy_url = "Javascript:alert(0);//"
+      assert_not @stack.valid?
```
rafaelfranca: `assert_not_predicate`

byroot: I think you mean `refute_predicate` ;)

rafaelfranca: I think both are valid in a `ActiveSupport::TestCase` subclass https://github.com/rails/rails/blob/69256df843c40c6a8c9e9ecc91f4c59d7cbcd646/activesupport/lib/active_support/test_case.rb#L65

app/models/shipit/stack.rb
```diff
@@ -58,6 +58,7 @@ def lock_author=(user)
     validates :repo_owner, format: {with: /\A[a-z0-9_\-\.]+\z/}, length: {maximum: REPO_OWNER_MAX_SIZE}
     validates :repo_name, format: {with: /\A[a-z0-9_\-\.]+\z/}, length: {maximum: REPO_NAME_MAX_SIZE}
     validates :environment, format: {with: /\A[a-z0-9\-_\:]+\z/}, length: {maximum: ENVIRONMENT_MAX_SIZE}
+    validates :deploy_url, format: {with: URI.regexp(%w(http https))}, allow_blank: true
```
byroot: We have a couple stacks without a web frontend, so we set an `ssh://` url. I think we should allow it.

PR #605, opened 2016-09-19
Report github status if something is wrong

javierhonduco:
When GitHub's having some issue, Shipit may not properly work as it relies on lots of the GitHub's machinery such as webhooks, its REST API, etc.

This PR attempts to provide more information to the engineers by showing a small flash message whenever GitHub is in trouble indicating its current status and when was updated for the last time.

@byroot :)

### DIFF
--- a/app/controllers/shipit/shipit_controller.rb
+++ b/app/controllers/shipit/shipit_controller.rb
@@ -17,7 +17,7 @@ module Shipit
       :ensure_required_settings,
       :force_github_authentication,
       :set_variant,
-      :check_github_status
+      :check_github_status,
     )
 
     # Respond to HTML by default
--- a/app/controllers/shipit/shipit_controller.rb
+++ b/app/controllers/shipit/shipit_controller.rb
@@ -10,12 +10,14 @@ module Shipit
 
     helper Shipit::Engine.routes.url_helpers
     include Shipit::Engine.routes.url_helpers
+    include ActionView::Helpers::DateHelper
 
     before_action(
       :toogle_bootstrap_feature,
       :ensure_required_settings,
       :force_github_authentication,
       :set_variant,
+      :check_github_status
     )
 
     # Respond to HTML by default
@@ -65,6 +67,15 @@ module Shipit
 
       request.format = :html
       request.variant = :partial
+    end
+
+    def check_github_status
+      github_status = Rails.cache.read('github::status')
+      return if github_status.nil? || github_status[:status] == 'good'
+
+      flash[:warning] = "It seems that GitHub is having issues:
+        status: #{github_status[:status]} '#{github_status[:body]}'
+        updated: #{time_ago_in_words(github_status[:last_updated])} ago"
     end
   end
 end
--- a/app/controllers/shipit/shipit_controller.rb
+++ b/app/controllers/shipit/shipit_controller.rb
@@ -69,7 +69,7 @@ module Shipit
     end
 
     def check_github_status
-      github_status = Rails.cache.read('github::status')
+      github_status = Shipit::GithubStatus.status
       return if github_status.nil? || github_status[:status] == 'good'
 
       flash[:warning] = "It seems that GitHub is having issues:
--- a/app/controllers/shipit/shipit_controller.rb
+++ b/app/controllers/shipit/shipit_controller.rb
@@ -16,7 +16,6 @@ module Shipit
       :ensure_required_settings,
       :force_github_authentication,
       :set_variant,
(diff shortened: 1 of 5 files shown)


app/controllers/shipit/shipit_controller.rb
```diff
@@ -10,12 +10,14 @@ class ShipitController < ApplicationController
 
     helper Shipit::Engine.routes.url_helpers
     include Shipit::Engine.routes.url_helpers
+    include ActionView::Helpers::DateHelper
```
rafaelfranca: Please don't include helpers in controller. Use `self.class.helpers.time_ago_in_words`.

javierhonduco: Didn't know it was a bad pattern. Thanks :)

app/models/shipit/github_status.rb
```diff
@@ -0,0 +1,11 @@
+module Shipit
+  class GithubStatus
```
byroot: This doesn't need to be a class if it's never going to be instantiated.

app/controllers/shipit/shipit_controller.rb
```diff
@@ -66,5 +67,14 @@ def set_variant
       request.format = :html
       request.variant = :partial
     end
+
+    def check_github_status
+      github_status = Rails.cache.read('github::status')
```
byroot: Since you have a `GithubStatus` namespace, that cache read should be encapsulated there as well.

app/controllers/shipit/shipit_controller.rb
rafaelfranca: Because if that helper have dependency in another helper you will have to track it and also include here and sometimes it is really hard to track all dependencies of a helper.

javierhonduco: Thanks for the explanation :)

app/models/shipit/github_status.rb
```diff
@@ -0,0 +1,17 @@
...
+  module GithubStatus
+    CACHE_KEY = 'github::status'.freeze
+
+    class << self
+      def status
+        Rails.cache.read(CACHE_KEY)
+      end
+
+      def refresh_status
+        Rails.cache.fetch(CACHE_KEY) do
```
byroot: Ok, sorry if I missed that previously, but actually this won't work. If you don't specify an expiry, the cache will stay pretty much forever. So `refresh` should use `.write`.

javierhonduco: Oh! True, missed that! Thanks again! :)

app/controllers/shipit/shipit_controller.rb
```diff
@@ -16,6 +16,7 @@ class ShipitController < ApplicationController
       :ensure_required_settings,
       :force_github_authentication,
       :set_variant,
+      :check_github_status,
```
casperisfine: Last minute though, IMO instead of a before_action that set something in the flash. We'd be better to render a banner in the layout.

javierhonduco: Done!

app/views/layouts/shipit.html.erb
```diff
@@ -26,6 +26,23 @@
...
+      <div class="banner__inner wrapper">
+        <div class="banner__content">
+          <h2 class="banner__title">GitHub is having issues</h2>
+            "<i><%= github_status[:body] %></i>"
+            <% unless github_status[:last_updated].nil? %>
+              <%= time_ago_in_words(github_status[:last_updated]) %> ago
+            <% end %>
+        </div>
+
+        <a class="banner__dismiss">&times;</a>
```
casperisfine: I don't think your banner is dismissable ;)

app/views/layouts/shipit.html.erb
```diff
@@ -26,6 +26,23 @@
     </div>
   </div>
 
+  <% github_status = Shipit::GithubStatus.status
+     unless github_status.nil? || github_status[:status] == 'good' %>
+    <div class="banner github-status banner--red">
```
casperisfine: I'd go with orange.

casperisfine:
Almost there! Also would you mind posting a screenshot after those changes?

app/views/layouts/shipit.html.erb
javierhonduco: Oops, my bad.

javierhonduco:
This is how it looks like :)

image

casperisfine:
Awesome job. Thank you!

PR #613, opened 2016-10-20
Update rails version on installation instructions

mariovisic:
Running the given command gives me the error:

> The template [https://raw.githubusercontent.com/Shopify/shipit-engine/master/template.rb] could not be loaded. Error: You need at least Rails 5.0.0 to install shipit

Using the latest rails 5.0.0.1 works fine :)


byroot:
thanks

PR #614, opened 2016-10-21
Check for github body status before printing quotes

alexcoco:
@byroot

+ Check if `github_status[:body]` is `present?` before printing empty string with quotes
+ Use `if ... present?` instead of `unless ... nil?` for consistency and clarity

image


casperisfine:
I was working on it as well ;).

Thanks!

cc @javierhonduco

Issue #253, opened 2014-10-22
abort button didn't abort


ibawt:
When I was (trying) to deploy the ash changes, I noticed an error in the deploy but the abort button wouldn't do anything.  I had to get Dale to kill it manually.  This kind of sucks.

It was the job restarting that I was trying to abort.

@camilo @jduff

PR #616, opened 2016-11-10
Deploy to gke 

n1koo:
**What** 

This is a similar util as https://github.com/Shopify/shipit-engine/blob/master/lib/snippets/push-to-heroku but for deploying to `gke`.

`Usage: deploy-to-gke <app's namespace> <gcloud deployment> <gcloud project>` to match up the correct destination

We assume the repo has the configs in `config/kubernetes/#{environment}/` or overwrites the templatepath via `K8S_TEMPLATE_FOLDER`. 

For testing something like 

```ENVIRONMENT=environment K8S_TEMPLATE_FOLDER=/tmp/template REVISION=deadcode ./deploy-to-gke.rb foobar test-cluster shopify-tiers``` can be used

**Notes/TODO etc.** 

This is still a bit raw, but it works (:tophat:). We are in a somewhat hurry to get this out so i'd appreciate reviews and please nitpick on all the things. I'll be raging through them on monday morning.

@KnVerey  @byroot

### DIFF
--- a/lib/snippets/deploy-to-gke
+++ b/lib/snippets/deploy-to-gke
@@ -1,0 +1,156 @@
+#!/usr/bin/env ruby
+
+# Usage: deploy-to-gke <app's namespace> <gcloud deployment> <gcloud project>
+
+# Requires the following to be available on Shipit machine:
+#  - gcloud binary available on the shipit machine's path
+#  - kubectl binary available in the shipit machine's path
+#  - $GCLOUD_CREDENTIALS_DIR/my-gcloud-project-name.json must exist
+
+# Optionally, the following variables can be used to override script defaults:
+#  - K8S_TEMPLATE_FOLDER: location of Kubernetes files to deploy. Default is config/deploy/#{environment}.
+
+require 'open3'
+require 'securerandom'
+require 'erb'
+require 'json'
+require 'yaml'
+require 'shellwords'
+require 'tempfile'
+
+class GKEDeployment
+  class FatalDeploymentError < StandardError; end
+
+  def initialize(namespace:, gcloud_deployment:, gcloud_project:, environment:, current_sha:, key_dir:, template_folder: nil)
+    @namespace = namespace
+    @gcloud_deployment = gcloud_deployment
+    @gcloud_project = gcloud_project
+    @current_sha = current_sha
+    @key_file = "#{key_dir}/#{@gcloud_project}.json"
+    @template_path = './' + (template_folder || "config/deploy/#{environment}")
+
+    # Validate params + check existance of auth key and template(s)
+    enforce_required_params
+    preliminary_check
+
+    # Max length of podname is only 63chars so try to save some room by truncating sha to 8 chars
+    @id = current_sha[0...8] + "-#{SecureRandom.hex(4)}"
+  end
+
+  def run
+    authorize_gcloud
+    fetch_clusters.each do |cluster|
+      set_kubectl_cluster(cluster)
+      apply_all_templates
+    end
+  rescue FatalDeploymentError => error
+    print_error(error.message)
+    exit 1
+  end
+
+  def template_variables
+    {
+      'current_sha' => @current_sha,
+      'deployment_id' => @id,
+    }
+  end
+
+  private
+
+  def enforce_required_params
+    [@namespace, @gcloud_project, @gcloud_deployment, @current_sha, @key_file].each do |required_param|
+      raise ArgumentError, "#{required_param} is required" unless required_param && !required_param.empty?
+    end
+  end
+
+  def preliminary_check
+    raise FatalDeploymentError, "Project config missing at #{@key_file}" unless File.file?(@key_file)
+    raise FatalDeploymentError, "#{@template_path} doesn't exist" unless File.directory?(@template_path)
+    raise FatalDeploymentError, "#{@template_path} doesn't have files with postfix .yml or .yml.erb" unless Dir.entries(@template_path).select {|file| file =~ /\.yml(.erb)?$/}.size > 0
+  end
+
+  def apply_all_templates
+    found = 0
+    Dir.foreach(@template_path) do |file|
+      file_path = "#{@template_path}/#{file}"
+      if File.extname(file) == '.yml'
+        found += 1
+        run_command('kubectl', 'apply', '-f', file_path, "--namespace=#{@namespace}")
+      elsif File.extname(file) == '.erb'
+        found += 1
+        render_and_apply_template(file_path)
+      end
+    end
+    raise FatalDeploymentError, "No templates found in #{@template_path}" if found.zero?
+  end
+
+  def render_and_apply_template(file_path)
+    erb_template = ERB.new(File.read(file_path))
+    erb_binding = binding
+    template_variables.each do |var_name, value|
+      erb_binding.local_variable_set(var_name, value)
+    end
+    content = erb_template.result(erb_binding)
+
+    f = Tempfile.new(['kube_template', '.yml'])
+    f.write(content)
+    f.close
+    run_command('kubectl', 'apply', '-f', f.path, "--namespace=#{@namespace}")
+  ensure
+    f.unlink if f
+  end
+
+  def authorize_gcloud
+    status = run_command('gcloud', '-q', 'auth', 'activate-service-account', '--key-file', @key_file)
+    status = run_command('gcloud', '-q', 'config', 'set', 'project', @gcloud_project) if status
+    raise FatalDeploymentError, "Failed to set gcloud project #{@gcloud_project}" unless status
+  end
+
+  def set_kubectl_cluster(cluster)
+    cluster_name = cluster[0]
+    cluster_zone = cluster[1]
+    status = run_command('gcloud', '-q', 'container', 'clusters', 'get-credentials', cluster_name, '--zone', cluster_zone)
+    raise FatalDeploymentError, "Failed to set cluster #{cluster_name}/#{cluster_zone}" unless status
+  end
+
+  def fetch_clusters
+    result = query_deployments
+    result['resources'].each_with_object([]) do |resource, data|
+      next unless resource['type'] == 'container.v1.cluster'
+      properties = YAML.load(resource['finalProperties'])
+      data << [resource['name'], properties['zone']]
+    end
+  end
+
+  def run_command(*args)
+    puts Shellwords.join(args)
+    out, err, st = Open3.capture3(*args)
+    puts out
+    print_error(err) unless st.success?
+    st.success?
+  end
+
+  def print_error(msg)
+    puts "\033[0;31m#{msg}\033[0m"
+  end
+
+  def query_deployments
+    out, err, st = Open3.capture3('gcloud', '-q', 'deployment-manager', 'deployments', 'describe', @gcloud_deployment, '--format=json')
+    unless st.success?
+      print_error(err)
+      raise FatalDeploymentError, "Failed to fetch cluster with deployment #{@gcloud_deployment}"
+    end
+    JSON.parse(out)
+  end
+end
+
+deployment = GKEDeployment.new(
+  namespace: ARGV[0],
+  gcloud_deployment: ARGV[1],
+  gcloud_project: ARGV[2],
+  environment: ENV['ENVIRONMENT'],
+  template_folder: ENV['K8S_TEMPLATE_FOLDER'],
+  current_sha: ENV['REVISION'],
+  key_dir: ENV['GCLOUD_CREDENTIALS_DIR']
+)
+deployment.run

### CODE REVIEW

File: lib/snippets/deploy-to-gke.rb [hunk 1]
```diff
@@ -0,0 +1,137 @@
...
+  def set_kubectl_cluster(cluster)
+    cluster_name = cluster[0]
+    cluster_zone = cluster[1]
+    st = run_command("gcloud container clusters get-credentials #{cluster_name} --zone #{cluster_zone}")
+    raise FatalDeploymentError, "Failed to set cluster #{cluster_name}/#{cluster_zone}" unless st
+  end
+
+  def fetch_clusters
+    cmd = "gcloud deployment-manager deployments describe #{@gcloud_deployment} --format=json"
+    result = JSON.parse(`#{cmd}`)
```
shopify-admins: Review this line for a possible security issue: [Possible command injection](http://brakemanscanner.org/docs/warning_types/command_injection/).
in the following block of code:
```ruby
`#{"gcloud deployment-manager deployments describe #{@gcloud_deployment} --format=json"}`
```
this user-controlled input appears unsafe:
```ruby
@gcloud_deployment
```
Drop by #security on slack for help.

ibawt:
all the commands will need the '-q' to indicate that it is being run from a script so that we can rely on the error status.

lib/snippets/deploy-to-gke.rb
```diff
@@ -0,0 +1,136 @@
...
+#  - kubectl binary available in the shipit machine's path
+#  - GCLOUD_CREDENTIALS_DIR: location of JSON service account auth files, e.g. /home/username/.google_iam
+
+# Optionally, the following variables can be used to override script defaults:
+#  - K8S_TEMPLATE_FOLDER: location of Kubernetes files to deploy. Default is config/<environment>/.
+
+require 'open3'
+require 'securerandom'
+require 'json'
+require 'pry'
```
ibawt: fine for testing but shouldn't be in the end result

lib/snippets/deploy-to-gke.rb
```diff
@@ -0,0 +1,136 @@
...
+
+  def set_kubectl_cluster(cluster)
+    cluster_name = cluster[0]
+    cluster_zone = cluster[1]
+    st = run_command("gcloud container clusters get-credentials #{cluster_name} --zone #{cluster_zone}")
+    raise FatalDeploymentError, "Failed to set cluster #{cluster_name}/#{cluster_zone}" unless st
+  end
+
+  def fetch_clusters
+    cmd = "gcloud deployment-manager deployments describe #{@gcloud_deployment} --format=json"
```
ibawt: can we do this without a requirement on deployment manager?  Like the cluster to deploy to is just passed in via the shipit.yml somehow.

lib/snippets/deploy-to-gke.rb
```diff
@@ -0,0 +1,136 @@
...
+
+  def initialize(namespace:, gcloud_deployment:, gcloud_project:, environment:, current_sha:, template_folder: nil)
+    @namespace = namespace
+    @gcloud_deployment = gcloud_deployment
+    @gcloud_project = gcloud_project
+    @current_sha = current_sha
+    enforce_required_parmas
+    @id = current_sha + "-#{SecureRandom.hex(4)}"
+    @template_path = template_folder || "config/kubernetes/#{environment}/"
+    @template_path =  "./" + @template_path
```
KnVerey: Why not modify the line above instead?

lib/snippets/deploy-to-gke.rb
```diff
@@ -0,0 +1,136 @@
...
+  end
+
+  def apply_all_templates
+    unless Dir[File.expand_path("#{@template_path}", __FILE__)]
+      raise FatalDeploymentError, "No templates found in #{@template_path}"
+    end
+    Dir.foreach(File.expand_path("#{@template_path}", __FILE__)) do |file|
+      file_path = "#{@template_path}/#{file}"
+      if File.extname(file) == '.yml'
+        run_command("kubectl apply -f file_path --namespace=#{@namespace}")
```
KnVerey: Presumably `file_path` should be interpolated

lib/snippets/deploy-to-gke.rb
```diff
@@ -0,0 +1,136 @@
...
+      elsif File.extname(file) == 'yml.erb'
```
KnVerey:
```
[1] pry(main)> File.extname "foo.yml.erb"
=> ".erb"
```

lib/snippets/deploy-to-gke.rb
```diff
@@ -0,0 +1,136 @@
...
+        render_and_apply_template(render_template(file_path))
```
KnVerey: `render_template` doesn't seem to be a thing...

lib/snippets/deploy-to-gke.rb
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env ruby
+
+# Usage: deploy-to-gke <app's namespace> <gcloud deployment> <gcloud project>
+
+# Requires the following to be available on Shipit machine:
+#  - gcloud binary available on the shipit machine's path
+#  - kubectl binary available in the shipit machine's path
+#  - GCLOUD_CREDENTIALS_DIR: location of JSON service account auth files, e.g. /home/username/.google_iam
```
KnVerey: I think this is something you got from my gist, but the example is weird. :P
How about: "i.e. `$GCLOUD_CREDENTIALS_DIR/my-gcloud-project-name.json` must exist"

lib/snippets/deploy-to-gke.rb
KnVerey: I agree with Ian. The `shipit.yml` entry can be `deploy-to-gke #{namespace} #{project} #{cluster}`.

lib/snippets/deploy-to-gke.rb
```diff
@@ -0,0 +1,136 @@
...
+  class FatalDeploymentError < StandardError; end
+
+  def initialize(namespace:, gcloud_deployment:, gcloud_project:, environment:, current_sha:, template_folder: nil)
+    @namespace = namespace
+    @gcloud_deployment = gcloud_deployment
+    @gcloud_project = gcloud_project
+    @current_sha = current_sha
+    enforce_required_parmas
+    @id = current_sha + "-#{SecureRandom.hex(4)}"
+    @template_path = template_folder || "config/kubernetes/#{environment}/"
```
KnVerey: (also for line 11) The default I went with in my other PR is `config/deploy/#{env}`. We can change that if you think this is better, but we should choose the same thing.

lib/snippets/deploy-to-gke.rb
KnVerey: FWIW we can get the zone from `gcloud container clusters list --format='value(zone)' --filter='name:#{name}'`, so the yaml doesn't need to pass that in.

lib/snippets/deploy-to-gke.rb
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env ruby
```
KnVerey: Where does this script get invoked from? Does it care whether this file is `deploy-to-gke.rb` or `deploy-to-gke`?

lib/snippets/deploy-to-gke.rb
```diff
@@ -0,0 +1,136 @@
...
+    erb_binding = binding
+    template_variables.each do |var_name, value|
+      erb_binding.local_variable_set(var_name, value)
+    end
+    content = erb_template.result(erb_binding)
+
+    f = Tempfile.new(['kube_template', '.yml'])
+    f.write(content)
+    f.close
+    run_command("kubectl apply -f #{f.path} --namespace#{@namespace}")
```
byroot: you miss the `=` in `--namespace#{@namespace}`

lib/snippets/deploy-to-gke.rb
```diff
@@ -0,0 +1,136 @@
...
+    result = JSON.parse(`#{cmd}`)
+    result['resources'].each_with_object([]) do |resource, data|
+      next unless resource['type'] == 'container.v1.cluster'
+      properties = YAML.load(resource['finalProperties'])
+      data << [resource['name'], properties['zone']]
+    end
+  end
+
+  def run_command(cmd)
+    out, err, st = Open3.capture3(cmd)
```
byroot: Even though passing the entire command as a single string works, it's better to pass it with individual arguments (`Open3.capture3('echo', '-n', 'foo')`) for several reasons:
  - You don't have to deal with escaping when you interpolate arguments into the command
  - In some cases the command can lead Open3 to decide to run it in a shell to handle the command parsing, which can cause problems when trying to abort, (e.g https://github.com/Shopify/shipit-engine/issues/253)

So I suggest:
```ruby
run_command(*args)
out, err, st = Open3.capture3(*args)
...
```

And then things like:
```ruby
run_command("gcloud container clusters get-credentials #{cluster_name} --zone #{cluster_zone}")
```

becomes:

```ruby
run_command('gcloud', 'container', 'clusters', 'get-credentials', cluster_name, '--zone', cluster_zone)
```

byroot: Also `s/st/status/`

lib/snippets/deploy-to-gke.rb
```diff
@@ -0,0 +1,136 @@
...
+  def set_kubectl_cluster(cluster)
+    cluster_name = cluster[0]
+    cluster_zone = cluster[1]
+    st = run_command("gcloud container clusters get-credentials #{cluster_name} --zone #{cluster_zone}")
+    raise FatalDeploymentError, "Failed to set cluster #{cluster_name}/#{cluster_zone}" unless st
+  end
+
+  def fetch_clusters
+    cmd = "gcloud deployment-manager deployments describe #{@gcloud_deployment} --format=json"
+    result = JSON.parse(`#{cmd}`)
```
byroot: You never check this command actually worked. Also ``command`` is a bit smelly. Ideally you'd have a proper `capture('cmd', 'arg1', 'arg2')` that uses open3 for the same reasons explained after.

lib/snippets/deploy-to-gke.rb
```diff
@@ -0,0 +1,136 @@
...
+    f = Tempfile.new(['kube_template', '.yml'])
+    f.write(content)
+    f.close
+    run_command("kubectl apply -f #{f.path} --namespace#{@namespace}")
+  ensure
+    f.unlink if f
+  end
+
+  def authorize_gcloud
+    return if system('gcloud container clusters list')
```
byroot: Same thing here, I don't see why you use `system` instead of `run_command`.

lib/snippets/deploy-to-gke.rb
n1koo: The problem with that is that then you have to know the cluster name which points to a zone. Cluster for example is `test-us-east1-cluster` where as DM is simply `test-cluster`. Thats the reason this was build this way, eg. user can think of the DM name as the cluster and the fact its actually DM is just an implementation detail.

lib/snippets/deploy-to-gke.rb
n1koo: Does shipit handle the coloring out of box for the stdout/stderr in that case?

byroot: Shipit simply respect ANSI color codes.

n1koo:
https://github.com/Shopify/shipit-engine/pull/616/commits/19d4d4e276a5115008017d7f09e02063407ce1b7 should cover all the comments here. I also 🎩 'd the `ERB` side of things now. 

The one open thing still is the DM vs cluster name, but to recap I think using the deployment name is simplier for users since its the simple name (eg. `test-cluster`) whereas cluster is `test-us-eas-1-cluster` etc. This way user doesn't have to care where the cluster is. 

And by user I mean computes but really the same thing :)

lib/snippets/deploy-to-gke
```diff
@@ -0,0 +1,142 @@
...
+    f = Tempfile.new(['kube_template', '.yml'])
+    f.write(content)
+    f.close
+    run_command('kubectl', 'apply', '-f', f.path, "--namespace=#{@namespace}")
+  ensure
+    f.unlink if f
+  end
+
+  def authorize_gcloud
+    status = run_command('gcloud', '-q', 'auth', 'activate-service-account', '--key-file', "#{ENV['GCLOUD_CREDENTIALS_DIR']}/#{@gcloud_project}.json")
```
casperisfine: It would be interesting to check for `GCLOUD_CREDENTIALS_DIR` presence as early as possible, and exit 1 if it's missing.

lib/snippets/deploy-to-gke
```diff
@@ -0,0 +1,142 @@
...
+    result = query_deployments
+    result['resources'].each_with_object([]) do |resource, data|
+      next unless resource['type'] == 'container.v1.cluster'
+      properties = YAML.load(resource['finalProperties'])
+      data << [resource['name'], properties['zone']]
+    end
+  end
+
+  def run_command(*args)
+    puts args.join(" ")
```
casperisfine: I'd use `Shellwords.join` so that if we ever have arguments with spaces or quotes, this output won't be misleading as of what command was ran exactly.

lib/snippets/tmp
```diff
@@ -0,0 +1 @@
+/tmp
```
casperisfine: 🔥 ?

dwradcliffe:
One note from my perspective: it would be nice if it didn't matter which provider the cluster was running on either. This can be addressed later but something to keep in mind. Cluster names should be globally unique so we should be able to just say "deploy to k8s on this cluster" and shipit can figure it out.

n1koo:
@dwradcliffe even if the names are globally unique (dangerous assumption since theres nothing limiting this) theres no database to match up. I mean we could hardcode them to a config but thats ugly too. I think we need some sort of signal of the provider. Which can be simply saying `gke/foo-cluster` and `aws/bar-cluster` etc.

I'd keep that out of this for now though, but its fairly simple to change

ibawt:
@dwradcliffe we'll do something ot support the on prem/aws k8s deployments but we aren't there yet.  Solving todays problem today and that jazz

dwradcliffe:
@ibawt @n1koo Cool just wanted to make sure it was in the back of your mind

ibawt:
I'll add it to teh backlog @dwradcliffe , when you get a cluster up we can start hacking on it!

lib/snippets/deploy-to-gke
```diff
@@ -0,0 +1,147 @@
...
+  class FatalDeploymentError < StandardError; end
+
+  def initialize(namespace:, gcloud_deployment:, gcloud_project:, environment:, current_sha:, template_folder: nil)
+    @namespace = namespace
+    @gcloud_deployment = gcloud_deployment
+    @gcloud_project = gcloud_project
+    @current_sha = current_sha
+    enforce_required_parmas
+    @id = current_sha + "-#{SecureRandom.hex(4)}"
+    @template_path = './' + template_folder || "config/deploy/#{environment}"
```
mkobetic: Missing brackets?
```
irb(main):007:0> 'a' + nil || 'b'
TypeError: no implicit conversion of nil into String
	from (irb):7:in `+'
	from (irb):7
	from /usr/bin/irb:12:in `<main>'
irb(main):008:0> 'a' + (nil || 'b')
=> "ab"
```

n1koo: Good catch!

lib/snippets/deploy-to-gke
```diff
@@ -0,0 +1,147 @@
...
+
+# Usage: deploy-to-gke <app's namespace> <gcloud deployment> <gcloud project>
+
+# Requires the following to be available on Shipit machine:
+#  - gcloud binary available on the shipit machine's path
+#  - kubectl binary available in the shipit machine's path
+#  - $GCLOUD_CREDENTIALS_DIR/my-gcloud-project-name.json must exist
+
+# Optionally, the following variables can be used to override script defaults:
+#  - K8S_TEMPLATE_FOLDER: location of Kubernetes files to deploy. Default is config/<environment>/.
```
KnVerey: Default below is actually `config/deploy/<environment>`.

lib/snippets/deploy-to-gke
```diff
@@ -0,0 +1,147 @@
...
+class GKEDeployment
+  class FatalDeploymentError < StandardError; end
+
+  def initialize(namespace:, gcloud_deployment:, gcloud_project:, environment:, current_sha:, template_folder: nil)
+    @namespace = namespace
+    @gcloud_deployment = gcloud_deployment
+    @gcloud_project = gcloud_project
+    @current_sha = current_sha
+    enforce_required_parmas
+    @id = current_sha + "-#{SecureRandom.hex(4)}"
```
KnVerey: I think we should truncate the sha here. This is going to be appended to pod names to make them unique per deploy, and the max length of a pod name is 63 characters. This would only leave 15 characters for the real pod name. Could be needlessly annoying for our end users.

lib/snippets/deploy-to-gke
```diff
@@ -0,0 +1,147 @@
...
+    @template_path = './' + (template_folder || "config/deploy/#{environment}")
+
+    raise FatalDeploymentError, "Project config missing at #{ENV['GCLOUD_CREDENTIALS_DIR']}/#{@gcloud_project}.json" unless File.file?("#{ENV['GCLOUD_CREDENTIALS_DIR']}/#{@gcloud_project}.json")
```
KnVerey: Nit: `#{ENV['GCLOUD_CREDENTIALS_DIR']}/#{@gcloud_project}.json` is in here three times. Should we make an `@key_file` or something?

lib/snippets/deploy-to-gke
```diff
@@ -0,0 +1,147 @@
...
+  private
+
+  def enforce_required_parmas
+    [@namespace, @gcloud_project, @gcloud_deployment, @current_sha].each do |required_param|
+      raise ArgumentError, "#{required_param} is required" unless required_param && !required_param.empty?
+    end
+  end
+
+  def apply_all_templates
+    unless File.directory?(@template_path)
```
KnVerey: Similar to what @casperisfine said, it might be nice to bail earlier if we have nothing to deploy for this reason. Suggestion: we could make a validation method that handles the existing presence checks from `enforce_required_parmas` (there's a typo in "params" there btw), plus the keys file and template path validation. Potentially the presence of at least one yml/erb in the folder as well.

n1koo:
@KnVerey I think I addressed those, can you check the latest commit?

lib/snippets/deploy-to-gke
```diff
@@ -53,16 +57,19 @@ class GKEDeployment
...
+  def enforce_required_params
+    [@namespace, @gcloud_project, @gcloud_deployment, @current_sha, @key_file].each do |required_param|
       raise ArgumentError, "#{required_param} is required" unless required_param && !required_param.empty?
     end
   end
 
+  def preliminary_check
+    raise FatalDeploymentError, "Project config missing at #{@key_file}" unless File.file?(@key_file)
+    raise FatalDeploymentError, "#{@template_path} doesn't exist" unless File.directory?(@template_path)
+    raise FatalDeploymentError, "#{@template_path} doesn't have files with postfix .yml or .yml.erb" unless Dir.entries('/tmp').select {|file| file =~ /\.yml(.erb)?$/}.size > 0
```
KnVerey: Why is this looking in `/tmp`?
(FYI, you could do `.count { ... }` in place of `.select { ... }.size`)

n1koo: Argh, I fail at copypaste. It was my debug stuff. Also true, will change!

lib/snippets/deploy-to-gke
```diff
@@ -53,16 +57,19 @@ class GKEDeployment
...
+  end
+
   def apply_all_templates
-    unless File.directory?(@template_path)
-      raise FatalDeploymentError, "#{@template_path} doesn't exist"
-    end
     found = 0
```
KnVerey: Do we still need to count here if we've checked for templates on line 69?

n1koo: I felt like it would be nice if we ever had problems matching weird names or whatnot. Eg. if you have n files there and suddenly only see n-x being applied. I dont feel strongly, but just added info really

KnVerey:
Just the `tmp/` thing and the `0...8`. 🚢

PR #619, opened 2016-11-18
Get sha when running as task

KnVerey:
@Shopify/cloudplatform 
This is the bare minimum required to run the `deploy-to-gke` script from a task rather than a deploy. This is probably not the way we want to do it (would let deploys fall back to last revision, which makes no sense), but this would make it possible to give people a `migrate` button simply by modifying their `shipit.yml` during the code freeze. WDYT?

Here's the `shipit.yml` I was playing with locally: https://github.com/KnVerey/trashbin/blob/master/shipit.yml#L5-L17.

### DIFF
--- a/lib/snippets/deploy-to-gke
+++ b/lib/snippets/deploy-to-gke
@@ -43,6 +43,7 @@ class GKEDeployment
       set_kubectl_cluster(cluster)
       apply_all_templates
     end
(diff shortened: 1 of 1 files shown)


KnVerey:
Just chatted with @ibawt and we decided this isn't worthwhile. I'll keep working on the real feature and our testers can continue running their migrations asynchronously on deploy.

PR #622, opened 2016-11-30
Allow running a single test file

kmcphillips:
@byroot @jules2689 

Allows a single test file to be run using `dev test test/file_test.rb`

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -16,7 +16,7 @@ commands:
   bootstrap: bin/bootstrap shopify
   console: test/dummy/bin/rails console
   server: test/dummy/bin/rails server -b 192.168.64.1 -p 55330
-  test: bin/rake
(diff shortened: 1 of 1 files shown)


dev.yml
```diff
@@ -16,7 +16,16 @@ commands:
   bootstrap: bin/bootstrap shopify
   console: test/dummy/bin/rails console
   server: test/dummy/bin/rails server -b 192.168.64.1 -p 55330
-  test: bin/rake
```
casperisfine: Now that shipit-engine is on Rails 5, you can simply invoke `bin/rails test "$@"`, no need for conditionals.

kmcphillips: Sure!

kmcphillips:
Well that's way cleaner.

casperisfine:
Thanks.

PR #623, opened 2016-11-30
Allow variable in task to be a dropdown

kmcphillips:
@byroot @jules2689 @jonathankwok 

## Problem

In a few places we are using a `<input type="text">` to define a variable in a task. But that variable only has a few fixed values. Think regions, providers, caches, etc..

## Solution

If present, the `select:` key will replace it with a `<select>` dropdown with the preset values.

shipit_-_shipit-test

### DIFF
--- a/app/models/shipit/variable_definition.rb
+++ b/app/models/shipit/variable_definition.rb
@@ -1,11 +1,12 @@
 module Shipit
   class VariableDefinition
-    attr_reader :name, :title, :default
+    attr_reader :name, :title, :default, :select
 
     def initialize(attributes)
       @name = attributes.fetch('name')
       @title = attributes['title']
       @default = attributes['default'].to_s
+      @select = attributes['select'].presence
     end
 
     def to_h
@@ -13,6 +14,7 @@ module Shipit
         'name' => @name,
         'title' => @title,
         'default' => @default,
+        'select' => @select,
       }
     end
   end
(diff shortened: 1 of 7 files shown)


app/models/shipit/variable_definition.rb
```diff
@@ -1,18 +1,20 @@
 module Shipit
   class VariableDefinition
-    attr_reader :name, :title, :default
+    attr_reader :name, :title, :default, :select
 
     def initialize(attributes)
       @name = attributes.fetch('name')
       @title = attributes['title']
       @default = attributes['default'].to_s
+      @select = attributes['select'] if attributes['select'] && !attributes['select'].empty?
```
kmcphillips: Make it `nil` if we don't set it. So we can do `if variable.select` in the view.

test/unit/variable_definition_test.rb
```diff
@@ -0,0 +1,59 @@
+require 'test_helper'
+
+module Shipit
+  class VariableDefinitionTest < ActiveSupport::TestCase
```
kmcphillips: This class was completely untested. Got it passing first, then added my changes.

casperisfine: It was tested in `deploy_spec_test` but 👍

kmcphillips:
Updated CHANGELOG. Will merge on green.

PR #626, opened 2016-12-02
Properly set default env on deploys triggered by continuous delivery

casperisfine:
Even if you were to set a variable with a default, it would never be set on an automatically triggered deploy.

Now the defaults are set.

@rafaelfranca for review please.


rafaelfranca:
Awesome! Thanks

PR #627, opened 2016-12-06
Document shell command timeout

etiennebarrie:
Adds documentation for the shell command [option parsing](https://github.com/Shopify/shipit-engine/blob/f3f38004882118b787f71b51da511544e174567b/lib/shipit/command.rb#L201-L212) which allows setting [command-specific timeouts](https://github.com/Shopify/shipit-engine/blob/f3f38004882118b787f71b51da511544e174567b/lib/shipit/command.rb#L20).

### DIFF
--- a/README.md
+++ b/README.md
@@ -463,6 +463,24 @@ review:
     - bundle exec rake db:migrate:status
 ```
 
+<h3 id="shell-commands-timeout">Shell commands timeout</h3>
+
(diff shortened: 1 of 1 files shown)


etiennebarrie:
Switched to a `<h3>` for better consistency. Not sure if I should add a link to the TOC of the reference.

PR #628, opened 2017-01-05
Pure kubernetes deploy script

KnVerey:
@Shopify/cloudplatform  Before the holidays, I discovered some things about kubeconfigs and kubernetes auth that made me think we should easily be able to pull the gcloud parts out of our deploy script, making it more universal (i.e. theoretically equally usable for deployment to not-gke) and removing deploy-time dependencies. This is my attempt to do just that. See https://github.com/Shopify/shipit/compare/kubernetes_vs_gke for the effects on the Shipit side.
cc @byroot 

This PR: 
- Introduces a pure-kubernetes deploy script.
  - Usage: `kubernetes-deploy <app's namespace> <kube context>`
    - `ENV["KUBECONFIG"]` must point to a valid config with all the contexts you want to deploy to
  - does not use gcloud or deployment manager, or reference gcloud deployments/projects (gke script usage is `deploy-to-gke <app's namespace> <gcloud deployment> <gcloud project>`)
  - should be able to deploy to non-gke environments as well (auth is set up in advance rather than during the deploy, e.g. gke service account activated on shipit boot)
  - logging and validation are enhanced slightly, but the deploy still works exactly the same way--enhancements coming in another PR
  - uses named kubernetes contexts to map the tier defined in someone's shipit.yml to the actual cluster (e.g. `kubernetes-deploy my-cool-app tier4` looks for a context called tier4)
  - if you give it its own kubeconfig locally, testing deployment won't mess with your local configuration at all (and if you point it to your normal kubeconfig, it will only write current-context)
- Switches the gke script to also assume the standard google env var (`GOOGLE_APPLICATION_CREDENTIALS`) points to a single service account file it should always use, regardless of the gke project the cluster is in. 
  - My understanding is that the service account is meant to be a user identity, so we should be adding shipit's service account--with appropriate IAM--to additional projects as needed. Then there's no need for a bunch of service account secrets and messing with the credential env var at runtime.
- Resolves an exception I kept getting in development--NameError: uninitialized constant Shipit::Stack::UpdateEstimatedDeployDurationJob--by renaming a file

Next steps:
- Update Shopify's shipit to pick up these changes, commit a kubeconfig and set the correct env vars: https://github.com/Shopify/shipit/compare/kubernetes_vs_gke
- Test the new script with the test app in tier4. Test using it for rollback too.
- Migrate apps to new script when all is well. 
- Enhance new script with kubectl 1.5.1 features (i.e. dry run and prune) - [smarter_k8s_deploy](https://github.com/Shopify/shipit-engine/compare/smarter_k8s_deploy) - and tests
- Update Shopify's shipit to pick up those changes and upgrade to 1.5.1
- Enchance the new script with visualizations


lib/snippets/kubernetes-deploy
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env ruby
+
+# Usage: kubernetes-deploy <app's namespace> <kube context>
+
+# Prerequisites:
+#  - kubectl 1.5.1+ binary must be available in the shipit machine's path
+#  - ENV['KUBECONFIG'] must point to a valid kubeconfig file that includes all the contexts you want to deploy to
```
casperisfine: So if I understand correctly, that kubeconfig is specific to each application deployed? It's the kind of thing that will be in the repo being deployed right?

If so then I think a script argument would be better than an env var. Also if there is any semi-standard path for it, the argument should default to it.

lib/snippets/kubernetes-deploy
```diff
@@ -0,0 +1,172 @@
...
+      errors << "Kube config not found at #{ENV["KUBECONFIG"]}"
+    end
+
+    if @current_sha.nil?  || @current_sha.empty?
+      errors << "Current SHA must be specified"
+    end
+
+    if !File.directory?(@template_path)
+      errors << "Template path #{@template_path} doesn't exist"
+    elsif Dir.entries(@template_path).select {|file| file =~ /\.yml(.erb)?$/}.size.zero?
```
casperisfine: The `.` should be escaped in this regex: `\.erb`.

Also: `.size.zero?` -> `empty?`

lib/snippets/kubernetes-deploy
```diff
@@ -0,0 +1,172 @@
...
+
+    if @namespace.nil? || @namespace.empty?
+      errors << "Namespace must be specified"
+    end
+
+    if @context.nil? || @context.empty?
+      errors << "Context must be specified"
+    end
+
+    raise FatalDeploymentError, "Configuration invalid: #{errors.join(", ")}" if errors.length.nonzero?
```
casperisfine: `if errors.length.nonzero?` -> `unless errors.empty?`

lib/snippets/kubernetes-deploy
```diff
@@ -0,0 +1,172 @@
...
+    logger.info("All required parameters and files are present")
+  end
+
+  def apply_all_templates
+    found = 0
+    Dir.foreach(@template_path) do |file|
+      file_path = "#{@template_path}/#{file}"
+      if File.extname(file) == '.yml'
```
casperisfine: Instead of calling `extename` twice, you could use:

```ruby
case File.extname(file)
when '.yml'
  ...
```

lib/snippets/kubernetes-deploy
```diff
@@ -0,0 +1,172 @@
...
+        found += 1
+        apply_template(file_path)
+      elsif File.extname(file) == '.erb'
+        found += 1
+        render_and_apply_template(file_path)
+      end
+    end
+    raise FatalDeploymentError, "No templates found in #{@template_path}" if found.zero?
```
casperisfine: Seems like `found` doesn't need to be an int:

```ruby
found = false
...
found ||= true
...
raise ... unless found
```

lib/snippets/kubernetes-deploy
```diff
@@ -0,0 +1,172 @@
...
+      erb_binding.local_variable_set(var_name, value)
+    end
+    content = erb_template.result(erb_binding)
+
+    f = Tempfile.new(['kube_template', '.yml'])
+    f.write(content)
+    f.close
+    apply_template(f.path, original_path: file_path)
+  ensure
+    f.unlink if f
```
casperisfine: `Tempfile` takes care of GCing the file in it's finalizer.

lib/snippets/kubernetes-deploy
ibawt: should  be one per cluster not per app.

lib/snippets/kubernetes-deploy
ibawt: TIL

lib/snippets/kubernetes-deploy
KnVerey: Actually, the kubeconfig is specific to the deployer (e.g. Shopify Shipit), not any group of things being deployed. You can check out the one I wrote for Shopify Shipit here: https://github.com/Shopify/shipit/pull/120/files#diff-af16ffbedffc7b5f9e26cab1dd292bee. 

Essentially, the kubeconfig contains a list of clusters (e.g. we'll have one for each service tier) and a list of user accounts (just "shipit" for us, but e.g. locally I have a service account and my google auth login) + a list of contexts that pair the two together. All the apps being deployed need to specify is the name of the context: https://github.com/KnVerey/trashbin/blob/master/shipit.yml. 

I don't feel strongly about this, but the reason I left it as an env var is that that seemed to be the best way to provide it from Shipit to the script (since it is always the same in a given Shipit deployment) and `kubectl` supports using it directly: http://kubernetes.io/docs/user-guide/sharing-clusters/#merging-kubeconfig-example. There is a standard path, but it's `~/.kube/config`.

lib/snippets/kubernetes-deploy
KnVerey: I have been in the habit of explicitly unlinking anyway because of this recommendation in the Tempfile docs:

> When a Tempfile object is garbage collected, or when the Ruby interpreter exits, its associated temporary file is automatically deleted. This means that's it's unnecessary to explicitly delete a Tempfile after use, though it's good practice to do so: not explicitly deleting unused Tempfiles can potentially leave behind large amounts of tempfiles on the filesystem until they're garbage collected. The existence of these temp files can make it harder to determine a new Tempfile filename. Therefore, one should always call unlink or close in an ensure block

I'm guessing this script's ruby process is really short-lived though, making this extra-unnecessary here?

lib/snippets/kubernetes-deploy
ibawt: so it's for the "actor" , ie service account in this case.  
Authentication is so confusing :|

KnVerey: Well, a service account would be more like a "user" in a kubeconfig, which can have multiple users. For example, we could have a user that is gcp auth'd, and one that is aws auth'd.

casperisfine: Ok my bad then, env is better in this case.

casperisfine:
Nope, good for me. :shipit:

n1koo:
Im good 🚢

lib/snippets/kubernetes-deploy
```diff
@@ -0,0 +1,171 @@
+#!/usr/bin/env ruby
+
+# Usage: kubernetes-deploy <app's namespace> <kube context>
```
n1koo: the context is not going to be as pretty as `tierX`. Do we care? Should we fuzzy match it? And/or wrap htis somehow?

PR #630, opened 2017-01-10
Smarter k8s deploy

KnVerey:
@Shopify/cloudplatform @byroot 

**Important**: This script now requires kubectl 1.5.1+. We'll have to replace the version in Shipit before picking up these changes.

This adds a bunch of features to `kubernetes-deploy`:

- It uses `apply --dry-run` to get up-front validation and discovery of all the individual resources in the deploy
- It deploys certain resources first, waiting for all resources of a given type to be up before continuing.
- It prunes resources that were previously `kubectl apply`'d and are not present in the current set  (if they're of a whitelisted type). This notably cleans up old migration pods.
- It is aware of some basic resource types (6 for starters) and knows how to detect whether they're running or failing...
- ...and as a result can make the deploy fail/succeed appropriately!

Notes:

- Only certain resource types are recognized. Theoretically this should match the prune whitelist, but for now I've got the set we're commonly providing: pods, deployments, configmaps, ingress, persistentvolumeclaims and services. Unrecognized types are applied during the bulk apply/prune and immediately considered to have been successful (though a warning is logged about this assumption).
- The script assumes that bare pods are always priority resources that should run to completion (e.g. migrations).
- If you kubectl apply something locally that isn't committed to the repo, and then deploy with this script, that thing will get pruned if it is of one of the types on the `PRUNE_WHITELIST`. If its type isn't on the list or you create it with a command other than `apply`, it will stick around.
- The script times out eventually if it can't determine all resources are up. (`STEP_COMPLETION_TIMEOUT_MINUTES=20`... too long?)
- The predeployed resources are re-applied during the bulk apply that uses prune (if they weren't included, they'd get pruned). This should be a no-op.
- The `[KUBESTATUS]` logging is really for the planned JS-based visualization (which doesn't exist at all yet), but could also be useful for debugging in the meantime.
- I still haven't added any tests. Since they'll be a huge pain to write and likely very stubbed, I'd like to continue deferring this until the script is more stable.

Please give this a try locally! I'm going to spend more time testing tomorrow as well, including the following scenarios I haven't tried recently:
- [x] invalid yaml - fails deploy during content parsing
- [x] valid yaml that is not a valid kube template - fails deploy during content parsing
- [x] unrecognized resource type (used a secret)
- [x] image pull error - warns on discovery, fails deploy as a timeout
- [x] configmap required by pod missing - warns on discovery, fails deploy as a timeout
- [x] failed bare pod - fails the predeploy phase
- [ ] "lost" pvc
- [x] massive scale up/down (that takes a while)
- [x] deploy with multiple bad resources
- [x] timeout - logs an error when it happens, then fails the phase
- [x] ingress template - works fine
- [x] rollback -- notably, does rolling back a deployment create a new rs or scale up the old one? Scales up the old one

### DIFF
--- a/lib/snippets/kubernetes-deploy
+++ b/lib/snippets/kubernetes-deploy
@@ -45,12 +45,10 @@ class KubernetesDeploy
   ).freeze
 
   PREDEPLOY_SEQUENCE = %w(
-    configmap
-    persistentvolumeclaim
-    pod
+    ConfigMap
+    PersistentVolumeClaim
+    Pod
   )
-
-  STEP_COMPLETION_TIMEOUT_MINUTES = 20
 
   def initialize(namespace:, environment:, current_sha:, template_folder: nil, context:)
     @namespace = namespace
@@ -104,6 +102,10 @@ class KubernetesDeploy
       next if matching_resources.empty?
       deploy_resources(matching_resources)
       wait_for_completion(matching_resources)
+      fail_count = matching_resources.count { |r| r.deploy_failed? || r.deploy_timed_out? }
+      if fail_count > 0
+        raise FatalDeploymentError, "#{fail_count} priority resources failed to deploy"
+      end
     end
   end
 
@@ -143,29 +145,30 @@ class KubernetesDeploy
   end
 
   def report_final_status(resources)
-    if resources.all?(&:succeeded?)
+    if resources.all?(&:deploy_succeeded?)
       log_green("Deploy succeeded!")
     else
-      fail_list = resources.select(&:failed?).map(&:id)
+      fail_list = resources.select { |r| r.deploy_failed? || r.deploy_timed_out? }.map(&:id)
       KubernetesDeploy.logger.error("The following resources failed to deploy: #{fail_list.join(", ")}")
       raise FatalDeploymentError, "#{fail_list.length} resources failed to deploy"
     end
   end
 
-  def wait_for_completion(resources_to_deploy)
-    timeout = Time.now.utc + (60 * STEP_COMPLETION_TIMEOUT_MINUTES)
+  def wait_for_completion(watched_resources)
     delay_sync_until = Time.now.utc
-
-    while !resources_to_deploy.empty?
-      if Time.now.utc > timeout
-        Kubernetes.logger.error("Phase #{@current_phase} timed out deploying the following resources: #{resources_to_deploy.map(&:id).join(", ")}")
-        raise FatalDeploymentError, "Deploy phase #{@current_phase} failed to complete within #{STEP_COMPLETION_TIMEOUT_MINUTES} minutes"
-      elsif Time.now.utc < delay_sync_until
+    while watched_resources.present?
+      if Time.now.utc < delay_sync_until
         sleep (delay_sync_until - Time.now.utc)
       end
       delay_sync_until = Time.now.utc + 3 # don't pummel the API if the sync is fast
-      resources_to_deploy.each(&:sync)
-      resources_to_deploy = resources_to_deploy.reject(&:deployed?)
+      watched_resources.each(&:sync)
+      newly_finished_resources, watched_resources = watched_resources.partition(&:deploy_finished?)
+      newly_finished_resources.each do |resource|
+        next unless resource.deploy_failed? || resource.deploy_timed_out?
+        KubernetesDeploy.logger.error("#{resource.id} failed to deploy with status '#{resource.status}'.")
+        KubernetesDeploy.logger.warn("Continuing to poll status of other resources in this phase. You may wish to abort the deploy.")
+        KubernetesDeploy.logger.error(resource.status_data)
+      end
     end
   end
 
@@ -215,6 +218,7 @@ class KubernetesDeploy
     resources.each do |r|
       KubernetesDeploy.logger.info("- #{r.id}")
       command.push("-f", r.file.path)
+      r.deploy_started = Time.now.utc
     end
 
     if prune
@@ -276,8 +280,8 @@ class KubernetesDeploy
       l.level = ENV["DEBUG"] ? Logger::DEBUG : Logger::INFO
       l.formatter = proc do |severity, _datetime, _progname, msg|
         case severity
-        when "FATAL" then "\033[0;31m[#{severity}]\t#{msg}\x1b[0m\n" # red
-        when "ERROR", "WARN" then "\033[0;33m[#{severity}]\t#{msg}\x1b[0m\n" # yellow
+        when "FATAL", "ERROR" then "\033[0;31m[#{severity}]\t#{msg}\x1b[0m\n" # red
+        when "WARN" then "\033[0;33m[#{severity}]\t#{msg}\x1b[0m\n" # yellow
         when "INFO" then "\033[0;36m#{msg}\x1b[0m\n" # blue
         else "[#{severity}]\t#{msg}\n"
         end
@@ -290,15 +294,16 @@ class KubernetesDeploy
     extend ActiveSupport::DescendantsTracker
 
     attr_reader :name, :namespace, :file
-    attr_writer :type
+    attr_writer :type, :deploy_started
 
+    TIMEOUT = 60 * 5
 
     def self.handled_type
-      name.split('::').last.downcase
+      name.split('::').last
     end
 
     def self.for_type(type, name, namespace, file)
-      if subclass = descendants.find { |subclass| subclass.handled_type == type }
+      if subclass = descendants.find { |subclass| subclass.handled_type.downcase == type }
         subclass.new(name, namespace, file)
       else
         self.new(name, namespace, file).tap { |r| r.type = type }
@@ -318,11 +323,11 @@ class KubernetesDeploy
       log_status(status_data)
     end
 
-    def failed?
+    def deploy_failed?
       false
     end
 
-    def succeeded?
+    def deploy_succeeded?
       KubernetesDeploy.logger.warn("Don't know how to monitor resources of type #{type}. Assuming #{id} deployed successfully.")
       true
     end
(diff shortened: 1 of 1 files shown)


KnVerey:
Screenshot of deploying https://github.com/knverey/trashbin with this script:

shipit1
shipit2

lib/snippets/kubernetes-deploy
```diff
@@ -75,90 +203,375 @@ class KubernetesDeploy
...
+
+  class KubernetesResource
+    attr_reader :name, :namespace, :file
+    attr_writer :type
+
+    # Source: Rails
+    # https://github.com/rails/rails/blob/b292b76c2dd0f04fb090d49b90716a0e6037b41a/activesupport/lib/active_support/core_ext/class/subclasses.rb#L22-L29
+    def self.descendants
+      descendants = []
+      ObjectSpace.each_object(singleton_class) do |k|
```
casperisfine: This implementation is good for had hock inspection of a class. But when you know you'll need to call it, there is a better implementation: https://github.com/rails/rails/blob/9f84e60ac9d7bf07d6ae1bc94f3941f5b8f1a228/activesupport/lib/active_support/descendants_tracker.rb

lib/snippets/kubernetes-deploy
```diff
@@ -75,90 +203,375 @@ class KubernetesDeploy
...
+    def failed?
+      false
+    end
+
+    def succeeded?
+      KubernetesDeploy.logger.warn("Don't know how to monitor resources of type #{type}. Assuming #{id} deployed successfully.")
+      true
+    end
+
+    def exists?
```
casperisfine: Does this empty method mean it has to be implemented by the subclass, or that the default value is `nil`?

If it's the former then please use `raise NotImplementedError, "KubernetesResource subclasses must implement #exists?"`.

If it's the later, then please return nil or false explicitly.

lib/snippets/kubernetes-deploy
```diff
@@ -75,90 +203,375 @@ class KubernetesDeploy
...
+  end
+
+  class Service < KubernetesResource
+    def initialize(name, namespace, file)
+      @name, @namespace, @file = name, namespace, file
+    end
+
+    def sync
+      endpoints, st = run_kubectl("get", "endpoints", @name, "--output=jsonpath={.subsets[*].addresses[*].ip}")
+      @num_endpoints = endpoints.split(/\s+/).length
```
casperisfine: `/\s+/` is the default. Just call `split` without arguments.

lib/snippets/kubernetes-deploy
```diff
@@ -50,6 +96,88 @@ class KubernetesDeploy
...
+        type, name = resource_id.split("/", 2) # e.g. "pod/web-198612918-dzvfb"
+        resources << KubernetesResource.for_type(type, name, @namespace, tempfile)
+        KubernetesDeploy.logger.info "Discovered template for #{resource_id}"
+      end
+    end
+    resources
+  end
+
+  def discover_resource_via_dry_run(tempfile)
+    resource_id, err, st = run_kubectl("apply", "-f", tempfile.path, "--dry-run", "--output=name")
```
mkobetic: Worth logging the error too?

KnVerey: The main script's `run_kubectl` always does automatically:
`KubernetesDeploy.logger.warn(err) unless st.success?`

Note that the resource instances' version doesn't though. My reasoning for that is that polling errors aren't blocking, and may of them will be 404s on GET for resources that aren't up yet... and that's not really an error as far as the script is concerned.

lib/snippets/kubernetes-deploy
```diff
@@ -75,90 +203,375 @@ class KubernetesDeploy
...
+      deployment_stats, st = run_kubectl("get", type, @name)
+      @found = st.success?
+      if @found
+        # deployment_stats looks like this:
+        # NAME                  DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
+        # jobs                  1         1         1            1           43d
+        labels, statuses = deployment_stats.split("\n")
+        labels = labels.split(/\s+/)[1..4]
+        statuses = statuses.split(/\s+/)[1..4]
+        @rollout_data = labels.zip(statuses).to_h # { "DESIRED" => 1 } etc.
```
mkobetic: Why not jsonpath here?

lib/snippets/kubernetes-deploy
KnVerey: Thanks! I'll switch to that.

lib/snippets/kubernetes-deploy
KnVerey: It's the latter (the base class will be used by any resource of a type there's no specific handler for). Will do.

lib/snippets/kubernetes-deploy
KnVerey: Good question... it was because this was exactly what I wanted, and I was having trouble digging it out of the `get deployment` data. Looking again, it shouldn't be _too_ bad. I'll give it another shot.

KnVerey:
My 🎩  of the deploy failure scenarios list in the description (using [this branch](https://github.com/KnVerey/trashbin/compare/errors) of my test repo) ended up pointing out a whole bunch of issues that I've addressed in the latest commits (I'll squash before merging). Please take another look, particularly at the timeout logic and the `Pod` and `Deployment` subclasses, which required extensive changes. 🙏 

@n1koo if you could please give this a 🎩  as well on Monday that would be fantastic!

lib/snippets/kubernetes-deploy
```diff
@@ -66,99 +200,455 @@ class KubernetesDeploy
...
-    content = erb_template.result(erb_binding)
 
-    f = Tempfile.new(['kube_template', '.yml'])
-    f.write(content)
-    f.close
-    apply_template(f.path, original_path: file_path)
-  end
+    if prune
+      command.push("--prune", "--all")
+      PRUNE_WHITELIST.each { |type| command.push("--prune-whitelist=#{type}") }
```
mkobetic: IIUC what's in the docs, --prune-whitelist should accept an array, are you sure using the option multiple times accumulates the entries?

lib/snippets/kubernetes-deploy
```diff
@@ -66,99 +200,455 @@ class KubernetesDeploy
...
+      log_status
+    end
+
+    def interpret_json_data(pod_data)
+      @phase = (pod_data["metadata"]["deletionTimestamp"] ? "Terminating" : pod_data["status"]["phase"])
+      @containers = pod_data["spec"]["containers"].map { |c| c["name"] }
+
+      if @deploy_started && pod_data["status"]["containerStatuses"]
+        pod_data["status"]["containerStatuses"].each do |status|
+          reason = status.dig("state", "waiting", "reason")
```
mkobetic: TIL `dig`

lib/snippets/kubernetes-deploy
```diff
@@ -66,99 +200,455 @@ class KubernetesDeploy
...
+
+        pod_list, st = run_kubectl("get", "pods", "-a", "-l", "name=#{name}", "--output=json")
+        if st.success?
+          pods_json = JSON.parse(pod_list)["items"]
+          pods_json.each do |pod_json|
+            pod_name = pod_json["metadata"]["name"]
+            pod = Pod.new(pod_name, namespace, nil, parent: "#{@name.capitalize} deployment")
+            pod.deploy_started = @deploy_started
+            pod.interpret_json_data(pod_json)
+            pod.log_status
```
mkobetic: Do we want to log just first few in case there are hundreds?

lib/snippets/kubernetes-deploy
```diff
@@ -66,99 +200,455 @@ class KubernetesDeploy
...
-    erb_binding = binding
-    template_variables.each do |var_name, value|
-      erb_binding.local_variable_set(var_name, value)
+  def deploy_resources(resources, prune: false)
+    command = ["apply", "--namespace=#{@namespace}"]
+    KubernetesDeploy.logger.info("Deploying resources:")
+
+    resources.each do |r|
+      KubernetesDeploy.logger.info("- #{r.id}")
+      command.push("-f", r.file.path)
```
n1koo: Could we use `-R` and just point it to the path?

lib/snippets/kubernetes-deploy
KnVerey: Yes, definitely. In fact, this seems to be the only way to do it... I initially got frustrated trying to do things like `=[list]` `=thing1,thing2`. Turns out there's an example on the PR that introduced the option: https://github.com/kubernetes/kubernetes/pull/35220

lib/snippets/kubernetes-deploy
KnVerey: So then instead of saving the rendered templates to `{tempfile}` we'd save them to `{type}/{tempfile}`? Seems like slightly more work--is there a concrete advantage to doing that?

lib/snippets/kubernetes-deploy
KnVerey: It's a ruby 2.3.1 thing... actually, our Shipit is on 2.3.1, but that doesn't mean everyone's is. It seems like the engine's official requirement is 2.2.2 though (https://github.com/Shopify/shipit-engine/blob/de0aeeaa4920a520ff0ec4a3b228795adc3dd0ec/template.rb#L3), so I actually shouldn't use this!

lib/snippets/kubernetes-deploy
n1koo: hmm right yeah, probably not worth it. I was mostly thinking in terms of not having to loop here but guess the amount of things isn't that big and this is super cheap anyways so kcso

lib/snippets/kubernetes-deploy
KnVerey: The idea behind this logging is that a visualization script will make little boxes for all the resources and make them change colours as they change state. Therefore, it's important to write a line for every resource. It's not meant to be human-readable (though it is somewhat useful for debugging in the meantime).

Note: I discovered that, somewhat unsurprisingly, the overhead for including logs and/or events in this output was unacceptably enormous, so  I ripped it all out; this logging no longer triggers any additional requests. After all, logs are available in splunk, and events _should_ be (but aren't iirc). A potential future enhancement could be to fetch and dump logs and events for failed resources only at the time a deploy fails.

lib/snippets/kubernetes-deploy
```diff
@@ -66,99 +200,455 @@ class KubernetesDeploy
...
+      false
+    end
+
+    def exists?
+      @found
+    end
+  end
+
+  class Pod < KubernetesResource
+    TIMEOUT = 60 * 15
```
n1koo: 🚲 🏠  but isn't this a bit excessive?

KnVerey: Totally a fair question! Please feel free to comment on things like this. It's super arbitrary. What timeout would you suggest here?

My thought was that run-once pods could theoretically be doing something that takes quite a while (e.g. a dirty `rake db:migrate` that does maintenance instead of enqueuing jobs to do maintenance), and they can always manually abort sooner. Fundamentally, these timeouts are to protect Shipit from running forever for no reason. App developers can set timeouts on their pods, and if those get hit, the deploy will hard fail immediately.

ibawt: if we added activesupport does that mean we can use .seconds ?

ibawt:
looks great, I think my only comment is that it's a lot of code for a script.  Not sure how best to fix that short term though.

lib/snippets/kubernetes-deploy
KnVerey: I only added specific methods from ActiveSupport that I really wanted to use, but presumably I could cherry-pick that one too.

n1koo: Makes sense to prepare for `db:migrate`. I am okay not babysitting people doing silly stuff so KCSO :)

kirs:
:clap: :clap: :clap:

sirupsen:
Just looked through the code, besides the missing tests this looks ✨ !

PR #631, opened 2017-01-13
Basic NPM support

GoodForOneFare:
This is basic, but I'm using it to deploy [my simple test stack](https://github.com/goodForOneFare/shipit_test):
- For dependencies, it just does `npm install`
- For deploy steps, checks tags, then does `npm run deploy`
  - `npm run deploy` is what all Shopify repos currently use as a deployment target
  - Maybe we should standardise on `prepublish`? /cc @lemonmade 

Stuff that isn't implemented but could happen:
- Adding `node`/`npm` to `dev.yml` (will appear in this PR soon)
- Yarn install + discovery
- `nvm` install + support for `node` version specified in a repo's `package.json`
- Publishing with beta tags
- Publishing to private registries
- Publishing scoped packages to the main npm registry
- Warning that `shopify-dep` hasn't been added as an npm collaborator


byroot:
> Maybe we should standardise on prepublish? /cc @lemonmade

`npm publish` seem to be the standard way to publish a package. It'd be best to use it.

app/models/shipit/deploy_spec/npm_discovery.rb
```diff
@@ -0,0 +1,49 @@
...
+        [%(npm install --no-progress)]
+      end
+
+      def discover_review_checklist
+        discover_npm_checklist || super
+      end
+
+      def discover_npm_checklist
+        [%(<strong>Don't forget version and tag before deploying!</strong> You can do this with:<br/>
+          npm version <strong>&lt;patch type&gt;</strong> && git push --tags</pre>)] if npm?
```
lemonmade: The <patch type> part looks weird, maybe <type> or [major, minor, patch]

GoodForOneFare: Agreed. I left it vague because there are many more options, but anyone looking for those likely has the confidence to ignore these suggestions.

GoodForOneFare:
The deploy steps now just do `npm publish`, and the `npm version` instructions are clearer.  Please take another look.

app/models/shipit/deploy_spec/npm_discovery.rb
```diff
@@ -0,0 +1,49 @@
...
+        [%(<strong>Don't forget version and tag before publishing!</strong> You can do this with:<br/>
+          npm version <strong>&lt;major|minor|patch&gt;</strong> && git push --tags</pre>)] if npm?
+      end
+
+      def npm?
+        package_json.exist?
+      end
+
+      def package_json
+        file('./package.json')
```
casperisfine: the `./` part is useless.

lemonmade:
Ah, good catch @casperisfine. I think we should probably both move the npm one to be lowest priority and check to make sure the `private` field in `package.json` isn't `true`.

GoodForOneFare:
Using `private` will work for most Shopify repos.  Is Twine an oddball in that it requires Gem + NPM publication?

GoodForOneFare:
npm is now the lowest priority; private packages are ignored; `./package.json` corrected; more tests.  Please take another look.  Thanks!

test/unit/deploy_spec_test.rb
```diff
@@ -359,5 +359,37 @@ class DeploySpecTest < ActiveSupport::TestCase
     test "#clear_working_directory? returns true by default" do
       assert_predicate @spec, :clear_working_directory?
     end
+
+    test '#npm? is false if there is no package.json' do
```
lemonmade: Can we add a test to make sure that `npm` takes lower precedence than the other ones?

GoodForOneFare:
Tests added for bundler + gems precedence.  Thank you both for your feedback.

GoodForOneFare:
@lemonmade good to go?

Issue #632, opened 2017-01-16
Make Kubernetes deploys a first-class citizen


KnVerey:
Right now, our Kubernetes deploys use an override snippet. That's okay while it's experimental, but won't be great when this is a primary deployment strategy. Currently, the `shipit.yml` for a kube app needs to look like this:

```yaml
deploy:
  override:
    - kubernetes-deploy trashbin tier5

rollback:
  override:
    - kubernetes-deploy trashbin tier5
```

This would be much better:

```yaml
kubernetes:
  namespace: my-app-production
  context: tier4
```

@byroot
cc @Shopify/cloudplatform

KnVerey:
Two caveats related to the WIP nature of our k8s deploys:
* We may also want a third, optional key: `template_dir`. That value is provided as an env override in the current script. The caveat on this is that we're still working out our multi-env template strategy. For example, we may add more sophisticated rendering and have a single directory + `production.yml`, `staging.yml` etc. instead of a directory per env, which would affect this input requirement.

* We _may_ also want to support something along the lines of the following, for weird cases like core that may want to deploy a single env to multiple namespaces (@ibawt):
  ```yaml
  kubernetes:
    - namespace: my-app-production-shardA
      context: tier4
      template_dir: /app/config/deploy/production/shardB/
    - namespace: my-app-production-shardB
      context: tier4
      template_dir: /app/config/deploy/production/shardB/
  ```
  This is **highly** speculative though, so I'd go with what I originally posted unless it would prohibit introducing a variant like the above in the future.

dwradcliffe:
for production vs staging, the pattern is a separate shipit.yml

KnVerey:
> for production vs staging, the pattern is a separate shipit.yml

Yes, totally. The complication with core is that you also have "pod1 production" vs "pod2 production", which should be deployed to separate namespaces at the same time with a single button click. (Or such was the plan last I heard.)

dwradcliffe:
> deployed to separate namespaces at the same time

cool, yep. 👍

PR #633, opened 2017-01-17
First class kubernetes integration

casperisfine:
Fix: https://github.com/Shopify/shipit-engine/issues/632

@Shopify/cloudplatform for review please.

Note: I didn't add support for multiple namspaces as it's not well specified yet. Since `kubernetes-deploy` can only deploy to a single namespace, I don't know if you want this to happen sequentially or concurrently. I suggest we discuss that feature in another issue when needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,6 +320,19 @@ For example:
 fetch:
   curl --silent https://app.example.com/services/ping/version
 ```
+<h3 id="kubernetes">Kubernetes</h3>
+
+**<code>kubernetes</code>** allows to specify a Kubernetes namespace and context to deploy to.
+
+For example:
+```yml
+kubernetes:
+  namespace: my-app-production
+  context: tier4
+```
+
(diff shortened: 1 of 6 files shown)


ibawt:
yah not sure what a multinamespace deploy will look like.  Just shopify at this point though.

README.md
```diff
@@ -320,6 +320,18 @@ For example:
...
+<h3 id="kubernetes">Kubernetes</h3>
+
+**<code>kubernetes</code>** allows to specify a Kubernetes namespace and context to deploy to.
+
+For example:
+```yml
+kubernetes:
+  namespace: my-app-production
+  context: tier4
+  template_dir: config/deploy/k8s_templates
```
KnVerey: nit: the default if you leave this out is `config/deploy/$ENVIRONMENT` so perhaps we should state that and/or make the example reference an environment.

casperisfine: Done.

KnVerey:
Gave this a quick 🎩  as well. 👌

PR #634, opened 2017-01-18
Move kubernetes-deploy bin into a gem

kirs:
Not sure it's the best way (requiring the gem to be installed in `KubernetesDiscovery` and failing if it's not). But from the other hand, I don't want to make the gem a hard dependency of `shipit-engine`.

We will add the gem as a dependency into `Shopify/shipit` and it will use the bin from https://github.com/Shopify/kubernetes-deploy/blob/master/exe/kubernetes-deploy

review @casperisfine @KnVerey @sirupsen


casperisfine:
Would be weird to have https://github.com/Shopify/shipit-engine/pull/633 but having to also add the gem in your installation.

I'd rather have a very open requirement on the gem. e.g. `~> 0.1`.

kirs:
Sounds good. I published the gem and updated the PR.

casperisfine:
The test failure is legit (bundle exec shouldn't be prepended, because the gem won't be in the deployed project.).

kirs:
I updated the code to use `Gem.bin_path("kubernetes-deploy")` to locate the bin. IMO it's the most consistent way; other projects like Rake and Rails are using it too.
CI is now green.

PR #635, opened 2017-01-19
Replace with_clean_env to with_original_env

kirs:
review @casperisfine

### DIFF
--- a/app/jobs/shipit/perform_task_job.rb
+++ b/app/jobs/shipit/perform_task_job.rb
@@ -52,7 +52,7 @@ module Shipit
     end
 
     def perform_task
-      Bundler.with_clean_env do
+      Bundler.with_original_env do
         capture_all! @commands.install_dependencies
         capture_all! @commands.perform
       end


arthurnn:
What is the reason behind this change? 
I think we(rubygems.org) are seeing the `BUNDLE_GEMFILE` leak from the main Shipit rails app to our stack app.. during the `bundle install` stage.
cc @dwradcliffe

dwradcliffe:
Confirmed by changing this back, everything works as it did before.

byroot:
IIRC we were trying to make https://github.com/Shopify/shipit-engine/pull/634 works, but in the end it was too compllicated and we moved to a totally different solution.

And https://github.com/bundler/bundler/blob/42b1e1b9c3f4b3a45cd27214ef656911951b4d94/lib/bundler.rb#L256-L257 also hinted that `with_clean_env` might not be a good thing.

I'll revert this since it does no good, and definitely does some harm.

PR #636, opened 2017-01-19
Remove kubernetes-deploy as a dependency

kirs:
As discussed, we decided to stop shipping `kubernetes-deploy` as a dependency of Shipit. Instead, we expect developers to prepare `gem install kubernetes-deploy` in the environment where Shipit is hosted.

review @casperisfine
vv @KnVerey


KnVerey:
Should we explain this requirement in the readme?

casperisfine:
> Should we explain this requirement in the readme?

Good point.

kirs:
I was going to do that in another PR if you don't mind.

PR #638, opened 2017-01-19
Basic yarn support

GoodForOneFare:
Very similar to npm's logic.  If a project has a `yarn.lock` file, `yarn` is used to install/publish the package.

Note that this can't ship until my [yarn borg setup PR is merged into shipit](https://github.com/Shopify/shipit/pull/129).

cc @lemonmade


app/models/shipit/deploy_spec/yarn_discovery.rb
```diff
@@ -0,0 +1,49 @@
...
+        discover_yarn_checklist || super
+      end
+
+      def discover_yarn_checklist
+        [%(<strong>Don't forget version and tag before publishing!</strong> You can do this with:<br/>
+          yarn version --new-version <strong>&lt;major|minor|patch&gt;</strong> && git push --tags</pre>)] if yarn?
+      end
+
+      def yarn?
+        yarn_lock.exist? && public?
```
GoodForOneFare: This is relying on `NpmDiscovery#public?` being merged into the module.  That feels wrong, so is there a better way to do this?

casperisfine: Yarn and Npm are very similar tools, and both work against the `npm.org` service. It wouldn't seem wrong to handle boths in the `NpmDiscovery` module

lemonmade: Yeah, I like the idea of them being a single module, and we just change the commands to match `yarn` if we find `yarn.lock`, `npm` if we don't.

GoodForOneFare: ¯\\_(ツ)_/¯ I liked that yarn's precedence over npm was made clear in `FileSystem`'s include order, and -- while it might not be obvious -- the tests are harder to reason about now.

Regardless, I've updated to a more concise implementation (a more greppable/verbose interim implmentation is at https://github.com/Shopify/shipit-engine/pull/638/commits/5a728f8d583fd587f371939a02bc51fef2a491bf).

casperisfine:
You have a couple stylecheck failures, other than that LGTM.

PR #639, opened 2017-01-20
Documentation for Heroku and Kubernetes integrations

kirs:
review @casperisfine @KnVerey

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ This guide aims to help you [set up](#installation-and-setup), [use](#using-ship
 * [Format and content of shipit.yml](#configuring-shipit)
 * [Format and content of secrets.yml](#configuring-secrets)
 * [Script parameters](#script-parameters)
+* [Configuring providers](#configuring-providers)
 * [Free samples](/examples/shipit.yml)
 
 * * *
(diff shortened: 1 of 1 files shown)


KnVerey:
Shouldn't the docs include an example of what the shipit.yml needs to look like to use these integrations?  There are a few other points that might be very helpful as well re k8s:
- The `kubectl` requirement is specifically for v 1.5 or higher, which is very new. We use several features introduced in that version, and the script will not be able to deploy with anything older.
-  `ENV['KUBECONFIG']` (via Secrets[:env]) must point to a valid kubeconfig file that includes all the contexts you want to deploy to
- `ENV['GOOGLE_APPLICATION_CREDENTIALS']` must point to the credentials for an authenticated service account if your have a user with the gcp auth provider
- Optionally, you can set either the `template_dir` key or `K8S_TEMPLATE_FOLDER` machine env in a shipit.yml to override the default file location (which is `config/deploy/#{environment}`)

kirs:
Thanks for suggestions!

I was planning to add the `kubectl` version requirement directly to `kubernetes-deploy` because it's up to that utility to require specific version.

For the rest, I'll prepare another PR with more details.

PR #640, opened 2017-01-28
Pass k8s templates as --template-path

kirs:
To support the flag introduced in https://github.com/Shopify/kubernetes-deploy/pull/11

cc @sirupsen @KnVerey

### DIFF
--- a/app/models/shipit/deploy_spec/kubernetes_discovery.rb
+++ b/app/models/shipit/deploy_spec/kubernetes_discovery.rb
@@ -9,24 +9,21 @@ module Shipit
         discover_kubernetes || super
       end
 
-      def discover_machine_env
-        env = super
-        env = env.merge('K8S_TEMPLATE_FOLDER' => kube_config['template_dir']) if kube_config['template_dir']
-        env
-      end
-
       private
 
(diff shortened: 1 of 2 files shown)


app/models/shipit/deploy_spec/kubernetes_discovery.rb
```diff
@@ -9,24 +9,13 @@ def discover_rollback_steps
...
 
-        [
-          Shellwords.join([
-            "kubernetes-deploy",
-            kube_config['namespace'],
-            kube_config['context'],
-          ]),
-        ]
+        template_dir = kube_config['template_dir'] || "config/deploy/$ENVIRONMENT"
+        ["kubernetes-deploy #{kube_config['namespace']} #{kube_config['context']} --template-dir #{template_dir}"]
```
sirupsen: Why do you not need to escape this anymore with `Shellwords#join`?

casperisfine: I think the problem is that `Shellwords.join` escapes the `$` in `$ENVIRONMENT`.

Using `Shellwords.escape` on the other arguments might be a good idea though.

PR #642, opened 2017-02-01
Yarn deployments should not hang

GoodForOneFare:
`yarn publish` [has an unskippable user input prompt](https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/publish.js#L136), so always use `npm publish` for JS projects.


GoodForOneFare:
@casperisfine there's [an open issue for this](https://github.com/yarnpkg/yarn/issues/2445).  I'll keep an eye on that (maybe even take a stab at fixing it).

PR #643, opened 2017-02-02
Experimental merge queue

casperisfine:
This implements a simple merge queue for stacks.
  - Pull Requests registered
  - If the stack allows merges (is not locked, is not backlogged, is not failing), the pull request will be merged.
  - If the pull request is not mergeable (merge conflict, failing commit status, etc) it will be rejected.
  - Pull requests can be removed from the queue.
  - API endpoints to list / add / remove pull requests from the queue.

Future features:
  - UI to add a PR to the queue
  - UI to list PRs in the queue

### DIFF
--- a/app/models/shipit/pull_request.rb
+++ b/app/models/shipit/pull_request.rb
@@ -1,0 +1,84 @@
+module Shipit
+  class PullRequest < ApplicationRecord
+    belongs_to :stack
+    belongs_to :head, class_name: 'Shipit::Commit'
+
+    validates :number, presence: true, uniqueness: {scope: :stack_id}
+
+    state_machine :merge_status, initial: :fetching do
+      state :fetching
+      state :pending
+      state :rejected
+      state :canceled
+      state :merged
+
+      event :fetched do
+        transition fetching: :pending
+      end
+
+      event :reject do
+        transition pending: :rejected
+      end
+
+      event :cancel do
+        transition pending: :canceled
+      end
+
+      event :complete do
+        transition pending: :complete
+      end
+
+      event :retry do
+        transition %i(rejected canceled) => :pending
+      end
+    end
+
+    def self.request_merge!(stack, number)
+      pull_request = begin
+        create_with(merge_requested_at: Time.now.utc).find_or_create_by!(
+          stack: stack,
+          number: number,
+        )
+      rescue ActiveRecord::RecordNotUnique
+        retry
+      end
+
+      pull_request.schedule_refresh!
+      pull_request
+    end
+
+    def schedule_refresh!
+      RefreshPullRequestJob.perform_later(self)
+    end
+
+    def refresh!
+      update!(github_pull_request: Shipit.github_api.pull_request(stack.github_repo_name, number))
+      head.refresh_statuses!
+      fetched! if fetching?
+    end
+
+    def github_pull_request=(github_pull_request)
+      self.github_id = github_pull_request.id
+      self.api_url = github_pull_request.url
+      self.title = github_pull_request.title
+      self.state = github_pull_request.state
+      self.mergeable = github_pull_request.mergeable
+      self.additions = github_pull_request.additions
+      self.deletions = github_pull_request.deletions
+      self.head = find_or_create_commit_from_github_by_sha!(github_pull_request.head.sha, detached: true)
+    end
+
+    private
+
+    def find_or_create_commit_from_github_by_sha!(sha, attributes)
+      if commit = stack.commits.by_sha(sha)
+        return commit
+      else
+        github_commit = Shipit.github_api.commit(stack.github_repo_name, sha)
+        stack.commits.create_from_github!(github_commit, attributes)
+      end
+    rescue ActiveRecord::RecordNotUnique
+      retry
+    end
+  end
+end
--- a/app/models/shipit/pull_request.rb
+++ b/app/models/shipit/pull_request.rb
@@ -1,5 +1,6 @@
 module Shipit
   class PullRequest < ApplicationRecord
+    WAITING_STATUSES = %w(fetching pending).freeze
     InvalidTransition = Class.new(StandardError)
 
     class StatusChecker < Status::Group
@@ -35,6 +36,7 @@ module Shipit
 
     validates :number, presence: true, uniqueness: {scope: :stack_id}
 
+    scope :waiting, -> { where(merge_status: WAITING_STATUSES) }
     scope :pending, -> { where(merge_status: 'pending') }
     scope :to_be_merged, -> { pending.order(merge_requested_at: :asc) }
 
@@ -54,7 +56,7 @@ module Shipit
       end
 
       event :cancel do
-        transition pending: :canceled
+        transition %i(fetching pending) => :canceled
       end
 
       event :complete do
@@ -81,7 +83,7 @@ module Shipit
       pull_request = begin
         create_with(
           merge_requested_at: now,
-          merge_requested_by: user,
+          merge_requested_by: user.presence,
         ).find_or_create_by!(
           stack: stack,
           number: number,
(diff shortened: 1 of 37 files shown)


casperisfine:
cc @DazWorrall @wvanbergen @pior @etiennebarrie 

The PR is a bit massive and you don't have all the context on Shipit, so I don't expect a deep review. But it could be worth it to give it a quick glance.

The key parts being `MergePullRequestsJob` and the `PullRequest` model.

app/models/shipit/pull_request.rb
```diff
@@ -0,0 +1,184 @@
+module Shipit
+  class PullRequest < ApplicationRecord
```
wvanbergen: Do we keep track of all of them? Or just the ones that were requested to be merged?

app/models/shipit/pull_request.rb
```diff
@@ -0,0 +1,184 @@
...
+    belongs_to :head, class_name: 'Shipit::Commit'
+    belongs_to :merge_requested_by, class_name: 'Shipit::User'
+
+    validates :number, presence: true, uniqueness: {scope: :stack_id}
+
+    scope :waiting, -> { where(merge_status: WAITING_STATUSES) }
+    scope :pending, -> { where(merge_status: 'pending') }
+    scope :to_be_merged, -> { pending.order(merge_requested_at: :asc) }
+
+    state_machine :merge_status, initial: :fetching do
```
wvanbergen: I wonder if it makes sense to have some docs about this state machine and the process behind merging a PR available.

app/models/shipit/unlimited_api_client.rb
```diff
@@ -0,0 +1,10 @@
+module Shipit
+  class UnlimitedApiClient
```
wvanbergen: This is for development purposes?

test/controllers/api/commits_controller_test.rb
```diff
@@ -9,7 +9,7 @@ class CommitsControllerTest < ActionController::TestCase
       end
 
       test "#index returns a list of commits" do
-        commit = @stack.commits.last
+        commit = @stack.commits.reachable.last
```
wvanbergen: Is this related to supporting PRs? Should be have a test for this?

app/models/shipit/pull_request.rb
casperisfine: Ultimately it will be documented in the `README.md`, but for now it's an experimental feature marked as alpha that is likely to change in the future. It won't be documented until it's dimmed stable enough.

app/models/shipit/unlimited_api_client.rb
casperisfine: Yes, I was tired of crafting queries. It's optionally enabled with an environment variable and disabled by default.

test/controllers/api/commits_controller_test.rb
casperisfine: It's a side effect. Commits in Shipit have a `detached: boolean` attribute. When someone force push a tracked branch, some commits we recorded are not on any branch anymore, and we mark them as such.

And since this only happen on force pushes, the last commit of a stack was incidentally never detached.

But Pull Request do reuse the commit model, they point to a `head` that is a detached commit. Because of this it is now possible for the last commit to be detached which broke that assumption, and several tests.

casperisfine:
> Is this for the simple use case (as defined by our RFC), where it simply immediately merges a PR if allowed when a request comes in, and fail otherwise?

Because the process have to be asynchronous anyway, as it's not acceptable to reach to the GitHub API inline.

wvanbergen:
> Because the process have to be asynchronous anyway, as it's not acceptable to reach to the GitHub API inline. 

Makes sense. It also sets us up for later versions where we may not immediately reject, but wait for a suitable merge window.

casperisfine:
> Makes sense. It also sets us up for later versions where we may not immediately reject, but wait for a suitable merge window.

Well, this version only reject PRs that are not mergeable by nature (CI failing or pending, merge conflict, etc).

But if it can't be merged because of externalities (stack locked, branch failing or backlogged), then Shipit will wait for a suitable merge window.

One thing that could change is the rejection of PRs with pending CI status. I though that I don't want people to put PRs in the queue before they are ready, but one could argue it makes sense (i.e. Ship this ASAP please).

wvanbergen:
> But if it can't be merged because of externalities (stack locked, branch failing or backlogged), then Shipit will wait for a suitable merge window.

Should we put a timeout on this? E.g. maximum an hour.

> One thing that could change is the rejection of PRs with pending CI status. I though that I don't want people to put PRs in the queue before they are ready, but one could argue it makes sense (i.e. Ship this ASAP please).

We also have some cases where we want to get PRs (especially reverts) to be deployed ASAP. We can just work around the queue for those cases just now, but might be nice to have an override available for those cases.


Something that is a bit hard for me to figure out is what happens when the stack gets locked. Should we cancel all merge requests?

casperisfine:
> Should we put a timeout on this? E.g. maximum an hour.

That's a good idea. I'd do that in a followup though.

> We also have some cases where we want to get PRs (especially reverts) to be deployed ASAP. We can just work around the queue for those cases just now, but might be nice to have an override available for those cases.

IMO in that scenario we should stick with manual merges. If we need a PR to go out ASAP, I don't see the point asking Shipit when it thinks it's be best to merge it.

Also Shipit is entirely built around the idea of not preventing behaviors, only honest mistakes, so personally I'm not really fond of the idea of restricting merge permissions. But I'm open to it if early experiments show that people can't be trusted.

> Something that is a bit hard for me to figure out is what happens when the stack gets locked. Should we cancel all merge requests?

Nope. The entire point of the queue is to reduce the impact of externalities on the shipping pipeline. If we were to invalidate all the request son lock, that wouldn't be very effective.

So in this current implementation, nothing will be merged while the lock is in place, but as soon as it's released, merging will resume.

wvanbergen:
Initial rollout plan:

1. Add spy command.
2. Tell people they can use this spy command instead of merging themselves?

???

casperisfine:
Yes, I already started the spy integration. But I think the best would be HCTW integration, and it's actually even simpler to implement.

app/models/shipit/stack.rb
```diff
@@ -179,19 +181,23 @@ def head
     end
 
     def merge_status
-      if locked?
-        'locked'
+      return 'locked' if locked?
+      return 'failure' if branch_status == 'failure'
+      if undeployed_commits_count > (maximum_commits_per_deploy || DEFAULT_MAX_COMMITS_PER_DEPLOY) * 1.5
```
etiennebarrie: Do you think it would be worthwhile to override `maximum_commits_per_deploy` to have the default there?

test/fixtures/shipit/pull_requests.yml
```diff
@@ -0,0 +1,46 @@
...
+  merge_status: pending
+  merge_requested_at: <%= 3.minute.ago.to_s(:db) %>
+  github_id: 45454454545454
+  api_url: https://api.github.com/repos/shopify/shipit-engine/pulls/64
+  state: open
+  head_id: 8
+  mergeable: null
+  additions: 23
+  deletions: 43
+
```
etiennebarrie: blank line

test/fixtures/shipit/statuses.yml
```diff
@@ -90,3 +90,12 @@ cyclimse_success_travis:
...
 
+shipit_pending_pr_success_travis:
+  stack: shipit
+  commit_id: 8 # shipit_pending_pr_head
+  description: Super feature
+  context: ci/travis
+  created_at: <%= 9.days.ago.to_s(:db) %>
+  state: success
+  target_url: "http://www.example.com"
+
```
etiennebarrie: whitespace

casperisfine:
> I think we can go further and remove the hidden? and required? methods in Status::Common. allowed_to_fail? is still used it in the view, but it will probably have to change for the queue UI (to show the proper allowed to fail statuses as ignored).

Yeah I'm not super happy with that either. In the end it's because of the disparity between "branch" commits and "PR head" commits. So these methods should probably move in some kind of decorator. But as you said, not a huge deal right now.

Thanks for the review!

PR #645, opened 2017-02-08
Link merge commits and pull requests together

casperisfine:
Followup to: https://github.com/Shopify/shipit-engine/pull/643

One of the main drawback of the merge queue is that now the merge commits are authored by the Shipit user. In itself it's not a big deal, as it's the reality, but it's quite a breaking change for the Shipit API and the Shipit Hooks.

So this makes a few changes for usability:
  - The merge commit will contain a `Merge-Requested-By: @login` field for traceability
  - If the merge commit is identified and match a known pull request, the merge commit will be recorded in Shipit as if it was authored by the merge requester.

Another thing we could eventually do is that Shipit could do the merge on the behalf of the requester. This isn't guaranteed to work as the requester might not have sufficient permissions.

@Shopify/pipeline @etiennebarrie thoughts ?

### DIFF
--- a/app/models/shipit/pull_request.rb
+++ b/app/models/shipit/pull_request.rb
@@ -33,6 +33,7 @@ module Shipit
     belongs_to :stack
     belongs_to :head, class_name: 'Shipit::Commit'
     belongs_to :merge_requested_by, class_name: 'Shipit::User'
+    has_one :merge_commit, class_name: 'Shipit::Commit'
 
     validates :number, presence: true, uniqueness: {scope: :stack_id}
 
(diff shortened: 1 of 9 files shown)


app/models/shipit/pull_request.rb
```diff
@@ -171,6 +172,11 @@ def github_pull_request=(github_pull_request)
       self.head = find_or_create_commit_from_github_by_sha!(github_pull_request.head.sha, detached: true)
     end
 
+    def merge_message
+      return title unless merge_requested_by
+      "#{title}\n\nMerge-Requested-By: #{merge_requested_by.login}\n"
```
wvanbergen: Any reason why we can't use spaces instead of `-` here?

casperisfine: No reason. I just felt like respecting the "header" format: e.g.: https://github.com/torvalds/linux/commit/912964eacb111551db73429719eb5fadcab0ff8a

wvanbergen: Makes sense :+1:

PR #647, opened 2017-02-09
kubernetes keyword in shipit.yml should take precedence over Capfile

KnVerey:
@byroot @Shopify/cloudplatform 
cc @JamesOwenHall 

This fixes the issue where a shipit env that explicitly uses the `kubernetes` keyword will still try to deploy with Capistrano if there is a Capfile in the repo.

### DIFF
--- a/app/models/shipit/deploy_spec/file_system.rb
+++ b/app/models/shipit/deploy_spec/file_system.rb
@@ -2,11 +2,11 @@ module Shipit
   class DeploySpec
     class FileSystem < DeploySpec
       include NpmDiscovery
-      include KubernetesDiscovery
       include PypiDiscovery
       include RubygemsDiscovery
       include CapistranoDiscovery
       include BundlerDiscovery
+      include KubernetesDiscovery
 
       def initialize(app_dir, env)
         @app_dir = Pathname(app_dir)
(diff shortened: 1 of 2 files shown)


n1koo:
Hmm, dumb question but will this break cap deploys though or do we handle that somewhere in kubernetes deploy side? 

> returns kubernetes-deploy command if both capfile and `kubernetes` are present"

casperisfine:
@n1koo the `kubernetes` directive lives in the repository `shipit.yml`, so if you defined it, it's very unlikely you actually wanted Shipit to execute capistrano.

n1koo:
I was thinking of the usecase of `reportifydb` where they legit need both things

casperisfine:
Then they can use `deploy.pre` or `deploy.post`. 

Automatic discovery of the deploy tool, will only ever select one.

KnVerey:
> I was thinking of the usecase of reportifydb where they legit need both things

Do they actually need both things at once, or just to deploy different shipit stacks different ways? If the latter, defining `kubernetes: ` in `shipit.foo.yml` will only affect the "foo" shipit environment.

JamesOwenHall:
> Do they actually need both things at once

No, we're using separate targets for k8s

PR #648, opened 2017-02-09
Delete branches after merging pull requests

casperisfine:
There is no reason to leave them around.

@wvanbergen for review please.

### DIFF
--- a/db/migrate/20170209160355_add_branch_to_pull_requests.rb
+++ b/db/migrate/20170209160355_add_branch_to_pull_requests.rb
@@ -1,0 +1,5 @@
+class AddBranchToPullRequests < ActiveRecord::Migration[5.0]
+  def change
+    add_column :pull_requests, :branch, :string, null: true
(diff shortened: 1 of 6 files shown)


db/migrate/20170209160355_add_branch_to_pull_requests.rb
```diff
@@ -0,0 +1,5 @@
+class AddBranchToPullRequests < ActiveRecord::Migration[5.0]
+  def change
+    add_column :pull_requests, :branch, :string, null: true
```
wvanbergen: Do we need to backfil?

casperisfine: Not really. That feature being alpha, we're the sole users of it, and only used it on a test stask that I will delete.

Also I checked if you try to delete the `nil` branch, it behave like the branch doesn't exist so it's handled fine.

PR #646, opened 2017-02-09
Queue timeout

casperisfine:
Add `merge.timeout` to the `shipit.yml`. Defaults to `1h`.

If a pull request can't be merged in that timeframe, it will be rejected.

Additionally, now that there is a `merge` namespace in `shipit.yml`, `ci.pr.require` and `ci.pr.ignore` have been renamed to `merge.require` and `merge.ignore` respectively.

@wvanbergen for review please.

PR #650, opened 2017-02-10
Rework the pull request timeout feature

casperisfine:
Following up on: https://github.com/Shopify/shipit-engine/pull/646

After thoughts, rejecting a pull request off the queue if it couldn't be merged after a certain amount of time is elapsed doesn't make much sense.

The whole idea behind the merge queue is to free people from overwatching the stack to know when they can merge, and this doesn't help achieve that goal in any way.

This change makes it so that now the PR is only rejected if it could have actually been merged. Which means that it's very likely that it's enqueued back in soon, it's will be merged very shortly.

This allow to leverage the webhooks that is fired when the PR is rejected, to notify the requester, and offer an easy way to enqueue it back.

Summary of the changes:
  - To reflect the new meaning the directive is now: `merge.revalidate_after`
  - It now defaults to no limit.
  - The `rejection_reason` is now `expired`

### DIFF
--- a/app/models/shipit/deploy_spec/file_system.rb
+++ b/app/models/shipit/deploy_spec/file_system.rb
@@ -32,7 +32,7 @@ module Shipit
           'merge' => {
             'require' => pull_request_required_statuses,
             'ignore' => pull_request_ignored_statuses,
-            'timeout' => pull_request_timeout.to_i,
+            'revalidate_after' => revalidate_pull_requests_after.try!(:to_i),
           },
           'ci' => {
             'hide' => hidden_statuses,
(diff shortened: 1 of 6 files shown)


app/models/shipit/deploy_spec/file_system.rb
```diff
@@ -32,7 +32,7 @@ def cacheable_config
           'merge' => {
             'require' => pull_request_required_statuses,
             'ignore' => pull_request_ignored_statuses,
-            'timeout' => pull_request_timeout.to_i,
```
wvanbergen: `&.to_i?`

casperisfine: I'd like Shipit to remain MRI 2.2 compatible for a little more. But ultimately yes.

PR #652, opened 2017-02-15
Refresh PRs before attempting to merge them

casperisfine:
Reported by @berkcaputcu 

Everytime we merge a PR, all the other ones should be considered stale because GitHub will more than likely be checking asynchronously if they are still mergeable.

### DIFF
--- a/app/jobs/shipit/merge_pull_requests_job.rb
+++ b/app/jobs/shipit/merge_pull_requests_job.rb
@@ -13,7 +13,13 @@ module Shipit
       return false unless stack.allows_merges?
 
       pull_requests.select(&:pending?).each do |pull_request|
-        return false unless pull_request.merge!
+        pull_request.refresh!
+        begin
+          pull_request.merge!
+        rescue PullRequest::NotReady
+          MergePullRequestsJob.set(wait: 10.seconds).perform_later(stack)
+          return false
+        end
       end
     end
   end
--- a/app/models/shipit/pull_request.rb
+++ b/app/models/shipit/pull_request.rb
@@ -5,6 +5,7 @@ module Shipit
     WAITING_STATUSES = %w(fetching pending).freeze
     REJECTION_REASONS = %w(ci_failing merge_conflict expired).freeze
     InvalidTransition = Class.new(StandardError)
+    NotReady = Class.new(StandardError)
 
     class StatusChecker < Status::Group
       def initialize(commit, statuses, deploy_spec)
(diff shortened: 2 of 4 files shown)


app/jobs/shipit/merge_pull_requests_job.rb
```diff
@@ -13,7 +13,14 @@ def perform(stack)
       return false unless stack.allows_merges?
 
       pull_requests.select(&:pending?).each do |pull_request|
-        return false unless pull_request.merge!
+        # Any merged PR might conflict with others, so we need to refresh.
+        # If the PR is unchanged, we should get a cache hit anyway.
+        pull_request.refresh!
+        unless pull_request.merge!
+          # That PR isn't mergeable yet, we'll try again a bit later.
+          MergePullRequestsJob.set(wait: 10.seconds).perform_later(stack)
```
wvanbergen: Are we risking an endless loop here if the PR that got merged now legitimately has merge conflicts? Or does the "refresh" make sure it is no longer pending it that case?

casperisfine: The interface isn't very expressive, that is why I added a comment.

`pull_request.merge!` returns a boolean, `true` means the PR was either merged or rejected, and that the job should proceed to the next PR in the queue.

`false` mean the PR can't be merged yet, and the job need to stop and retry later.

casperisfine:
> Would it makes sense to let the job only merge one PR, and reschedule itself if the queue is not empty after 10 seconds?

It could.

app/jobs/shipit/merge_pull_requests_job.rb
wvanbergen: Hmmm, yeah, the boolean return values are kind of confusing. I would personally expect `true` to mean merged, `false` to be rejected, and anything else to be an exception, or not use boolean values. /shrug

If this works, it's fine to leave it like this though.

casperisfine: Yeah I guess not returning anything, and simply raising some `NotReady` exception would be clearer.

casperisfine:
@wvanbergen I refactored the code, should be clearer now.

As for 

> Would it makes sense to let the job only merge one PR, and reschedule itself if the queue is not empty after 10 seconds?

After a bit of though, we can't really guarantee we'd space out merges by an amount of time, since the job could trigger at any time for any reason. SO not sure it's really worth it.

PR #653, opened 2017-02-15
Allow to enable or disable the merge queue

casperisfine:
Step 1 in providing a UI for the merge queue. It need to be enabled on a per stack basis.

### DIFF
--- a/db/migrate/20170215123538_add_merge_queue_enabled_to_stacks.rb
+++ b/db/migrate/20170215123538_add_merge_queue_enabled_to_stacks.rb
@@ -1,0 +1,5 @@
+class AddMergeQueueEnabledToStacks < ActiveRecord::Migration[5.0]
+  def change
+    add_column :stacks, :merge_queue_enabled, :boolean, default: false, null: false
+  end
+end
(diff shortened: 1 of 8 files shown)


db/migrate/20170215123538_add_merge_queue_enabled_to_stacks.rb
```diff
@@ -0,0 +1,5 @@
+class AddMergeQueueEnabledToStacks < ActiveRecord::Migration[5.0]
```
wvanbergen: Curious: what does the `[5.0]` do?

db/migrate/20170215123538_add_merge_queue_enabled_to_stacks.rb
```diff
@@ -0,0 +1,5 @@
...
+  def change
+    add_column :stacks, :merge_queue_enabled, :boolean, default: false, null: false
```
wvanbergen: I assume this table is small enough to not need LHMs.

casperisfine: Indeed. Less than 500 rows. Even shipit biggest tables don't need LHMs.

db/migrate/20170215123538_add_merge_queue_enabled_to_stacks.rb
casperisfine: Since Rails 5.0, migrations are tagged with the Rails version against which they were wrote. So that in the future Rails can do backward incompatible changes to the migration API, without breaking historic migrations.

PR #657, opened 2017-02-17
Refactor and improve live update

casperisfine:
Followup to: https://github.com/Shopify/shipit-engine/pull/656

There is several changes that the merge queue feature introduced, that made the current live update code insufficient:

  - The PR count in the header wasn't part of the live updated area.
  - Since the merge queue page contains a form, and that updating it would lead to a poor UX (your input could be wiped while you type), we need to be able to update several distinct areas.

And also some long standing annoying bugs:
  - The update code was using the URL present in the SSE stream, so it could mess with the emergency mode among other things.
  - Since many changes on the stack trigger asynchronous jobs, which might update the stack again, it can easily result in an thundering herd effect.

This PR solves all this:
  - It now fetch the new document by using `window.location`
  - It supports updating multiple fragments of the page
  - The update is debounced, if many updates are triggered in under 100ms (arbitrary choice), they are collapsed
  - The JS code now check the `Last-Modified` header, so it doesn't update the DOM if it's not required (This doesn't change much now, but it allows to introduce a more fine grained `last-modified` computation in the future).

### DIFF
--- a/app/assets/javascripts/shipit/page_updater.js.coffee
+++ b/app/assets/javascripts/shipit/page_updater.js.coffee
@@ -1,0 +1,63 @@
+class PageUpdater
+  DEBOUNCE = 100
+  RETRY_DELAY = 5000
+  MAX_RETRIES = 5
+
+  @callbacks: []
+  @afterUpdate: (callback) ->
+    @callbacks.push(callback)
+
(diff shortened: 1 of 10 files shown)


byroot:
> Anyone else we could loop in?

@gmalette or @davidcornu maybe ?

Issue #347, opened 2015-05-07
CCMenu support


byroot:
Some people are using CCMenu to know the state of master, so they don't merge in master if the CI is red.

Since stacks on Shipit can aggregate multiple CIs, and that it is also recommended not to merge when the stack is locked, a CCMenu.xml endpoint would be nice.

Travis-CI support as example http://docs.travis-ci.com/user/cc-menu/

cc @cjoudrey @JakeCataford

davidcornu:
@byroot I really want this. 

> The url with the access token would be freely accessible from the settings page of each stacks.

Should each user have a token for each stack?

byroot:
One token per use should enough I think.

edward:
Please yes still relevant

DazWorrall:
Rev 1 of this is on the way: #659

DazWorrall:
@edward @davidcornu @gmalette this has been deployed, you can get the URL from a stack's settings page - give it a try and let us know what you think.

edward:
@DazWorrall here’s the error that I’m running into:

![image](https://cloud.githubusercontent.com/assets/325/23222435/dd41f79e-f8f5-11e6-81ca-6541c648d3a6.png)

DazWorrall:
@edward you have to pick 'Use URL as entered above' - does that work?

edward:
That’s it! SO NICE THANK YOU

I’ll make a PR for being more descriptive with CCMenu setup.

DazWorrall:
@edward do give us some feedback after you've used it for a little while, hopefully it behaves as you expect :)

edward:
Looks great so far!

PR #659, opened 2017-02-20
Add CCMenu/CCTray endpoint

DazWorrall:
Adds a CCMenu/CCTray endpoint, as described in #347. This is my first contribution to shipit and to a rails project at all I think, so there's likely things wrong here - some things to check:

* Is using a named `APIClient` the correct approach here? It does implement everything needed so it felt like a sin to create something that would be so similar. 
* Lazily creating the client in the stack controller doesn't feel correct but it will prevent creating a bunch of tokens unnecessarily - I imagine most people never look at the settings nor would use the endpoint.
* Should the whole thing be put behind a feature flag, so administrators can disable access?

Something else to note - I tested this with CCMenu myself and noticed that the default poll interval was 5 seconds, which I think is pretty aggressive and it wouldn't take long before a few users checking a few stacks might cause a problem, we should document and discourage that default somewhere.

### DIFF
--- a/app/controllers/shipit/api/ccmenu_controller.rb
+++ b/app/controllers/shipit/api/ccmenu_controller.rb
@@ -11,19 +11,10 @@ module Shipit
         def ended_at
           Time.now.utc
         end
-
-        def status
-          false
-        end
-
-        alias_method :running?, :status
-        alias_method :failed?, :status
-        alias_method :error?, :status
-        alias_method :success?, :status
       end
 
       def show
-        latest_deploy = stack.deploys.first || NoDeploy.new
+        latest_deploy = stack.deploys_and_rollbacks.last || NoDeploy.new
         render 'shipit/ccmenu/project.xml.builder', formats: [:xml], locals: {stack: stack, deploy: latest_deploy}
       end
 
--- a/app/controllers/shipit/api/ccmenu_controller.rb
+++ b/app/controllers/shipit/api/ccmenu_controller.rb
@@ -32,6 +32,11 @@ module Shipit
       def stack
         @stack ||= Stack.from_param!(params[:stack_id])
       end
+
+      def authenticate_api_client
+        @current_api_client = ApiClient.authenticate(params[:token])
+        super unless @current_api_client
+      end
     end
   end
 end
--- a/app/controllers/shipit/api/ccmenu_controller.rb
+++ b/app/controllers/shipit/api/ccmenu_controller.rb
@@ -1,0 +1,37 @@
+module Shipit
+  module Api
+    class CcmenuController < BaseController
+      require_permission :read, :stack
+
+      class NoDeploy
+        def id
+          0
+        end
+
+        def ended_at
+          Time.now.utc
+        end
+
+        def status
+          false
+        end
+
+        alias_method :running?, :status
+        alias_method :failed?, :status
+        alias_method :error?, :status
+        alias_method :success?, :status
+      end
+
+      def show
+        latest_deploy = stack.deploys.first || NoDeploy.new
+        render 'shipit/ccmenu/project.xml.builder', formats: [:xml], locals: {stack: stack, deploy: latest_deploy}
+      end
+
+      private
+
+      def stack
+        @stack ||= Stack.from_param!(params[:stack_id])
+      end
+    end
+  end
+end
--- a/app/controllers/shipit/api/ccmenu_controller.rb
+++ b/app/controllers/shipit/api/ccmenu_controller.rb
@@ -1,6 +1,6 @@
 module Shipit
   module Api
-    class CcmenuController < BaseController
+    class CCMenuController < BaseController
       require_permission :read, :stack
 
       class NoDeploy
--- a/app/controllers/shipit/ccmenu_url_controller.rb
+++ b/app/controllers/shipit/ccmenu_url_controller.rb
@@ -1,7 +1,7 @@
 require 'uri'
 
 module Shipit
-  class CcmenuUrlController < ShipitController
+  class CCMenuUrlController < ShipitController
     def fetch
       uri = URI(api_stack_ccmenu_url(stack_id: stack.to_param))
       uri.query = {'token' => client.authentication_token}.to_query
--- a/app/controllers/shipit/ccmenu_url_controller.rb
+++ b/app/controllers/shipit/ccmenu_url_controller.rb
@@ -4,7 +4,7 @@ module Shipit
   class CcmenuUrlController < ShipitController
     def fetch
       uri = URI(api_stack_ccmenu_url(stack_id: stack.to_param))
-      uri.user = client.authentication_token
+      uri.query = {'token' => client.authentication_token}.to_query
       render json: {ccmenu_url: uri.to_s}
     end
 
--- a/app/controllers/shipit/ccmenu_url_controller.rb
+++ b/app/controllers/shipit/ccmenu_url_controller.rb
@@ -1,0 +1,22 @@
+require 'uri'
+
+module Shipit
+  class CcmenuUrlController < ShipitController
+    def fetch
+      uri = URI(api_stack_ccmenu_url(stack_id: stack.to_param))
+      uri.user = client.authentication_token
+      render json: {ccmenu_url: uri.to_s}
+    end
+
+    private
+
+    def client
+      @client ||= Shipit::ApiClient.create_with(permissions: %w(read:stack))
+                                   .find_or_create_by!(creator: current_user, name: 'CCMenu Client')
+    end
+
+    def stack
+      @stack ||= Stack.from_param!(params[:stack_id])
+    end
+  end
+end
--- a/app/controllers/shipit/ccmenu_url_controller.rb
+++ b/app/controllers/shipit/ccmenu_url_controller.rb
@@ -11,8 +11,8 @@ module Shipit
     private
 
     def client
-      @client ||= Shipit::ApiClient.create_with(permissions: %w(read:stack))
-                                   .find_or_create_by!(creator: current_user, name: 'CCMenu Client')
+      @client ||= ApiClient.create_with(permissions: %w(read:stack))
+                           .find_or_create_by!(creator: current_user, name: 'CCMenu Client')
     end
 
     def stack
--- a/app/views/shipit/ccmenu/project.xml.builder
+++ b/app/views/shipit/ccmenu/project.xml.builder
@@ -1,0 +1,13 @@
+# Derived from http://timnew.me/blog/2013/04/07/multiple-project-summary-reporting-standard-cctray-xml-feed/
+status_map = {'backlogged' => 'failure', 'locked' => 'failure'}
+xml.instruct!
+xml.Projects do
+  xml.Project '', {
+    :name => stack.to_param,
+    :lastBuildStatus => status_map.fetch(stack.merge_status, stack.merge_status).capitalize,
+    :activity => deploy.running? ? 'Building' : 'Sleeping',
+    :lastBuildTime => deploy.ended_at || deploy.started_at || deploy.created_at,
+    :lastBuildLabel => deploy.id,
+    :webUrl => stack_url(stack),
+  }
+end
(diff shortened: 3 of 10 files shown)


app/controllers/shipit/ccmenu_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+  class CcmenuController < ActionController::Base
+    before_action :verify_signature, :check_permission
+
+    class NoDeploy
+      def id
+        0
+      end
+
+      def ended_at
+        DateTime.now.utc
```
casperisfine: `DateTime` is rarely used in Ruby, it's almost deprecated. `Time.now.utc` will do just have well.

app/controllers/shipit/ccmenu_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+      end
+
+      alias_method :running?, :status
+      alias_method :failed?, :status
+      alias_method :error?, :status
+      alias_method :success?, :status
+    end
+
+    def show
+      latest_deploy = stack.deploys[0] || NoDeploy.new
```
casperisfine: `stack.deploy[0]` -> `stack.deploy.first`.

The former will load every single records in memory, and then call `Array#[]`. The later will add a `LIMIT 1` to the query and load a single record.

casperisfine:
> I tested this with CCMenu myself and noticed that the default poll interval was 5 seconds, which I think is pretty aggressive and it wouldn't take long before a few users checking a few stacks might cause a problem

It's not that much of a problem honestly. Here Comes the Walrus (chrome extension to inject Shipit merge status into GitHub for any external reader wondering) has a 2 seconds poll interval, and generate quite a few RPM. But it implement HTTP caching (Last-Modified), so in the end the average response time is only 8ms. Hopefully CCMenu respect HTTP caching as well, if it does, 5 seconds poll interval is absolutely not a problem.

If it doesn't, then we'll have to do server side caching. No huge deal.

DazWorrall:
> If it doesn't, then we'll have to do server side caching. No huge deal.

👍 

@casperisfine Do you think it makes sense to have the xml generation as an api endpoint? It would be I think the only endpoint to return xml, but the authentication is going to be identical now though so it doesn't make sense to duplicate. Or would you prefer to just keep this separate, I'll still need a small endpoint to retrieve the ccmenu url after all, though that request will have a cookie.

Also, in the #347 you say one token per user would be sufficient, but above you say one per stack and per user - which would you prefer?

casperisfine:
> Do you think it makes sense to have the xml generation as an api endpoint?

Yes. It's a very specific API, but an API noneless. The fact that it returns XML doesn't bother me much.

> Also, in the #347 you say one token per user would be sufficient, but above you say one per stack and per user - which would you prefer?

I'm honestly not really settled on this. But on first though, one token per user would mean users could craft CCMenu urls for other stacks without going through the UI again. So I think I'd prefer that.

app/controllers/shipit/stacks_controller.rb
```diff
@@ -39,6 +39,11 @@ def destroy
     end
 
     def settings
+      client = Shipit::ApiClient.find_by(creator: current_user, name: "#{current_user.name} ccmenu client")
+      client ||= Shipit::ApiClient.create!(
```
casperisfine: Forgot to mention it:

```ruby
Shipit::ApiClient.create_with(permissions: %w(read:stack)).find_or_create_by!(creator: current_user, name: 'CCMenu Client')
```

DazWorrall: Ah thank you, putting permissions in the find_or_create_by was getting me `SerializationTypeMismatch`, couldn't figure out the foo :)

casperisfine: Yeah permission (for the worse or the better) is a serialized field, as such it can't be used for querying, only inserting. And `find_or_create_by()` uses the arguments for both querying and inserting.

app/controllers/shipit/ccmenu_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+      render 'shipit/ccmenu/project.xml.builder', formats: [:xml], locals: {stack: stack, deploy: latest_deploy}
+    end
+
+    private
+
+    def verify_signature
+      head(422) unless @client = Shipit::ApiClient.authenticate(params[:token])
```
wvanbergen: Is a 422 normal in these situations?

app/views/shipit/ccmenu/project.xml.builder
```diff
@@ -0,0 +1,26 @@
...
+  if deploy.running?
+    last_status = 'pending'
+  elsif deploy.failed?
+    last_status = 'Failure'
+  elsif deploy.error?
+    last_status = 'Exception'
+  elsif deploy.success?
+    last_status = 'Success'
+  end
+end
```
wvanbergen: Not sure if builder supports it, but things like this usually go into a view helper: http://guides.rubyonrails.org/action_view_overview.html

test/controllers/ccmenu_controller_test.rb
```diff
@@ -0,0 +1,54 @@
+require 'test_helper'
+
+module Shipit
+  class CcmenuControllerTest < ActionController::TestCase
+    def get_project_from_xml(xml)
+      Hash.from_xml(xml)['Projects']['Project']
+    end
```
wvanbergen: style nitpick: I usually move these to the bottom of the test class, and mark them as `private`.

app/views/shipit/ccmenu/project.xml.builder
wvanbergen: Also, in Ruby everything is an expression, so you can do:

``` ruby
last_status = if stack.locked?
  'Failure'
else
  if deploy.running?
    'pending'
  ....
```

or wrap it in a function even:

``` ruby
def last_status
if stack.locked?
  'Failure'
else
  if deploy.running?
    'pending'
  ....
end

**Comment by DazWorrall** [comment]
Ok, just discovered CCMenu doesn't parse a URI with a username in it, so I've added support for authentication via query string just for the ccmenu endpoint, happy @casperisfine?

**Comment by casperisfine** [comment]
> so I've added support for authentication via query string just for the ccmenu endpoint, happy @casperisfine?

Fine by me.

File: app/controllers/shipit/api/ccmenu_controller.rb [hunk 7]
```diff
@@ -0,0 +1,42 @@
...
+        end
+
+        alias_method :running?, :status
+        alias_method :failed?, :status
+        alias_method :error?, :status
+        alias_method :success?, :status
+      end
+
+      def show
+        latest_deploy = stack.deploys.first || NoDeploy.new
```
casperisfine: `stack.deploys` default ordering is `id ASC`, so if you want the most recent deploy, you need to use `stack.deploys.last`.

Also you likely want rollbacks as well, so `deploys_and_rollbacks` would be a better association: https://github.com/Shopify/shipit-engine/blob/e4162e2f0cd263723305e77eb5ccb76cbde09c5d/app/models/shipit/stack.rb#L31

app/controllers/shipit/api/ccmenu_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+        end
+
+        def status
+          false
+        end
+
+        alias_method :running?, :status
+        alias_method :failed?, :status
+        alias_method :error?, :status
+        alias_method :success?, :status
```
casperisfine: I don't think you need those status related methods anymore.

app/controllers/shipit/ccmenu_url_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    def fetch
+      uri = URI(api_stack_ccmenu_url(stack_id: stack.to_param))
+      uri.query = {'token' => client.authentication_token}.to_query
+      render json: {ccmenu_url: uri.to_s}
+    end
+
+    private
+
+    def client
+      @client ||= Shipit::ApiClient.create_with(permissions: %w(read:stack))
```
casperisfine: Nitpick, since you are in the `Shipit` namespace alreasy, you can reference `ApiClient` directly.

DazWorrall:
Ok, this is ready for another review then :) The actual xml payload may need tweaking after feedback from users but I hope the code is now acceptable.

app/views/shipit/stacks/settings.html.erb
```diff
@@ -80,3 +90,11 @@
 
   </section>
 </div>
+<script type="text/javascript">
+function fetchCCMenuURL() {
+  $.ajax({url: "<%= ccmenu_url_url(stack_id: @stack.to_param) %>", success: function(result){
+    console.log(result);
+    $("#ccmenu-url").html("<input type='text' disabled value='" + result.ccmenu_url + "'/>");
```
casperisfine: You forgot a `console.log` here.

Also I'm not a huge fan of inline JS in general, and this particular instance is vulnerable to XSS (granted you display some relatively restricted value, so it's likely not exploitable, but as a good practice it's better to never bypass the XSS security layers).

I suggest you put the disabled input in the DOM from the beginning, but with a HTML class that hides it. Then your JS populates it, and remove then `hidden` class. As for the URL, you can store it in the DOM, so that the script doesn't have to craft it.

app/views/shipit/stacks/settings.html.erb
```diff
@@ -71,6 +71,16 @@
...
 
+    <div class="setting-section">
+      <h5>Miscellaneous</h5>
+      <table>
+        <tr>
+          <td>CCTray/CCMenu URL</td>
+          <td id="ccmenu-url" style="width: 100%"><%= button_to "Fetch URL", "javascript:fetchCCMenuURL()", class: 'btn' %></td>
+        </tr>
+      </table>
+        <%#input type="text" disabled value="<%= ccmenu_url(stack_id: @stack.to_param, token: @cc_token) >"%>
```
casperisfine: 🔥

app/views/shipit/stacks/settings.html.erb
```diff
@@ -71,6 +71,16 @@
       </table>
     </div>
 
+    <div class="setting-section">
+      <h5>Miscellaneous</h5>
+      <table>
+        <tr>
+          <td>CCTray/CCMenu URL</td>
+          <td id="ccmenu-url" style="width: 100%"><%= button_to "Fetch URL", "javascript:fetchCCMenuURL()", class: 'btn' %></td>
```
casperisfine: Please avoid inline style and inline scripts.

app/views/shipit/stacks/settings.html.erb
DazWorrall: > As for the URL, you can store it in the DOM, so that the script doesn't have to craft it.

Not sure I understand here - do you mean generate the url on page load and avoid the ajax entirely? We would probably create client unnecessarily then so I assume you mean something else?

byroot: I meant the URL of the endpoint that this JS has to hit. `ccmenu_url_url(stack_id: @stack.to_param)` in this case.

Something like this:

```ruby
<%= link_to "Get CCMenu URL", ccmenu_url_url(stack_id: @stack.to_param), class: 'fetch-ccmenu-url' %>
```

```coffeescript
$(document).on 'click', '.fetch-ccmenu-url', (event) ->
  event.preventDefault()
  # Disable button
  $.get(event.target.href).done((token) ->
    $('#ccmenu-url').text(token).removeClass('hidden')
    # Remove button entirely
  ).fail(->
    # Reactivate the button somehow
  )
  
  
```

DazWorrall:
@casperisfine @wvanbergen this is ready for review again I think!

casperisfine:
Code looks good to me know. I'll try it locally.

app/controllers/shipit/api/ccmenu_controller.rb
```diff
@@ -0,0 +1,33 @@
+module Shipit
+  module Api
+    class CcmenuController < BaseController
```
wvanbergen: I think you can register `CCMenu` as acronym, which means you can capitalize it: http://api.rubyonrails.org/classes/ActiveSupport/Inflector/Inflections.html#method-i-acronym

There's usually an initializer (in `config/initializers`) that sets acronyms, irregular plurals and such,

wvanbergen:
🎉

PR #667, opened 2017-02-28
Synchronously increment `undeployed_commits_count` on merge

DazWorrall:
Prevents `undeployed_commits_count` becoming stale.

/cc @byroot

### DIFF
--- a/app/models/shipit/pull_request.rb
+++ b/app/models/shipit/pull_request.rb
@@ -95,6 +95,10 @@ module Shipit
       before_transition any => :pending do |pr|
         pr.revalidated_at = Time.now.utc
       end
+
+      before_transition %i(pending) => :merged do |pr|
+        Stack.increment_counter(:undeployed_commits_count, pr.stack_id)
+      end
     end
 
     def self.schedule_merges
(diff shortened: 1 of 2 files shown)


app/models/shipit/pull_request.rb
```diff
@@ -163,6 +163,8 @@ def merge!
         commit_message: 'Merged by Shipit',
         merge_method: 'merge',
       )
+      stack.undeployed_commits_count += 1
```
casperisfine: This way of incrementing a column is heavily subject to race conditions. Rails provides an API for atomic increments of db columns: http://api.rubyonrails.org/classes/ActiveRecord/CounterCache/ClassMethods.html#method-i-increment_counter

casperisfine: Also I believe this should be done as part of the `complete!` state transition, that way they'll be together in a single transaction.

DazWorrall: > Also I believe this should be done as part of the `complete!` state transition
Can you share an example of that? I see the `event :complete do` block but not sure if/how I'd access instance information inside it.

byroot: You can use before transition blocks: https://github.com/Shopify/shipit-engine/blob/d42d6c5f13bf4102152cbdcc15bf70c7e4a71958/app/models/shipit/pull_request.rb#L87

DazWorrall: Fixed, ty!

PR #668, opened 2017-03-01
Set html lang attribute in layout

casperisfine:
Hopefully that should fix:

image

@disaacs for review please.

### DIFF
--- a/app/views/layouts/shipit.html.erb
+++ b/app/views/layouts/shipit.html.erb
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html data-controller="<%= controller_name %>" data-action="<%= action_name %>">
+<html lang="<%= I18n.locale %>" data-controller="<%= controller_name %>" data-action="<%= action_name %>">
 <head>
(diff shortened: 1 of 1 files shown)


disaacs:
👍  Worth a shot. No harm in trying.

However I am skeptical. Only the ShipIt home pages shows the translate prompt, none of the specific stack pages do, even though they also have no lang attribute.

casperisfine:
Yes, but chrome's lang detector only triggers if the language declared is different from your system language, or is not declared at all.

On other pages, there is probably nothing that makes chrome think it's not english, whereas some stack name must look like Danish for Chrome.

Edouard-chin:
Worth trying, but I'm skeptical too. We should maybe just disable the chrome feature
```
<meta name="google" content="notranslate" />
```

casperisfine:
@Edouard-chin oops, I merged just before you answered.

I'd prefer for avoid browser specific tags if possible, but if it doesn't work I'll do that. Thanks for the input.

disaacs:
Just tested it in production. It works. I am no longer getting the translate prompt.

PR #669, opened 2017-03-03
Don't automatically deploy a commit without it's revert

fw42:
This PR tries to ensure that a commit that has been reverted can only be automatically deployed if the deploy contains the revert as well.

@tjoyal @byroot @daniellaniyo

### DIFF
--- a/app/models/shipit/deploy.rb
+++ b/app/models/shipit/deploy.rb
@@ -36,6 +36,11 @@ module Shipit
 
     delegate :broadcast_update, :filter_deploy_envs, to: :stack
 
+    def reload(*args)
+      @commits = @commits_since = @default_since_commit_id = nil
+      super
+    end
+
     def build_rollback(user = nil, env: nil)
       Rollback.new(
         user_id: user.try!(:id),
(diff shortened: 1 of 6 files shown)


app/models/shipit/deploy.rb
```diff
@@ -136,6 +141,27 @@ def confirmed?
...
+      return if commits_since.empty?
+
+      commits_since.each do |commit|
+        matches = commit.message.match(/^\s*Revert "(.*)"\s*$/)
+        next unless matches
+        parent_message = matches[1]
+
+        commits.each do |other_commit|
+          next if commit == other_commit
+          return true if parent_message == other_commit.message
```
fw42: Hmmm now that I look at this again, I think this needs to be a bit smarter and ignore the "Merge pull request ..." prefix that the parent might have. Will update this PR on Monday.

casperisfine:
I wouldn't do it like this.

Right now you give the Deploy model the responsibility to bail out if the commit range it's supposed to deploy is somehow invalid. 

Imagine the following master:

  - revert aaaa -> pending
  - aaaa -> success
  - bbbb -> success
  - cccc -> success

With this PR, the continuous delivery job will attempt to deploy the `aaaa...cccc` commit range, only to receive a refusal from the deploy object.

However it could (and should) have deployed the `bbbb...cccc` range.

This kind of concerns is currently owned by the `Commit` and `UndeployedCommit` classes, they are responsible to determine if they are themselves safe to ship or not.

You'd need a small refactor so that `UndeployedCommit` can inspect the commit range they are part of, but it doesn't seem very hard.

PR #672, opened 2017-03-09
Use default ENV variables when a task is triggered from API

kirs:
I've noticed that when the restart task is triggered by API with a `spy` command, the deploy script doesn't get the default ENV variables, unlike when the task is triggered from the browser.

IMO, when you declare something like:

```
tasks:
  restart:
    action: "Restart Application"
    steps:
      - bin/cap $ENVIRONMENT deploy:restart --trace
    variables:
      -
        name: MYVARIABLE
        title: blahblah
        default: "1"
```

You rely on the fact that `MYVARIABLE` is always present, not only when the task is triggered from the web.

This PR fixes the edge case and extends the test coverage.

review @DazWorrall @casperisfine

### DIFF
--- a/app/models/shipit/task_definition.rb
+++ b/app/models/shipit/task_definition.rb
@@ -49,7 +49,7 @@ module Shipit
     end
 
     def variables_with_defaults
-      @variables_with_defaults ||= variables.select { |v| v.default.present? }
+      @variables_with_defaults ||= variables.select(&:default_provided?)
     end
 
     private
--- a/app/models/shipit/task_definition.rb
+++ b/app/models/shipit/task_definition.rb
@@ -48,6 +48,10 @@ module Shipit
       EnvironmentVariables.with(env).permit(variables)
     end
 
+    def variables_with_defaults
+      @variables_with_defaults ||= variables.select { |v| v.default.present? }
+    end
+
     private
 
     def task_variables(config_variables)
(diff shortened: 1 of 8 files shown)


test/dummy/db/schema.rb
```diff
@@ -189,24 +189,24 @@
   end
 
   create_table "tasks", force: :cascade do |t|
-    t.integer  "stack_id",              limit: 4,                         null: false
-    t.integer  "since_commit_id",       limit: 4,                         null: false
-    t.integer  "until_commit_id",       limit: 4,                         null: false
-    t.string   "status",                limit: 10,    default: "pending", null: false
+    t.integer  "stack_id",              limit: 4,                            null: false
```
casperisfine: You are not supposed to have any schema change.

app/models/shipit/task_definition.rb
```diff
@@ -48,6 +48,10 @@ def filter_envs(env)
       EnvironmentVariables.with(env).permit(variables)
     end
 
+    def variables_with_defaults
+      @variables_with_defaults ||= variables.select { |v| v.default.present? }
```
casperisfine: What if the default is a blank string?

```ruby
>> ' '.present?
=> false
```

kirs: I don't think that someone will add a default ENV variable with `' '` as a default value, but I think we can stop doing `to_s` in [models/shipit/variable_definition.rb](https://github.com/Shopify/shipit-engine/blob/master/app/models/shipit/variable_definition.rb#L8) and we'll know when the default value is `nil`. I'll update the PR.

kirs:
Updated the PR 👌

PR #674, opened 2017-03-14
Expose last deployed time via stack endpoint

lalithr95:
Shipit api doesn't provide last deployment time through stack endpoint. This PR exposes last deployed time for each stack. This time would be useful in [services-db](https://github.com/Shopify/services-db/pull/871) to keep track of actively deployed services. Updated minor piece of unrelated code in stack mode and added `@stack.reload` in commit test since `@stack.commits.new` is going to instantiate new commit which doesn't have any foreign key data assigned.

@byroot @wvanbergen

### DIFF
--- a/app/models/shipit/deploy.rb
+++ b/app/models/shipit/deploy.rb
@@ -183,7 +183,7 @@ module Shipit
     end
 
     def update_last_deploy_time
-      stack.update(deployed_at: updated_at)
+      stack.update(last_deployed_at: ended_at)
     end
   end
 end
--- a/app/models/shipit/deploy.rb
+++ b/app/models/shipit/deploy.rb
@@ -10,6 +10,7 @@ module Shipit
       after_transition to: :success, do: :update_undeployed_commits_count
       after_transition to: :aborted, do: :trigger_revert_if_required
       after_transition any => any, do: :update_commit_deployments
+      after_transition any => any, do: :update_last_deploy_time
     end
 
     has_many :commit_deployments, dependent: :destroy, inverse_of: :task, foreign_key: :task_id do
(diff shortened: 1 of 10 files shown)


test/dummy/db/schema.rb
```diff
@@ -174,6 +174,7 @@
     t.datetime "continuous_delivery_delayed_since"
     t.datetime "locked_since"
     t.boolean  "merge_queue_enabled",                             default: false,        null: false
+    t.datetime "deployed_at"
```
casperisfine: I'd name it `last_deployed_at`.

app/models/shipit/deploy.rb
```diff
@@ -180,5 +181,9 @@ def last_successful_deploy
     def update_undeployed_commits_count
       stack.update_undeployed_commits_count(until_commit)
     end
+
+    def update_last_deploy_time
+      stack.update(deployed_at: updated_at)
```
casperisfine: I'd use the `ended_at` timestamp.

lalithr95: Good one 👍

lalithr95:
> Should we backfill the data?

No need to backfill column data, this way we could start with fresh data.
Fixed comments and added column to `StackSerializer`

test/models/deploys_test.rb
```diff
@@ -225,6 +225,13 @@ def setup
       end
     end
 
+    test "transitions to any state updates last deploy time to stack record" do
+      @deploy = shipit_deploys(:shipit_running)
+      @deploy.complete!
+      @stack.reload
+      assert_in_delta @deploy.updated_at, @stack.last_deployed_at, 1
```
casperisfine: That test should be updated.

Issue #270, opened 2014-11-13
Allow setting the 'next deployable commit'


jstorimer:
It happens often enough that picking which commit to deploy is critical.

This can happen if we deploy bad code, do a rollback, then need to make sure the next deploy includes the revert lest we introduce the problem again. We'd want to pick the revert as the 'next deployable commit' and not let anyone deploy a rev below that.

This also happens frequently with including / excluding servers from deploy rotation.

PR #675, opened 2017-03-17
Don't automatically deploy a commit without it's revert

fw42:
Alternative implementation of https://github.com/Shopify/shipit-engine/pull/669.

@kmcphillips @casperisfine, thoughts?

### DIFF
--- a/app/models/shipit/commit.rb
+++ b/app/models/shipit/commit.rb
@@ -109,6 +109,21 @@ module Shipit
       success? || stack.ignore_ci?
     end
 
+    def safe_to_deploy?
+      !revert_being_left_behind?
+    end
+
+    def revert_being_left_behind?
+      revert_commits = stack.commits.reachable.newer_than(id).where("LOWER(message) LIKE '%revert%'")
+
+      revert_commits.each do |revert_commit|
+        next unless matches = revert_commit.message.match(/^\s*Revert "(.*)"\s*$/m)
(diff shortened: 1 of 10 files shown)


casperisfine:
It's better, but it still has a couple issues:

  - The revert detection is quite fragile
  - There is no way for a human to overrule it.
  - It requires relatively heavy queries and computations to generate the `stacks#show` page.

I'll believe I'll go with something along the line of https://github.com/Shopify/shipit-engine/issues/270

I'm thinking adding a boolean `locked` attribute on `Commit`. When a revert reaches master, if the commit it reverts is not deployed yet, then all the range will get the `locked` property.

In additions, on the `stacks#show` page, each undeployed commit will get a button to toggle locks on commits. That way we get both automated security, and fine grained manual management.

app/models/shipit/commit.rb
```diff
@@ -109,6 +109,21 @@ def deployable?
       success? || stack.ignore_ci?
     end
 
+    def safe_to_deploy?
```
fw42: maybe this should say "safe_to_deploy_automatically?"

fw42:
> I'll believe I'll go with something along the line of #270

ok, fair enough. I'll leave this alone for now then and give you some time to experiment with your ideas. Let me know if you need me to revisit this.

casperisfine:
👌

PR #676, opened 2017-03-20
Implement locking of commits

casperisfine:
Fix: https://github.com/Shopify/shipit-engine/issues/270
Closes: https://github.com/Shopify/shipit-engine/pull/669
Closes: https://github.com/Shopify/shipit-engine/pull/675

capture d ecran 2017-03-20 a 14 37 33

Every undeployed commit get a little lock icon. It's greyed by default, but you can toggle it to red by clicking on it. A locked commit won't be automatically deployed, and require emergency mode to be deployed manually.

In a followup, I'll look at automatically locking commits when a revert is merged, so that it addresses #669 / #675. 

As usual the UI is quite ugly, so suggestions are welcome.

@Shopify/pipeline @fw42 @tjoyal @daniellaniyo thoughts?

### DIFF
--- a/app/models/shipit/undeployed_commit.rb
+++ b/app/models/shipit/undeployed_commit.rb
@@ -11,7 +11,7 @@ module Shipit
       state = deployable? ? 'allowed' : status.state
       unless bypass_safeties
         state = 'deploying' if stack.active_task?
-        state = 'locked' if stack.locked?
+        state = 'locked' if locked?
       end
       state
     end
@@ -19,7 +19,6 @@ module Shipit
     def redeploy_state(bypass_safeties = false)
       state = 'allowed'
       unless bypass_safeties
-        state = 'locked' if stack.locked?
         state = 'deploying' if stack.active_task?
       end
       state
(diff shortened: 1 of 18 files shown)


app/models/shipit/undeployed_commit.rb
```diff
@@ -11,15 +11,14 @@ def deploy_state(bypass_safeties = false)
       state = deployable? ? 'allowed' : status.state
       unless bypass_safeties
         state = 'deploying' if stack.active_task?
-        state = 'locked' if stack.locked?
+        state = 'locked' if locked?
```
fw42: is the `stack.locked?` not important anymore? maybe `locked? || stack.locked?` would be better?

casperisfine: Good question.

I think `Locked` was always a bit confusing to people. So I though, might as well make the stack level lock not change the button caption, since it's obvious enough already.

casperisfine:
> it's usually a range of commits/merges that you'd want to mark as not deployable.

Indeed. And I though about that. The reason I went with he current solution is:

  - The stack page being refreshed automatically, only atomic user interactions are possible (otherwise you'd need to lock the refresh loop, which adds a lot of complexity which I'd prefer to avoid).
  - You need to disable a big range when you revert something, hopefully this will be automated in a followup.

PR #677, opened 2017-03-21
Automatically lock commits on revert

casperisfine:
Followup https://github.com/Shopify/shipit-engine/pull/676

- Detect merged reverts, both manual and from GitHub UI, as long as the default commit message is unmodified.
- When a revert is merged (or pushed), scan the undeployed commits to lock all the entire commit range between the revert and the reverted commit (included).

@Shopify/pipeline @fw42 @tjoyal @daniellaniyo for review please.

### DIFF
--- a/app/jobs/shipit/github_sync_job.rb
+++ b/app/jobs/shipit/github_sync_job.rb
@@ -17,11 +17,22 @@ module Shipit
         @stack.transaction do
           shared_parent.try!(:detach_children!)
           new_commits.each do |gh_commit|
-            @stack.commits.create_from_github!(gh_commit)
+            append_commit(gh_commit)
           end
         end
       end
       CacheDeploySpecJob.perform_later(@stack)
+    end
+
+    def append_commit(gh_commit)
+      appended_commit = @stack.commits.create_from_github!(gh_commit)
+      if appended_commit.revert?
+        impacted_commits = @stack.undeployed_commits.reverse.drop_while { |c| !appended_commit.revert_of?(c) }
+        impacted_commits.pop # appended_commit
+        impacted_commits.each do |impacted_commit|
+          impacted_commit.update!(locked: true)
+        end
+      end
     end
 
     def fetch_missing_commits(&block)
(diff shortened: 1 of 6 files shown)


tjoyal:
To me it seems if I'm reverting something that has already beed deployed, it'll lock all the the commits that aren't deployed (even if there is no need as the original commit is already on production).

Was that the desired effect?

Also: this can't be done comparing commit content instead of the text?

casperisfine:
> if I'm reverting something that has already beed deployed, it'll lock all the the commits that aren't deployed

No. The code will only look for the reverted commit in the undeployed ones. If you revert something that is already deployed, it won't have any effect.

> Also: this can't be done comparing commit content instead of the text?

It would be way more complicated, and not guaranteed to be more accurate.

test/jobs/github_sync_job_test.rb
```diff
@@ -36,6 +36,55 @@ class GithubSyncJobTest < ActiveSupport::TestCase
       assert shipit_commits(:fifth).reload.detached?
     end
 
+    test "#perform lock all commits leading to a revert" do
```
fw42: s/lock/locks/ 🇫🇷

app/jobs/shipit/github_sync_job.rb
```diff
@@ -17,13 +17,24 @@ def perform(params)
...
         end
       end
       CacheDeploySpecJob.perform_later(@stack)
     end
 
+    def append_commit(gh_commit)
+      appended_commit = @stack.commits.create_from_github!(gh_commit)
+      if appended_commit.revert?
+        impacted_commits = @stack.undeployed_commits.reverse.drop_while { |c| !appended_commit.revert_of?(c) }
+        impacted_commits.pop # appended_commit
```
daniellaniyo: So here the revert is popped so that all the other commits underneath do not ship without it right?

casperisfine: The pop is simply for this working state. Wehn we load the undeployed commits, the revert was just added on top. We don't want to lock the revert itself, so we pop it from the array.

PR #678, opened 2017-03-21
Add tooltips and confirmations to lock on commit

jules2689:
Using the new lock feature, I was a bit confused as to what was going on. When I hovered over the link, it appeared as a normal link. Little did I know that JS overrode this and performed a PATCH request, leading me to accidentally unlocking a commit.

To resolve that confusion, I added a tooltip to the lock and a confirmation to unlock..

shipit

cc @byroot @wvanbergen

### DIFF
--- a/app/assets/javascripts/shipit/stacks.js.coffee
+++ b/app/assets/javascripts/shipit/stacks.js.coffee
@@ -4,19 +4,8 @@ $(document).on 'click', '.commit-lock a', (event) ->
   $link = $(event.target).closest('a')
 
   locked = $commit.hasClass('locked')
-  new_toolip = ''
-
-  if locked
-    confirmation = confirm("Mark this commit as safe to deploy?");
-    unless confirmation
-      return
-    new_toolip = 'This commit is safe to deploy. Click to mark it as unsafe'
-  else
-    new_toolip = 'This commit is unsafe to deploy. Click to mark it as safe'
-
-  $link.attr('data-tooltip', new_toolip) # Change DOM
-  $link.data('tooltip', new_toolip) # Break data cache
   $commit.toggleClass('locked')
+
   $.ajax($link.attr('href'), method: 'PATCH', data: {commit: {locked: !locked}})
 
 jQuery ($) ->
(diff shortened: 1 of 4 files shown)


byroot:
I'm 👎 for that confirmation box. You can easily have to lock 10 or more commits, confirming each ones would be very bad UX wise.

I'm not against the tooltip, but I considered it in the original PR but went without it because of the update latency.

Your PR solves it somehow, but the duplication of wording, as well as wording inside JS is not acceptable. A better way to do it would be to have both tooltips in the DOM and to switch them over.

jules2689:
Why do you need to confirm 10 different commits? That seems like a useless thing to do to me. Can't you just confirm the latest one you want to deploy and shipit will deploy all up until that one?

Confirmation dialogs are very important on potentially destructive actions.

casperisfine:
> Why do you need to confirm 10 different commits?

Because the lock only impact a single commit. It's not rare to have to lock a commit range.

> shipit will deploy all up until that one?

No, shipit will simply not consider locked commits for deploy. It can pick the one right after it.

> Confirmation dialogs are very important on potentially destructive actions.

That action is all but destructive. It's actually the inverse of destructive as all it can do is prevent something from happening.

wvanbergen:
> That action is all but destructive. It's actually the inverse of destructive as all it can do is prevent something from happening.

So the potentially destructive part is disabling the lock, e.g. marking it safe to deploy again. Not sure if `confirm` is the right way, but I fear that shipit may immediately start to deploy a commit if you accidentally remove a lock.

casperisfine:
> I fear that shipit may immediately start to deploy a commit if you accidentally remove a lock.

Then you have time to abort it.

Dunno. I really fell like that confirm step will be more a burden than anything else. Shipit is a sharp tool, I think we need to accept that.

jules2689:
The confirm only happens if you are locked. example:

- commit 1: locked
- commit 2: locked
- commit 3: locked

We only need to unlock commit 1 for 2,3 to deploy. This however causes commits to immediately deploy when auto-deploy is enabled. To confirm this, we can present the modal asking the user to verify that this was the intention. Since we only need to unlock one commit for the others to deploy, the impact to the experience is negligible but saves confusion and prevents people from getting nervous.

byroot:
> The confirm only happens if you are locked

Oh. That's what I missed. Then yes I'm fine with the confirm in that case. Please re-use https://github.com/Shopify/shipit-engine/issues/679 as a base implementation, and avoid user readable strings in JS.

Issue #680, opened 2017-03-22
Inefficient stack.destroy!


szemek:
Today with colleague we came across an issue:
```
ActiveRecord::StatementInvalid: Mysql2::Error: Lock wait timeout exceeded;
try restarting transaction: DELETE FROM `commits` WHERE `commits`.`id` = 131
```
after trying to delete a stack.
In our production environment we deploy quite frequently. In database we had more than 350 thousand of all records related to that stack.

As I checked, simple `destroy!` is performed and `destroy` is called on dependent records 
[app/jobs/shipit/destroy_stack_job.rb](https://github.com/Shopify/shipit-engine/blob/13e328ee6e404853c856ca8bd2c4200b43fc1d9e/app/jobs/shipit/destroy_stack_job.rb#L6)
[app/models/shipit/stack.rb](https://github.com/Shopify/shipit-engine/blob/13e328ee6e404853c856ca8bd2c4200b43fc1d9e/app/models/shipit/stack.rb#L26-L34)

For records, that don't have special behaviour connected to destroy, it should be possible to remove them in batch.

I'd like to help with pull request, but I'm not sure whether it is possible to configure ActiveRecord models to behave more efficiently or custom implementation in DestroyStackJob will be needed.

casperisfine:
I've hit that issue a couple times too. A pull request would be most welcome.

I think what you could do is:
  - First "soft delete" the stack.
  - Then trigger a background job that would delete the stack in several batches.

One thing to worry about is that there is a unicity constraint on `org/repo/env`, so while the stack is being deleted, a similar one can't be created. Not sure if it's really a big issue.

szemek:
@casperisfine I prepared pull request with proof of concept. Let me know what do you think.

PR #681, opened 2017-03-22
More explicit removal of dependent records when destroying a stack

szemek:
PR related to issue https://github.com/Shopify/shipit-engine/issues/680

I replaced implicit calls to `#destroy` with explicit `.delete_all` or `.destroy_all` (for `Shipit::GithubHook`).
It should reduce total amount of calls to database and speed up removal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Unreleased
 
-* Automatically lock impacted commits when a revert is merged. This include the reverted commit as well as all its children up to the revert. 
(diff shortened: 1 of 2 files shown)


casperisfine:
Looks good to me. I'm not a huge fan of this kind of code in general. But in this instance it makes sense, and even if we introduce a new model and forget to add it here, the final `stack.destroy` will act as a catch all.

Please add a CHANGELOG entry, and I'll merge.

byroot:
Thank you!

PR #682, opened 2017-03-28
Bumping ruby version

kyledurand:
@kirs not sure about this travis file but shouldn't be an issue to bump the rvm version as well right?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,12 +14,12 @@ env:
     - SUITE=pg
     - SUITE=sqlite3
 rvm:
-  - 2.2.6
+  - 2.2.3
   - 2.3.3
 matrix:
   fast_finish: true
   include:
-    - rvm: 2.3.3
+    - rvm: 2.2.3
       env: SUITE=rubocop
 
 notifications:
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,11 +14,11 @@ env:
     - SUITE=pg
     - SUITE=sqlite3
 rvm:
-  - 2.2.3
+  - 2.3.1
 matrix:
   fast_finish: true
   include:
-    - rvm: 2.2.3
+    - rvm: 2.3.1
       env: SUITE=rubocop
 
 notifications:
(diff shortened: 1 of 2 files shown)


dev.yml
```diff
@@ -7,7 +7,7 @@ up:
     - openssl
     - sqlite
   - ruby:
-      version: 2.2.3p172-shopify
+      version: 2.3.1
```
rafaelfranca: Should not it be 2.3.3? Or greater? We use that in production, so it makes sense to use that here too.

.travis.yml
```diff
@@ -14,11 +14,11 @@ env:
     - SUITE=pg
     - SUITE=sqlite3
 rvm:
-  - 2.2.3
```
rafaelfranca: I'd keep both and also add 2.4 to the list, we want to make sure all ruby versions works unless we have a strong reason to remove support to 2.2.

.travis.yml
```diff
@@ -14,11 +14,13 @@ env:
...
+  - 2.2.6
+  - 2.3.3
+  - 2.4.1
```
rafaelfranca: Urgh, the app don't work with 2.4 yet, so maybe we can leave only 2.2 and 2.3 now?

dev.yml
```diff
@@ -7,7 +7,7 @@ up:
     - openssl
     - sqlite
   - ruby:
-      version: 2.2.3p172-shopify
+      version: 2.3.3
```
casperisfine: Please keep the dev version to a 2.x release.

kirs: > Shipit should stay 2.2.x compatible for the time being.

Why can't we use 2.3 in dev and keep 2.2 compatibility in Travis?

casperisfine: - People would use 2.3 feature in dev just to be surprised by the CI breaking when they submit their PR
- What advantage would that give us?

kyledurand: I couldn't dev up using `2.2.3p172-shopify` which is why I proposed this change. If it's going to cause issues I can work around that and close this PR

kirs: I think what was meant was pure 2.2.3 without shopify additions

PR #683, opened 2017-03-31
Tasks menu

kyledurand:
@kirs sorry this took me a couple of days to get to. Please 🎩  and ping whoever else you feel might want to review this before it ships

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 # 0.16.0
 
+* Move uncommon tasks to drop down menu.
+
 * Add a CCMenu compatible API.
 
 * Implement an optional merge queue, allowing to schedule pull requests to be merged as soon as the targeted branch is clear.
--- a/app/assets/images/caret-down.svg
+++ b/app/assets/images/caret-down.svg
@@ -1,0 +1,1 @@
+<svg viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg"><path d="M5 8l5 5 5-5" fill="#248AF2" fill-rule="evenodd"/></svg>
--- a/app/assets/stylesheets/_base/_base.scss
+++ b/app/assets/stylesheets/_base/_base.scss
@@ -106,10 +106,6 @@ a {
   text-decoration: none;
   cursor: pointer;
   color: $blue;
-
-  &:hover {
-    color: rgba($blue, 0.7);
-  }
 }
 
 .more {
--- a/app/assets/stylesheets/_base/_base.scss
+++ b/app/assets/stylesheets/_base/_base.scss
@@ -106,8 +106,10 @@ a {
   text-decoration: none;
   cursor: pointer;
   color: $blue;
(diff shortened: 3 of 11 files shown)


kirs:
Thanks! It looks just great. Here is the preview for others:

image

For the context, we've had people complaining about the the header growing too much and taking more than 50% on the deploy screen:

screen shot 2017-03-31 at 23 28 56

This comes when people add more and more tasks to shipit.yml. When I discussed the UX with @kyledurand he suggested that we could put tasks into a dropdown and save the room.

app/assets/images/caretDown.svg
```diff
@@ -0,0 +1 @@
+<svg viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg"><path d="M5 8l5 5 5-5" fill="#248AF2" fill-rule="evenodd"/></svg>
```
casperisfine: All other images use `lower-case-with-dashes` for names. Can you rename that file please for consistency.

casperisfine:
Fix: https://github.com/Shopify/shipit-engine/issues/411

kyledurand:
I just saw the issue and that there was a design for this and I did something similar but completely different. How do we feel about this version vs the design? I might change it to the designed version, since it was done by a designer and then get some fed eyes on this

casperisfine:
@kyledurand that was designed a very long time ago. Your solution is a huge improvement already.

If you want to go further, like only going with a dropdown if there is more than X task, you can do so in a followup.

dev.yml
```diff
@@ -7,7 +7,7 @@ up:
     - openssl
     - sqlite
   - ruby:
-      version: 2.2.3p172-shopify
+      version: 2.2.3
```
casperisfine: @jules2689 any idea why that ⏫  might be?

kyledurand: Good question although this seems to be an artifact from an ealier PR and should go away with a rebase because the none shopify version of `2.2.3` is what's currently in [master](https://github.com/Shopify/shipit-engine/blob/master/dev.yml#L10)

I'll rebase before 🚢

jules2689: 2.2.x wasnt allowed before because Shopify used a super special snowflake 2.2.x that was binary incompatible with standard 2.2.x

jules2689: Also, @kyledurand sounds like master is broken then

jules2689: https://github.com/Shopify/dev/pull/729
https://github.com/Shopify/dev-shopify/pull/96

Fixes this

kyledurand:
@casperisfine I just rebased. All good to 🚢   when green? I'll squash all commits when I do.

As Julian mentioned, these should fix dev issues with the version of ruby used here:
https://github.com/Shopify/dev/pull/729
https://github.com/Shopify/dev-shopify/pull/96

byroot:
Still looks good to me.

GoodForOneFare:
🚢

PR #690, opened 2017-04-24
Discover restart task from Kubernetes

kirs:
This adds restart task to all apps that have `kubernetes` support enabled. It works automatically and we won't need to update all repos to add the new task.

Preview:

screen shot 2017-04-24 at 18 11 10

IMO this is not the best UI because developers would need to enter the name of k8s deployment to restart. Any ideas how we can make it better?

cc @Shopify/cloudplatform @Shopify/undercloud

### DIFF
--- a/app/models/shipit/deploy_spec/kubernetes_discovery.rb
+++ b/app/models/shipit/deploy_spec/kubernetes_discovery.rb
@@ -9,6 +9,20 @@ module Shipit
         discover_kubernetes || super
       end
 
+      def discover_task_definitions
+        if kube_config.present?
+          {
+            'restart' => {
+              'action' => "Restart application",
+              'description' => "Simulates a rollout of Kubernetes deployments by using kubernetes-restart utility",
+              'steps' => [kubernetes_restart_cmd],
+            },
+          }
+        else
+          super
+        end
+      end
+
       private
 
       def discover_kubernetes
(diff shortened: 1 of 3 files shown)


app/models/shipit/deploy_spec/kubernetes_discovery.rb
```diff
@@ -20,15 +37,27 @@ def discover_kubernetes
...
+
+      def kubernetes_restart_cmd
+        cmd = ''
+        cmd << Shellwords.join([
+          "kubernetes-restart",
+          kube_config.fetch('namespace'),
+          kube_config.fetch('context'),
+          '--deployments',
+        ])
+        cmd << " $DEPLOYMENTS"
```
kirs: If we put `$DEPLOYMENTS` into `Shellwords.join` it will be escaped.

casperisfine: You can do without the temp variable:

```ruby
Shellwords.join([
  'kubernetes-restart',
  kube_config.fetch('namespace'),
  kube_config.fetch('context'),
  '--deployments',
]) + ' $DEPLOYMENTS'    
```

casperisfine:
You also have a rubocop failure.

dwradcliffe:
If we can discover the deployment options we can put them in a `<select>`

sirupsen:
You could also just have a `restart jobs` and `restart web`. I'm not sure why we'd ever need to restart one, and not the other, though?

Issue #685, opened 2017-04-06
Preventing duplicate deploys


sgringwe:
Is there anything to prevent duplicate deploys? I know I've seen warning messages in the UI if about to do so, but is there anything to actually prevent the same stack being deployed at the same time?

Sorry if I missed it! Couldn't find anything in the docs or in some basic code search.

byroot:
Yes. If you attempt to deploy while there is an active deploy on the way, you'll be taken back to the deploy page and asked to acknowledge you want to deploy concurrently (hint most likely you won't).

Line of code wise it's this: https://github.com/Shopify/shipit-engine/blob/1217e7d04b686ec0f2d7b354b83cbfd0e4ad8757/app/controllers/shipit/deploys_controller.rb#L23

Technically, a race condition is possible, but the window is small enough that it's unlikely.

PR #693, opened 2017-05-04
Fully prevent race condition while triggering deploys

casperisfine:
As described here: https://github.com/Shopify/shipit-engine/issues/685#issuecomment-292353111

Until now we had a protection against concurrent deploys, but it wasn't perfect since a race condition window, albeit relatively small, exists.

This PR move the concurrency protection inside the deploy creation transaction to prevent any such race condition.

@Shopify/pipeline for review please

### DIFF
--- a/app/controllers/shipit/deploys_controller.rb
+++ b/app/controllers/shipit/deploys_controller.rb
@@ -20,10 +20,15 @@ module Shipit
     end
 
     def create
-      return redirect_to new_stack_deploy_path(@stack, sha: @until_commit.sha) if !params[:force] && @stack.active_task?
-
-      @deploy = @stack.trigger_deploy(@until_commit, current_user, env: deploy_params[:env])
+      @deploy = @stack.trigger_deploy(
+        @until_commit,
+        current_user,
+        env: deploy_params[:env],
+        force: params[:force].present?,
+      )
       respond_with(@deploy.stack, @deploy)
+    rescue Task::ConcurrentTaskRunning
+      redirect_to new_stack_deploy_path(@stack, sha: @until_commit.sha)
     end
 
     def rollback
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -81,7 +81,7 @@ module Shipit
       undeployed_commits_count > 0
     end
 
-    def trigger_task(definition_id, user, env: nil)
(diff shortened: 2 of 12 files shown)


app/controllers/shipit/deploys_controller.rb
```diff
@@ -20,10 +20,15 @@ def show
...
-      @deploy = @stack.trigger_deploy(@until_commit, current_user, env: deploy_params[:env])
+      @deploy = @stack.trigger_deploy(
+        @until_commit,
+        current_user,
+        env: deploy_params[:env],
+        force: params[:force].present?,
+      )
       respond_with(@deploy.stack, @deploy)
+    rescue Task::ConcurrentTaskRunning
+      redirect_to new_stack_deploy_path(@stack, sha: @until_commit.sha)
```
DazWorrall: Worth adding a flash message or similar here and for rollbacks?

app/models/shipit/stack.rb
```diff
@@ -96,18 +96,20 @@ def trigger_task(definition_id, user, env: nil)
         until_commit_id: commit.id,
         since_commit_id: commit.id,
         env: definition.filter_envs(env),
+        allow_concurrency: definition.allow_concurrency? || force,
```
DazWorrall: I would have expected `force` to take precedence here.

app/controllers/shipit/deploys_controller.rb
casperisfine: It's already handled in the template. You got an orange warning asking you to confirm if you want to allow a concurrent deploy.

app/models/shipit/stack.rb
casperisfine: It doesn't really matter. Either the task already allow concurrency, and then `force` is pointless, or either it doesn't and then only `force` being `true` will change a thing.

casperisfine:
> So rather than not starting, this will immediately fail a concurrent deploy if it detects it's not the only one running. Do I understand this correctly?

It's the same behavior than before, just with the race condition removed. Meaning:

  - Only one deploy can be active at a time on a stack.
  - `active` is defined as being either scheduled or running
  - If you try to schedule a deploy while one is already active, you get redirected back with an error showing you the currently active deploy, and an option to ignore the concurency issue and deploy anyway.

Ref: https://github.com/Shopify/shipit-engine/pull/343

PR #695, opened 2017-05-08
Specify rails version in migration

sonalkr132:
Error during running migration:
StandardError: An error has occurred, this and all later migrations canceled:

Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:

  class CreateBaseline < ActiveRecord::Migration[4.2]


casperisfine:
Thanks

PR #696, opened 2017-05-13
Update Rails version recommendation on README

dlackty:
As the latest gem version now requires Rails `~> 5.1.0`, we should update README as well.


byroot:
Thanks!

PR #701, opened 2017-06-05
Update stack inaccessible message

DazWorrall:
A couple of times recently people have been confused by this message because the target repo does exist, in both cases the target branch had actually been deleted.

/r: @Shopify/pipeline @stefanmb

### DIFF
--- a/app/views/shipit/stacks/show.html.erb
+++ b/app/views/shipit/stacks/show.html.erb
@@ -23,7 +23,7 @@
     <div class="banner banner--red">
       <div class="banner__inner wrapper">
         <div class="banner__content">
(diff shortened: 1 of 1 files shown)


app/views/shipit/stacks/show.html.erb
```diff
@@ -23,7 +23,7 @@
     <div class="banner banner--red">
       <div class="banner__inner wrapper">
         <div class="banner__content">
-          <h2 class="banner__title">This repository seems to be missing!</h2>
+          <h2 class="banner__title">This repository or branch seems to be missing!</h2>
```
casperisfine: We could even go as far as mentioning the branch: `@stack.branch`.

But that will do.

Issue #699, opened 2017-05-18
Record deploys performed in emergency mode


casperisfine:
Feature request by @pallan 

By design Shipit tries to prevent honest mistakes, but leaves as much leeway as possible to users, so they can deal with emergencies. That is why, failing statuses prevent deploying by mistakes, but ultimately it's possible to turn on the emergency mode and deploy a failing build anyway.

However it doesn't mean there is no accountability, as such it's important to have proper traceability.

In that spirit, it would be good to record on the deploys if they were triggered with emergency mode.

As I see it:

  - We add a new `emergency boolean not null default false` column on `tasks`.
  - When the deploy / rollback record is created, we check if it would have been allowed without emergency mode, if not we set `emergency = true`.
  - We add an icon or something in the deploy list for deploys that bypassed the ci statuses.

cc @pallan @wvanbergen @DazWorrall

perobertson:
It would also be nice to have the ability for shipit to notify a slack channel on emergency deploys. That way anyone who wants to say informed can watch that channel.

casperisfine:
> shipit to notify a slack channel 

On that specific note, it was decided not to add any kind of messaging notification to Shipit itself a while ago.

The thing is, there is many different IM services, some people would like mail notifications as well maybe, etc, etc. It kinda feels like pandora's box from a project perspective.

That is why Shipit supports webhooks, and people wanting such feature are expected to use the webhooks infra for it.

And that's actually what we do at Shopify:
![capture d ecran 2017-05-18 a 17 37 09](https://cloud.githubusercontent.com/assets/19192189/26210668/a9910cda-3bf0-11e7-993b-1d6896c30532.png)

pallan:
Another potential piece of value would be a comment(s) the user can add pre- or post-deploy to indicate why they had to perform the emergency

orenmazor:
I really like the bit about having to explain the reason for the emergency :+1:  I'm sure Ulf the time it'll be "Fuuuuuuuu" or "fuck _______ remote tests" but still worth having. 

Re: notifications. I'd be pretty happy even if there was a FORCED_DEPLOY=1 env variable available to scripts.

byroot:
> I'd be pretty happy even if there was a FORCED_DEPLOY=1 env variable available to scripts.

Very good idea.

PR #708, opened 2017-06-29
Mark deploys if they ignored safeties

casperisfine:
Fix: https://github.com/Shopify/shipit-engine/issues/699

Ignoring safeties can mean two things:

  - Deploying while commit statuses wouldn't allow it using the emergency mode
  - Deploying or running a task concurrently

As usual, some awesome frontend work (:trollface:).

capture d ecran 2017-06-29 a 17 29 06

I guess an icon instead would have been nice, but it's out of my reach.

As for recording the reason, I just can't figure out a proper UX / UI for it, so I prefer not to do some subpar flow.

cc @pallan @orenmazor @perobertson

### DIFF
--- a/app/models/shipit/deploy.rb
+++ b/app/models/shipit/deploy.rb
@@ -46,6 +46,7 @@ module Shipit
         until_commit: until_commit,
         env: env.try!(:to_h) || {},
         allow_concurrency: force,
+        ignored_safeties: force,
       )
     end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+* Marks deploys that are triggered while ignoring safety features.
+
 # 0.19.0
 
 * Fix MRI 2.4 support (https://github.com/attr-encrypted/attr_encrypted/issues/258)
--- a/app/assets/stylesheets/_pages/_deploy.scss
+++ b/app/assets/stylesheets/_pages/_deploy.scss
@@ -212,3 +212,7 @@
     border: none;
   }
 }
+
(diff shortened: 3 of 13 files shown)


app/models/shipit/deploy.rb
```diff
@@ -46,6 +46,7 @@ def build_rollback(user = nil, env: nil, force: false)
         until_commit: until_commit,
         env: env.try!(:to_h) || {},
         allow_concurrency: force,
+        ignored_safeties: force,
```
orenmazor: bender

orenmazor:
@wvanbergen ya totally. I'd love to force people to have to write at least 300 characters on why they're forcing that deploy. But, baby steps.

jeremyhansonfinger:
Hi @casperisfine, I'm documenting this for processing integrity. Just to confirm: a deploy shouldn't be labelled as `ignored safeties` due to `!until_commit.deployable?` (`!locked? && !active_task?`) unless somehow it's forced through without using the UI?

jeremyhansonfinger:
@byroot, I know you can put the stack into emergency mode, but that's covered by `force`. Is it possible to force a deploy through the UI without putting it into emergency mode?

byroot:
When you attempt to deploy and there is already another deploy going through, you'll get a warning and a checkbox allowing you to ignore said warning.

So yeah, ignored safeties is:
  - Either you sued emergency mode to ignore CI statuses
  - Either you ignored the concurrent deploy safeties.

Both either via the UI or the API.

jeremyhansonfinger:
Okay, thank you!

PR #713, opened 2017-08-04
Make sure kubernetes discovery does not override existing task

kirs:
Related to https://github.com/Shopify/shipit/issues/214

When I added the kubernetes-restart task, I made sure that it wouldn't override existing `restart` task. After looking into https://github.com/Shopify/shipit/issues/214, I double-checked that and added a test case.

### DIFF
--- a/test/models/deploy_spec_test.rb
+++ b/test/models/deploy_spec_test.rb
@@ -346,6 +346,21 @@ module Shipit
       assert_equal ['bundle exec foo'], definition.steps
     end
 
+    test "#task_definitions does not add kubernetes-restart if restart task is already defined" do
+      @spec.stubs(:load_config).returns(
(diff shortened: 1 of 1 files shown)


DazWorrall:
Wait, if this new test is passing, how are we seeing the task overridden in Shopify/shipit#214?

kirs:
Not sure. Maybe it's caused by definition cache?

casperisfine:
I got bitten by that unit test a couple times. It's a bit hard to explain, but basically it isn't reflecting the reality well enough because it's bypassing the file system layer.

So I'm pretty sure that unit test is lying.

The way we test `DeploySpec`should be improved to be stubbing free.

bbuchalter:
The contents of this PR seems to have made it into master: https://github.com/Shopify/shipit-engine/blob/master/test/models/deploy_spec_test.rb#L353 and should probably be closed.

PR #714, opened 2017-08-07
Ability to explicitly disable `bundle exec` for tasks

DazWorrall:
Ref Shopify/shipit#214, this allows you to explicitly disable `bundle exec` for a task, and disables it for the default `kubernetes-restart` tasks.

### DIFF
--- a/app/models/shipit/deploy_spec/bundler_discovery.rb
+++ b/app/models/shipit/deploy_spec/bundler_discovery.rb
@@ -54,6 +54,7 @@ module Shipit
       end
 
       def coerce_task_definition(config)
+        return config if config.fetch('bundler', nil) == false
         config.merge('steps' => Array(config['steps']).map(&method(:bundle_exec)))
       end
     end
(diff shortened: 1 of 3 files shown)


casperisfine:
I think we should just not add the bundle exec automatically on tasks. it only make sense for commands that are inferred 

 It might break a few projects though, but I'd prefer that over adding a weird option to tasks like this.

DazWorrall:
Are you happy to just remove that or should we deprecate it first?

casperisfine:
> I think we need bundle exec at least for Capistrano.

Indeed, but the capistrano discovery could handle it itself. it would also allow it to favor `bin/cap` over `bundle exec cap` if the binstub is present.

casperisfine:
> should we deprecate it first?

I checked shopify's repo, I can only see a handful of tasks that would be broken, so we can easily fix them all in a few minutes.

As for other users of Shipit, well, it's indeed not super cool, so I guess a deprecation mechanism wouldn't be a bad idea. Some gobal option we'd default to the current behavior for now:

```ruby
Shipit.automatically_prepend_bundle_exec = true / false
```

DazWorrall:
Superseding this with #715

PR #715, opened 2017-08-07
Deprecate automatically prepending `bundle exec` to tasks

DazWorrall:
Supersedes #714, this deprecates automatically prepending `bundle exec` to tasks, though we maintain the default behaviour for Capistrano, and prepend `bundle exec` if we detect bundler. We wont duplicate `bundle exec` if it's already present, allowing projects to add it themselves before setting `SHIPIT_PREPEND_BUNDLE_EXEC=0`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 * Fix caching of discovered tasks.
 * Reconsider all undeployed commits for locking after a rollback has completed (https://github.com/Shopify/shipit-engine/issues/707).
 * Marks deploys that are triggered while ignoring safety features (https://github.com/Shopify/shipit-engine/issues/699).
-* Automatically prepending `bundle exec` to tasks is deprecated, set `SHIPIT_PREPEND_BUNDLE_EXEC=0` to test the future behaviour before it is enforced.
+* Automatically prepending `bundle exec` to tasks is deprecated, set `Shipit.automatically_prepend_bundle_exec = false` to test the future behaviour before it is enforced.
 
 # 0.19.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Fix caching of discovered tasks.
 * Reconsider all undeployed commits for locking after a rollback has completed (https://github.com/Shopify/shipit-engine/issues/707).
 * Marks deploys that are triggered while ignoring safety features (https://github.com/Shopify/shipit-engine/issues/699).
+* Automatically prepending `bundle exec` to tasks is deprecated, set `SHIPIT_PREPEND_BUNDLE_EXEC=0` to test the future behaviour before it is enforced.
 
 # 0.19.0
 
--- a/lib/shipit.rb
+++ b/lib/shipit.rb
@@ -57,6 +57,7 @@ module Shipit
   delegate :table_name_prefix, to: :secrets
 
   attr_accessor :disable_api_authentication
+  attr_writer :automatically_prepend_bundle_exec
 
   def app_name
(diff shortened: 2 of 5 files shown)


lib/shipit.rb
```diff
@@ -206,6 +206,11 @@ def default_inactivity_timeout
     secrets.commands_inactivity_timeout || 5.minutes.to_i
   end
 
+  def automatically_prepend_bundle_exec
+    # Prepending `bundle exec` automatically is deprecated
+    Rails.env.production? && ENV.fetch('SHIPIT_PREPEND_BUNDLE_EXEC', '1') != '0'
```
casperisfine: Why the production check?

casperisfine: Also we should probably enable the new behavior on new apps via the template.rb: https://github.com/Shopify/shipit-engine/blob/96e98e71742a85d47bf529e21b63b9e682147ef3/template.rb

And finally we should probably log an `ActiveSupport::Deprecation.warn` on boot if that env var isn't set.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 * Fix caching of discovered tasks.
 * Reconsider all undeployed commits for locking after a rollback has completed (https://github.com/Shopify/shipit-engine/issues/707).
 * Marks deploys that are triggered while ignoring safety features (https://github.com/Shopify/shipit-engine/issues/699).
+* Automatically prepending `bundle exec` to tasks is deprecated, set `SHIPIT_PREPEND_BUNDLE_EXEC=0` to test the future behaviour before it is enforced.
```
kirs: Is it a common practice to use ENV variables to configure Shipit?

lib/shipit.rb
kirs: Why only in production? Shouldn't it be consistent across environments?

casperisfine: Also why use an environment variable? A simple class method would do just fine I think.

DazWorrall: Yeah you're right, I forget this is a template.

template.rb
```diff
@@ -126,6 +126,8 @@
 end
 CODE
 
+environment 'Shipit.automatically_prepend_bundle_exec = false'
```
DazWorrall: I have no idea if this is right :sciencedog:

casperisfine: Seems about right.

lib/shipit.rb
```diff
@@ -206,6 +207,16 @@ def default_inactivity_timeout
     secrets.commands_inactivity_timeout || 5.minutes.to_i
   end
 
+  def automatically_prepend_bundle_exec
+    @automatically_prepend_bundle_exec ||= begin
```
casperisfine: That won't work. If you set the variable to false, the `||=` will set it back to `true` on the next invocation.

You need to use `unless defined?(@automatically_prepend_bundle_exec)`

test/models/deploy_spec_test.rb
```diff
@@ -318,14 +318,32 @@ class DeploySpecTest < ActiveSupport::TestCase
       assert_equal 'SAFETY_DISABLED', variable_definition.name
     end
 
-    test "task definitions prepend bundle exec if necessary" do
+    test "task definitions prepend bundle exec if enabled" do
+      Shipit.expects(:automatically_prepend_bundle_exec).returns(true).at_least_once
```
kirs: Should we also test the default behaviour when `automatically_prepend_bundle_exec` is not set?

DazWorrall: Good idea, done. I assert the deprecation warning too.

kirs:
:heart: for asserting the deprecation warning

PR #721, opened 2017-08-29
Always schedule a MergePullRequestsJob for merge-queue-enabled stacks

wvanbergen:
Schedule the job as part of the minutely cron, no matter if they have pending PRs or not. This is slightly less efficient, but will make it a lot easier to instrument.

### DIFF
--- a/app/models/shipit/pull_request.rb
+++ b/app/models/shipit/pull_request.rb
@@ -102,7 +102,7 @@ module Shipit
     end
 
     def self.schedule_merges
-      Shipit::Stack.where(id: pending.uniq.pluck(:stack_id)).find_each(&:schedule_merges)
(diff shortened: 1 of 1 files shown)


app/models/shipit/pull_request.rb
```diff
@@ -102,7 +102,7 @@ def required_statuses
     end
 
     def self.schedule_merges
-      Shipit::Stack.where(id: pending.uniq.pluck(:stack_id)).find_each(&:schedule_merges)
+      Shipit::Stack.where(merge_queue_enabled: true).find_each(&:schedule_merges)
```
casperisfine: Note: That query won't be indexed, but the table isn't very likely to be big enough to matter.

PR #723, opened 2017-08-31
Add branch name to settings screen.

garethson:
Makes it clear what branch you're deploying from, as this is only available in the stack list right now:

image


byroot:
Test failures are unrelated, I'll look at them soon.

PR #725, opened 2017-09-04
Workaround Travis MySQL issues

casperisfine:
Fixes the failure we had on MySQL builds.

Someone already reported the issue to Travis https://github.com/travis-ci/travis-ci/issues/8331

I subscribed to the issue, so if Travis gives a proper fix or a better workaround I'll update the travis.yml accordingly.


casperisfine:
They are, I should fix those as well.

PR #729, opened 2017-09-08
Add `Commit.deploy_requested_at`

DazWorrall:
This is attempt to identify when a commit entered the deploy pipeline:

* If the commit is a direct merge to master, use the local `created_at` timestamp - this is to more reliably identify when a commit was pushed to Github. `committed_at` is vulnerable to skew because commits could exist local to a developer long before they are actually pushed for  deployment
* If the commit was merged via the merge queue, use `merge_requested_at` - although not the time the commit was merged, this was the moment this commit entered the deploy pipeline
* Handle the case where the commit was added and then removed from the merge queue, and merged manually

### DIFF
--- a/app/models/shipit/commit.rb
+++ b/app/models/shipit/commit.rb
@@ -194,6 +194,14 @@ module Shipit
       self.pull_request_title = message_parser.pull_request_title unless self[:pull_request_title]
     end
 
+    def deploy_requested_at
+      if pull_request.try!(:merged?)
+        pull_request.merge_requested_at
+      else
+        created_at
+      end
+    end
+
     private
 
     def message_parser
--- a/app/jobs/shipit/merge_pull_requests_job.rb
+++ b/app/jobs/shipit/merge_pull_requests_job.rb
@@ -8,9 +8,7 @@ module Shipit
       pull_requests.each do |pull_request|
         pull_request.refresh!
         pull_request.reject_unless_mergeable!
-        if pull_request.closed?
-          pull_request.merged_upstream? ? pull_request.complete! : pull_request.cancel!
-        end
+        pull_request.cancel! if pull_request.closed?
       end
 
       return false unless stack.allows_merges?
--- a/test/fixtures/shipit/commits.yml
+++ b/test/fixtures/shipit/commits.yml
@@ -102,3 +102,17 @@ shipit_pending_pr_head:
   deletions: 24
   updated_at: <%= 8.days.ago.to_s(:db) %>
   detached: true
+
+cyclimse_merged:
+  id: 9
+  sha: 772b4795871e4a45aa17d53f02aa45d0
+  message: "Merge pull request #66 from cyclimse/yoloshipit\n\nyoloshipit!"
+  stack: cyclimse
+  author: walrus
+  committer: walrus
+  authored_at: <%= 4.days.ago.to_s(:db) %>
+  committed_at: <%= 3.days.ago.to_s(:db) %>
+  additions: 420
+  deletions: 342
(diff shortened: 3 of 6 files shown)


app/models/shipit/commit.rb
```diff
@@ -194,6 +194,13 @@ def identify_pull_request
       self.pull_request_title = message_parser.pull_request_title unless self[:pull_request_title]
     end
 
+    def deploy_requested_at
+      if pull_request
+        return pull_request.merge_requested_at unless pull_request.canceled?
```
wvanbergen: @byroot Is the presence of `pull_request` on a commit a certain sign that the merge queue was used?

DazWorrall: My reading of the code is we only create a pull request locally if added to the merge queue, for example `merge_requested_at `is [`NOT NULL`](https://github.com/Shopify/shipit-engine/blob/ce81004ea2f058815519d3fd581f83e363524ab1/test/dummy/db/schema.rb#L143), so always set.

casperisfine: > Is the presence of pull_request on a commit a certain sign that the merge queue was used?

It is, however it doesn't necessarily mean it's the merge queue that merged it.

I think the condition should be:

```
if pull_request.try!(:merged?)
  pull_request.merge_requested_at
else
  created_at
end
```

I don't see a case where a `pull_request`  could have a `merge_requested_at == nil`.

As for `merged?` I need to double check if it's 100% guaranteed it was merged by the merge queue.

casperisfine: I double checked, I think we're good.

DazWorrall: > As for merged? I need to double check if it's 100% guaranteed it was merged by the merge queue.

[Not currently](https://github.com/Shopify/shipit-engine/blob/ce81004ea2f058815519d3fd581f83e363524ab1/app/jobs/shipit/merge_pull_requests_job.rb#L12). I wonder if we could always `cancel!` there...

DazWorrall: It's another corner case tbf - added to the merge queue, then merged manually without first being removed.

casperisfine: Damn, nice catch. I knew we were handling this somewhere, but couldn't find it.



> I wonder if we could always cancel! there...

Hum. Yeah, either cancel, or a new dedicated status. `cancel` would make sense as if you merged manually, you canceled your merge request IMO.

DazWorrall: It highlights a slight ambiguity in the model name (:bikeshed:), the _merge request_ may have been cancelled (or invalidated?), but the pull request _was_ completed. Let's go with cancel for now, unless we find a need for another state.

casperisfine: Indeed. `PullRequest` represent both the GitHub `PullRequest`, and a merge request from the user. 

I definitely though about splitting the two during the original implementation, but though it complicated things. I kinda regret now.

DazWorrall: > I definitely though about splitting the two during the original implementation, but though it complicated things. I kinda regret now.

I think it could be decoupled with some tweaks to the state machine and schema - the model could represent the the pull request, and the state machine the merge queue. I'll look at that separately.

Happy for this to be merged?

casperisfine: > Happy for this to be merged?

Well, I still think checking for `merged?` would be better than `!canceled`.

casperisfine: Also right now, you don't fallback to `created_at` if the merge request was canceled.

DazWorrall: I made that change but didnt push it :doh: done now :)

PR #732, opened 2017-09-21
Move dev/railgun to myshopify.io domain

DazWorrall:
My Chrome gives connection refused for `shipit-engine.localhost` - I wonder if it considers itself authoritative for `localhost` and bypasses resolvers/hosts file. This works for me - and switches to SSL as a bonus - but requires Shopify developers to update their OAuth apps with a new callback URL.


DazWorrall:
> I am wondering how this will work out for non-shopify users.

It should be completely transparent, these files are only used by `dev [up]`.

PR #734, opened 2017-09-25
Add `force` to stack cache key

DazWorrall:
On our installation we've encountered situations when the 'Redeploy' button is disabled when it shouldn't be, fragment cache is the likely suspect. This is a best guess fix, I unfortunately have not been able to replicate this locally - in development mode, I'm not entirely sure I ever got the fragment cache actually enabled+working - I added this to `test/dummy/config/environments/development.rb`

```
  config.cache_store = :memory_store, { size: 64.megabytes }
  config.log_level = :debug
  config.action_controller.perform_caching = true
```

.. an ran `bin/rails app:dev:cache` to enable the cache. I saw some blogs suggesting I should see cache activity in the logs, but I never saw any :(

I also tried to write a unit test for this but struggled with the same issues :(

### DIFF
--- a/app/views/shipit/stacks/show.html.erb
+++ b/app/views/shipit/stacks/show.html.erb
@@ -21,7 +21,7 @@
     </ul>
   </section>
 
-  <% cache @stack do %>
+  <% cache [@stack, params[:force]] do %>
     <section>
       <header class="section-header">
         <h2>Previous Deploys</h2>


DazWorrall:
@rafaelfranca do you know anything which might help us out here? This caused another issue today so we need to either fix it or make the call to remove the caching.

app/views/shipit/stacks/show.html.erb
```diff
@@ -21,7 +21,7 @@
     </ul>
   </section>
 
-  <% cache @stack do %>
+  <% cache [@stack, params[:force]] do %>
```
esnunes: this PR is quite old but I guess it is still relevant. Correct me if I'm wrong @JackTLi but you were facing a similar cache issue, right? 

My suggestion in this case is to combine `stack` and `tasks`, so whenever the stack `updated_at` changes or any of tasks (collection) it is going to expire the cache, right @rafaelfranca ?

casperisfine: Shipit propagate the `updated_at` changes. If a `task.updated_at` change, it's `task.stack.updated` will be bumped as well.

esnunes: Thanks Jean, should we close this PR?

casperisfine: Well, the `force` params still make sense though.

DazWorrall: I feel like we tried this in production and it didnt work, but it's been so long I'm not sure 🤔 

Something a bit crazier we could try: disabling the cache entirely for a day and seeing if the perf is ok. Since continuous deployment and the cloud move I wonder if old assumptions about perf + usage patterns still apply.

JackTLi: I think we can def remove this particular cache since we still have caches for the individual resources

casperisfine: Keep in mind that `stack#show` is automatically pulled by everyone that has a tab open on it whenever something change on the stack.

Which mean you can easily get a big spike of hits on it. So as always measure is everyting, but I do believe that higher level cache do see a very decent hit rate, and is much master than the collection of lower caches.

rdji123:
This issue is [re-occurring](https://shopify.slack.com/archives/C016Z7T0X0X/p1636656121141700). @DazWorrall were you able to find a solution or workaround back in 2019?

DazWorrall:
@rdji123 I barely remember at this point 😅 From what I can tell we never managed to successfully replicate the issue in development, so never had confidence on a fix. We could rebase and ship this branch to test on our prod instance, if we are able to reliably replicate there.

stephsachrajda:
I've gone ahead and rebased this. It seems reproducible on https://shipit.shopify.io/shopify/canary-demo-app/production?force=1&noredirect=1 by locking and then going into emergency mode so we should be able to check

DazWorrall:
I would suggest shipping just this branch to prod to validate the fix there - if it works, this can be merged and we can move our instance back to master. Ideally we figure out a test for this, but at this point it would be nice to just get it fixed.

stephsachrajda:
> I would suggest shipping just this branch to prod to validate the fix there - if it works, this can be merged and we can move our instance back to master. Ideally we figure out a test for this, but at this point it would be nice to just get it fixed.

Yeah that sounds like a smart plan. We'll wait until after the BFCM infra freeze, then give this a shot!

PR #738, opened 2017-10-06
npm-publish versions with pre-release tags under a 'next' dist-tag

### LINKED ISSUE #735: How do I `yarn tag add react-utilities@2.0.0-beta.1 beta` using Shipit?
My use case is that I'd like to publish a beta to npm and I don't want it to be installed by default when users run `yarn add react-utilities`.

The way I'd publish a version behind a beta flag is:
- version & publish normally
- `yarn tag add react-utilities@2.0.0-beta.1 beta`

I couldn't find how to do this with Shipit - am I missing something?

kaelig:
Fixes #735

### DIFF
--- a/app/models/shipit/deploy_spec/npm_discovery.rb
+++ b/app/models/shipit/deploy_spec/npm_discovery.rb
@@ -40,6 +40,15 @@ module Shipit
         JSON.parse(file.read)['private'].blank?
       end
 
+      def dist_tag(version)
+        # Pre-release SemVer tags such as 'beta', 'alpha', 'rc' and 'next'
+        # are treated as 'next' npm dist-tags.
+        # An 1.0.0-beta.1 would be installable using both:
+        # `yarn add package@1.0.0-beta.1` and `yarn add package@next`
+        return 'next' if ['-beta', '-alpha', '-rc', '-next'].any? { |tag| version.include? tag }
+        'latest'
+      end
+
       def package_json
         file('package.json')
       end
@@ -60,10 +69,14 @@ module Shipit
         discover_npm_package || super
       end
 
+      def package_version
+        JSON.parse(package_json.read)['version']
+      end
+
       def publish_npm_package
         check_tags = 'assert-npm-version-tag'
         # `yarn publish` requires user input, so always use npm.
-        publish = 'npm publish'
+        publish = "npm publish --tag #{dist_tag(package_version)}"
 
         [check_tags, publish]
       end
--- a/test/models/deploy_spec_test.rb
+++ b/test/models/deploy_spec_test.rb
@@ -578,14 +578,6 @@ module Shipit
       refute @spec.npm?
     end
 
-    test '#npm? is false if npm package is private' do
-      file = Pathname.new('/tmp/fake_package.json')
-      file.write('{"private": true}')
-
-      @spec.expects(:package_json).returns(file)
-      refute @spec.npm?
-    end
-
     test '#npm? is true if npm package is public' do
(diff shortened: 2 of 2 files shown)


app/models/shipit/deploy_spec/npm_discovery.rb
```diff
@@ -40,6 +40,13 @@ def public?
         JSON.parse(file.read)['private'].blank?
       end
 
+      def dist_tag(version)
+        # All pre-release tags are treated as "next"
+        # e.g. 1.0.0-beta.1 is installable using `yarn add package@next``
+        return 'next' if version.include? '-'
```
ismail-syed: Can we check for an explicit `-beta` instead? 

We do some funny things in some places so this might get flagged. 
eg. https://github.com/Shopify/tinymce/blob/master/package.json#L3

test/models/deploy_spec_test.rb
```diff
@@ -600,14 +592,45 @@ class DeploySpecTest < ActiveSupport::TestCase
...
+
+    test '#dist_tag returns "latest" if the version does not contains a pre-release tag' do
+      assert_equal 'latest', @spec.dist_tag('1.0.0')
+    end
+
+    test '#dist_tag returns "next" if the version contains a pre-release tag' do
+      assert_equal 'next', @spec.dist_tag('1.0.0-rc.1')
+    end
+
+    test '#npm? is false if npm package is private' do
```
GoodForOneFare: Please move this test back up to line 581.  Easier to read when private package tests are kept together.

app/models/shipit/deploy_spec/npm_discovery.rb
GoodForOneFare: Hadn't thought of that 🤔 Great point.

kaelig: Hmm that depends. What’s the use case for -shopifyv4?

ismail-syed: There are some versioning concerns with TinyMCE so we had to do something custom-ish. I wouldn't be surprised if we had a couple other cases org-wide somewhere. 
cc/ @nickjs

kaelig: Thanks! Then to be safe let’s detect alpha, beta, rc, and next. Sounds good?

nickjs: In https://github.com/shopify/tinymce, we keep the package version pegged to the version of tinymce we're using, but then we provide additional meta information on top of tinymce, such as config and typings. These need their own versioning. So we publish it like so: `4.5.5-shopify1`, `4.5.5-shopify2`, etc., bumping the `shopify` version as we update the config and typings. This was the best way we could find to do it, but totally open to better options.

test/models/deploy_spec_test.rb
kaelig: Good point, thanks. I regrouped them back together. The diff looks a bit weird but if you look at the whole file I think it's fine.

kaelig:
@casperisfine Thanks! Where can I document these "magic" steps?

casperisfine:
The commands Shipit infers aren't really documented at the moment.

kaelig:
Cheers @casperisfine, at some point I'd really like to get it documented because as a new joiner it was not trivial to learn about that stuff.

Do I need to do anything for this to go to production?

casperisfine:
That's internal shopify stuff. Let's talk offline ;)

PR #741, opened 2017-10-13
Fix push-to-heroku listing members on error

etiennebarrie:
The command has been renamed to access:

https://github.com/heroku/legacy-cli/blob/b0d7f8ce53574937e8658f175294c8b81f2bd093/CHANGELOG#L154


casperisfine:
Any way to detect which one should be used ?

I'm afraid people might have different versions of the toolbelt on their Shipit installation.

etiennebarrie:
I don't think it's worth it, it's been more than a year and a half that's it's been changed.

PR #750, opened 2017-11-20
Optionally reject stale pull requests from the merge queue

### LINKED ISSUE #720: Allow specifying a maximum age for branches to be allowed to be added to the merge queue
This would prevent master failures when old, outdated branches get merged, and push people to rebase instead. This should be based on the merge base of the PR.

@Shopify/pipeline thoughts?

DazWorrall:
Fixes #720 

This will reject a PR from the merge queue if the head commit is either older than the base commit for a given timespan, or if the branch falls behind the base more than a given number of commits. WIP just after some feedback:

* Have I done the migration + relation correct for `PullRequest.base_commit`?
* Is supporting both a number of commits and time frame a good idea?
* Is `requires_rebase` a good status name? `excess_divergence` seems a bit much :trollface: 
* I'm sure there's a better name for `merge.require_rebase_commits` - after your branch falls behind by this # of commits, it's considered stale.

### DIFF
--- a/app/models/shipit/deploy_spec.rb
+++ b/app/models/shipit/deploy_spec.rb
@@ -167,12 +167,12 @@ module Shipit
       end
     end
 
-    def require_rebase_commits
-      config('merge', 'require_rebase_commits')
+    def max_divergence_commits
+      config('merge', 'max_divergence', 'commits')
     end
 
-    def require_rebase_after
-      if timeout = config('merge', 'require_rebase_after')
+    def max_divergence_age
+      if timeout = config('merge', 'max_divergence', 'age')
         begin
           Duration.parse(timeout)
         rescue Duration::ParseError
--- a/app/models/shipit/deploy_spec.rb
+++ b/app/models/shipit/deploy_spec.rb
@@ -167,6 +167,19 @@ module Shipit
       end
     end
 
+    def require_rebase_commits
+      config('merge', 'require_rebase_commits')
+    end
+
+    def require_rebase_after
+      if timeout = config('merge', 'require_rebase_after')
+        begin
+          Duration.parse(timeout)
+        rescue Duration::ParseError
+        end
+      end
+    end
+
     def review_checks
       config('review', 'checks') || []
     end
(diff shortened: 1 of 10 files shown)


app/models/shipit/deploy_spec.rb
```diff
@@ -167,6 +167,19 @@ def revalidate_pull_requests_after
       end
     end
 
+    def require_rebase_commits
+      config('merge', 'require_rebase_commits')
```
byroot: Maybe something like
```yml
merge:
  diversion:
    max_commit: 50
    max_age: 5d
```

db/migrate/20171120161420_add_base_info_to_pull_request.rb
```diff
@@ -0,0 +1,7 @@
+class AddBaseInfoToPullRequest < ActiveRecord::Migration[5.1]
+  def change
+    add_column :pull_requests, :base_ref, :string
```
byroot: Since shas have a specific length, might as well limit it to 40 here.

DazWorrall:
Should we backfill the `base_commit` data? Pending PRs will pull it in when they’re updated.

byroot:
> Should we backfill the base_commit data?

I wouldn't bother too much about it. That would cause lots of API calls for little benefits. Just make sure the code handle nil values properly.

Also while I think of it, new `shipit.yml` keys would need an entry in the README.

db/migrate/20171120161420_add_base_info_to_pull_request.rb
DazWorrall: The base ref is not a sha, it's (always?) the merge base branch name. I don't think git enforces a length limit on branches.

casperisfine: My bad. In this case, we should wonder if 255 chars is enough. If we bump that limit a tad higher, say 1024, I think it won't be worth handling overflows.

DazWorrall: Sanity: Do I just need to add `limit: 1024`?

casperisfine: Yes

PR #753, opened 2017-12-11
Publish lerna monorepos (fixed version only)

GoodForOneFare:
### What?
Shipit can now publish multiple JavaScript packages from a single [lerna monorepo](https://github.com/lerna/lerna).

This helps to:
* Colocate utility libraries, which'll make them easier to publish consistently
* ^ should also reduce dev friction in `shopify/web`/`shopify/polaris-react`
* Make `shopify/sewing-kit` modular

### How?
Really similar logic to `NpmDiscovery`.  This version just checks for a `lerna.json` and uses lerna to publish instead of npm.

### What's the catch?
* [Independently versioned packages](https://github.com/lerna/lerna#independent-mode---independent) aren't supported
  * I have this mostly written, but want feedback on this before I go further
* The publish flow is a little clunky; may have to submit lerna PRs to reduce friction

### DIFF
--- a/app/models/shipit/deploy_spec/file_system.rb
+++ b/app/models/shipit/deploy_spec/file_system.rb
@@ -2,6 +2,7 @@ module Shipit
   class DeploySpec
     class FileSystem < DeploySpec
       include NpmDiscovery
+      include LernaDiscovery
       include PypiDiscovery
       include RubygemsDiscovery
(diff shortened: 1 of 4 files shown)


GoodForOneFare:
cc @lemonmade - this is finally happening

PR #754, opened 2017-12-12
Support lerna independent package versions

GoodForOneFare:
Support for [independently versioned lerna monorepos](https://github.com/lerna/lerna#independent-mode---independent).  We depend on at least one of these ([graphql-tools](https://github.com/lemonmade/graphql-tools/)).

Publishing packages in the correct order necessitates heavy use of JavaScript.  This makes publication untestable :/  I'm eager for suggestions to improve on this!

### DIFF
--- a/lib/snippets/assert-lerna-independent-version-tags
+++ b/lib/snippets/assert-lerna-independent-version-tags
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+
+TAGS=$(git tag --points-at HEAD)
+
+if [ -z "$TAGS" ]; then
+  HEAD_REV=$(git rev-parse HEAD)
+  echo -e "\033[1;31mNo tags associated with $HEAD_REV.  Did you push your commits and tags?\033[0m"
+  exit 1
+fi
+
+for TAG in $TAGS
+do
+  PACKAGE_NAME=$(echo "$TAG" | cut -d \@ -f 1)
+  TAG_PACKAGE_VERSION=$(echo "$TAG" | cut -d \@ -f 2)
+  ACTUAL_PACKAGE_VERSION=$(node -e "console.log(require('./packages/$PACKAGE_NAME/package.json').version)")
+
+  if [ "$TAG_PACKAGE_VERSION" != "$ACTUAL_PACKAGE_VERSION" ]; then
+    echo -e "\033[1;31m$PACKAGE_NAME is tagged as $TAG_PACKAGE_VERSION, but its package.json version is $ACTUAL_PACKAGE_VERSION.\033[0m"
+    exit 1
+  fi
+done
+
+echo -e "\033[0;32mAll clear!\033[0m"
+exit 0
--- a/lib/snippets/publish-lerna-independent-packages
+++ b/lib/snippets/publish-lerna-independent-packages
@@ -1,0 +1,34 @@
+#!/usr/bin/env node
+const {resolve} = require('path');
+const {execSync} = require('child_process')
+const Repository = require(resolve('.', 'node_modules', 'lerna', 'lib', 'Repository'));
+const PackageUtilities = require(resolve('.', 'node_modules', 'lerna', 'lib', 'PackageUtilities'));
+
+function npmTag(version) {
+  const isNext = ['-beta', '-alpha', '-rc', '-next'].some(distributionType =>
+    version.includes(distributionType),
+  );
+  return isNext ? 'next' : 'latest';
+}
+
+const taggedPackages = execSync('git tag --points-at HEAD')
+  .toString()
+  .trim()
+  .split('\n')
+  .map(tag => tag.split('@')[0]);
+
+const repository = new Repository('.');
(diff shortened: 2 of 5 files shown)


lib/snippets/publish-lerna-independent-packages
```diff
@@ -0,0 +1,34 @@
...
+
+const taggedPackages = execSync('git tag --points-at HEAD')
+  .toString()
+  .trim()
+  .split('\n')
+  .map(tag => tag.split('@')[0]);
+
+const repository = new Repository('.');
+const unsortedPackages = PackageUtilities.getPackages({rootPath: '.', packageConfigs: repository.packageConfigs});
+const packages = PackageUtilities.topologicallyBatchPackages(unsortedPackages, true)
```
GoodForOneFare: This is the bit that really needs JS.  We _could_ build a dependency graph in Ruby.

lib/snippets/publish-lerna-independent-packages
```diff
@@ -0,0 +1,34 @@
...
+  .reduce((acc, packageGroup) => [...acc, ...packageGroup], [])
+  .filter(({name}) => taggedPackages.includes(name));
+
+packages.forEach(({name, version}) => {
+  const command = `lerna publish --yes --skip-git --force-publish=${name} --repo-version=${version} --scope=${name} --npm-tag ${npmTag(
```
GoodForOneFare: This is a mess, but it works.  We can raise an issue in the lerna repo, and offer to improve CI workflow support.

lib/snippets/publish-lerna-independent-packages
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env node
+const {resolve} = require('path');
+const {execSync} = require('child_process')
+const Repository = require(resolve('.', 'node_modules', 'lerna', 'lib', 'Repository'));
```
GoodForOneFare: Don't know what's up with module resolution here ¯\\_(ツ)_/¯  I'll fix this if the rest of the PR is approved.

lib/snippets/assert-lerna-independent-version-tags
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
```
casperisfine: Shellcheck gives me the following report:

```
In lib/snippets/assert-lerna-independent-version-tags line 6:
  HEAD_REV=`git rev-parse HEAD`
           ^-- SC2006: Use $(..) instead of legacy `..`.


In lib/snippets/assert-lerna-independent-version-tags line 13:
  PACKAGE_NAME=$(echo $TAG | cut -d \@ -f 1)
                      ^-- SC2086: Double quote to prevent globbing and word splitting.


In lib/snippets/assert-lerna-independent-version-tags line 14:
  TAG_PACKAGE_VERSION=$(echo $TAG | cut -d \@ -f 2)
                             ^-- SC2086: Double quote to prevent globbing and word splitting.


In lib/snippets/assert-lerna-independent-version-tags line 15:
  ACTUAL_PACKAGE_VERSION=`node -e "console.log(require('./packages/$PACKAGE_NAME/package.json').version)"`
                         ^-- SC2006: Use $(..) instead of legacy `..`.


In lib/snippets/assert-lerna-independent-version-tags line 17:
  if [ $TAG_PACKAGE_VERSION != $ACTUAL_PACKAGE_VERSION ]; then
       ^-- SC2086: Double quote to prevent globbing and word splitting.
                               ^-- SC2086: Double quote to prevent globbing and word splitting.
```

Some good stuff in there.

lib/snippets/publish-lerna-independent-packages
casperisfine: It doesn't have to be in Ruby, you need node to publish those packages anyway, I'm fine with the support script being JS.

lib/snippets/publish-lerna-independent-packages
```diff
@@ -0,0 +1,34 @@
...
+const PackageUtilities = require(resolve('.', 'node_modules', 'lerna', 'lib', 'PackageUtilities'));
+
+function npmTag(version) {
+  const isNext = ['-beta', '-alpha', '-rc', '-next'].some(distributionType =>
```
lemonmade: This seems wrong, it wouldn't catching things like `1.0.0-beta.1` would it? I think maybe we just want to match any version that is not strictly `/\d+\.\d+\.\d+/` and assume that's not meant to be a `latest`-tagged version.

GoodForOneFare: Ah, yeah, yeah.  We can't use the strict regexp, because we have to support stuff like `@shopify/tinymce@4.5.5-shopify5` as latest.

I've updated the logic to use `includes`, which matches the [ruby equivalent](https://github.com/Shopify/shipit-engine/blob/master/app/models/shipit/deploy_spec/npm_discovery.rb#L50).  LMK if that makes sense to you, please.

lemonmade: 👍

PR #756, opened 2018-01-25
Show how long deploys have been locked for

davidcornu:
image

When looking at a locked stack page it's sometimes hard to tell the gravity of the situation. If a stack has been locked for the better part of a day, odds are developers shouldn't bank on things being resolved any time soon.

**Changes**

- Moves `locked_since` logic to an `after_save` callback rather than relying on `Stack#lock`/`Stack#unlock`
- Updates the banner for locked stacks

### DIFF
--- a/app/views/shipit/stacks/_banners.html.erb
+++ b/app/views/shipit/stacks/_banners.html.erb
@@ -36,7 +36,10 @@
       <div class="banner__content">
         <h2 class="banner__title">
           <i class="icon icon--lock"></i>
-          Deploys are locked by <%= stack.lock_author.name %>
+          Deploys were locked by <%= stack.lock_author.name %>
+          <% unless stack.locked_since.nil? %>
+            <%= timeago_tag(stack.locked_since, force: true) %>
+          <% end %>
         </h2>
         <p class="banner__text">
           <%= auto_link emojify(stack.lock_reason) %>
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -44,6 +44,7 @@ module Shipit
 
     before_validation :update_defaults
     before_destroy :clear_local_files
+    before_save :set_locked_since
     after_commit :emit_lock_hooks
     after_commit :emit_added_hooks, on: :create
     after_commit :emit_updated_hooks, on: :update
@@ -362,7 +363,6 @@ module Shipit
 
     def lock(reason, user)
       params = {lock_reason: reason, lock_author: user}
-      params[:locked_since] = Time.current if locked_since.nil?
       update!(params)
     end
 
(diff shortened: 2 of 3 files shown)


casperisfine:
I'm not sure I like the before_save callback that much. Was it actually necessary, or did it just seem better to you?

Other than that I like the feature.

davidcornu:
> I'm not sure I like the before_save callback that much. Was it actually necessary, or did it just seem better to you?

There are quite a few places in the code where we're setting `lock_reason` on a `Stack` without setting the other settings, which means `lock_author_id` and `locked_since` aren't set correctly. 

This meant my tests were failing as `locked_since` would be `nil` which is why I had to wrap the template code in a `nil?` check.

There are three solutions to that IMHO

1. Add a `before_save` callback which ensures it's set correctly everywhere. This has the downside of adding more concerns to the model and makes updating the `lock_reason` an acceptable way of locking a stack.
2. Make `lock_reason`, `lock_author_id` and `locked_since` private somehow and force everyone to use `Stack#lock` and `Stack#unlock`.
3. Move `lock_reason`, `locked_since`, and `lock_author_id` into a `StackLock` model and add a `lock_id` to `Stack`. `Stack#locked?` would now just be `lock_id.present?` and we could move all the authorship concerns to another model.

I chose (1) because it's the least invasive change but my preferred solution is (3) which I would gladly implement.

casperisfine:
There's also:

Override `lock_reason=` to behave a bit the same.

But yeah, I see your point.

app/views/shipit/stacks/_banners.html.erb
```diff
@@ -36,7 +36,10 @@
       <div class="banner__content">
         <h2 class="banner__title">
           <i class="icon icon--lock"></i>
-          Deploys are locked by <%= stack.lock_author.name %>
+          Deploys were locked by <%= stack.lock_author.name %>
+          <% unless stack.locked_since.nil? %>
+            <%= time_ago_in_words(stack.locked_since) %> ago
```
casperisfine: Actually I just noticed this will be slightly bugged. 

It will be on `Stack#show` which is heavily cached.

Luckily we have a helper for this purpose that use jquery.timeago. Let me find it.

casperisfine: It's `timeago_tag`: https://github.com/Shopify/shipit-engine/blob/eba7226f04cf16f55fcaa0a89696aed8ee287974/app/views/shipit/tasks/show.html.erb#L3

Once you've changed that :shipit: from me.

davidcornu: Good catch. Fixing

davidcornu:
Fixed. Merging on 🍏.

PR #757, opened 2018-02-02
Add information about who aborted a deploy

rafaelfranca:
We had a problem because someone aborted a deploy and triggered a rollback and we had no way to know who triggered the abort.

Looking at the code I found that abort was an operation that didn't presented to the users who was the author of that operation so I added a new column in Task to store who aborted it and I'm presenting in the view.

cc @kmcphillips

### DIFF
--- a/app/models/shipit/task.rb
+++ b/app/models/shipit/task.rb
@@ -14,6 +14,7 @@ module Shipit
     belongs_to :deploy, foreign_key: :parent_id, required: false # required for fixtures
 
     belongs_to :user, optional: true
+    belongs_to :aborted_by, class_name: 'User', optional: true
     belongs_to :stack, counter_cache: true
     belongs_to :until_commit, class_name: 'Commit'
     belongs_to :since_commit, class_name: 'Commit'
@@ -236,8 +237,8 @@ module Shipit
       end
     end
 
-    def abort!(rollback_once_aborted: false)
-      update!(rollback_once_aborted: rollback_once_aborted)
+    def abort!(rollback_once_aborted: false, aborted_by:)
+      update!(rollback_once_aborted: rollback_once_aborted, aborted_by_id: aborted_by.id)
 
       if alive?
         aborting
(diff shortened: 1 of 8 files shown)


app/views/shipit/deploys/_deploy.html.erb
```diff
@@ -21,6 +21,7 @@
         <span class="code-additions">+<%= deploy.additions %></span>
         <span class="code-deletions">-<%= deploy.deletions %></span>
         <% if deploy.ignored_safeties? %><span class="ignored-safeties">ignoring safeties</span><% end %>
+        <% if (deploy.aborting? || deploy.aborted?) && deploy.aborted_by %><span class="aborted-by">aborted by <%= deploy.aborted_by&.login %></span><% end %>
```
rafaelfranca: I only added to the deploy partial, should we put in all the tasks partials?

kmcphillips: Not a big deal. If the information is available is the important part.

app/models/shipit/task.rb
```diff
@@ -236,8 +237,8 @@ def request_abort
       end
     end
 
-    def abort!(rollback_once_aborted: false)
-      update!(rollback_once_aborted: rollback_once_aborted)
+    def abort!(rollback_once_aborted: false, aborted_by:)
+      update!(rollback_once_aborted: rollback_once_aborted, aborted_by_id: aborted_by.id)
```
kmcphillips: This is being opinionated that aborted_by is always present and a user object. No nil. Which I think is ok. But the column is nulable, meaning that’s only for backwards compatibility. Not going forward.

app/views/shipit/deploys/_deploy.html.erb
casperisfine: Is the `&.login` necessary? Since the `if` already check for it's presence.

Also I'd remove the `aborting? || aborted?` check, since post abort it could transition to `flapping` state or something like that.

casperisfine: Oh and the `&.` makes the build fail since Shipit doesn't require 2.3 yet.

app/models/shipit/task.rb
rafaelfranca: The column needs to be null because tasks can be in any state, for example in the running state, and tasks in states other than aborted, aborting and flapping should not have a aborted _by.

app/views/shipit/deploys/_deploy.html.erb
rafaelfranca: Fixed

PR #760, opened 2018-03-02
Implement blocking statuses

casperisfine:
As described in the README, `blocking` statuses are statuses that are required to be successful for **all the commits included in a deploy**.

NB: There is a potential performance issue right now, but fixing it would require a pretty big refactoring, I rather tackle that in a later PR if it proved itself problematic, and keep this PR short.

### DIFF
--- a/test/models/commits_test.rb
+++ b/test/models/commits_test.rb
@@ -306,6 +306,37 @@ module Shipit
       refute_predicate commit, :deployable?
     end
 
+    test "#deployable? is false if a blocking status is missing on a previous undeployed commit" do
+      blocking_commit = shipit_commits(:soc_second)
+      blocking_commit.statuses.delete_all
+
+      assert_predicate blocking_commit, :pending?
+      assert_predicate blocking_commit, :blocking?
+
+      commit = shipit_commits(:soc_third)
+      refute_predicate commit, :deployable?
+    end
+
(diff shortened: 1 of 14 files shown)


test/models/commits_test.rb
```diff
@@ -16,9 +16,9 @@ class CommitsTest < ActiveSupport::TestCase
     end
 
     test "#pull_request? detects pull requests with unusual branch names" do
-      @pr.message = "Merge pull request #7 from Shopify/bump-👉-v1.0.1\n\nBump 👉 v1.0.1"
+      @pr.message = "Merge pull request #7 from Shopify/bump--v1.0.1\n\nBump  v1.0.1"
```
DazWorrall: Lost emojis.

test/models/commits_test.rb
```diff
@@ -16,9 +16,9 @@ class CommitsTest < ActiveSupport::TestCase
...
       assert @pr.pull_request?
-      assert_equal "Bump 👉 v1.0.1", @pr.pull_request_title
+      assert_equal "Bump  v1.0.1", @pr.pull_request_title
```
DazWorrall: Another lost emoji.

test/models/commits_test.rb
byroot: Arf, damn Textmate 😢

PR #761, opened 2018-03-05
Improve deploy button captions when blocked by a status

casperisfine:
Ref: https://github.com/Shopify/shipit-engine/pull/760

In #760 I forgot to properly handle this new `deploy_state` in the deploy button rendering.

I also chose to add a `hint` for it because it might be difficult to understand.

Looks like this:

capture d ecran 2018-03-05 a 12 19 47

### DIFF
--- a/app/models/shipit/undeployed_commit.rb
+++ b/app/models/shipit/undeployed_commit.rb
@@ -9,9 +9,15 @@ module Shipit
 
     def deploy_state(bypass_safeties = false)
       state = deployable? ? 'allowed' : status.state
+
       unless bypass_safeties
-        state = 'deploying' if stack.active_task?
-        state = 'locked' if locked?
+        if blocked?
+          state = 'blocked'
+        elsif locked?
+          state = 'locked'
+        elsif stack.active_task?
+          state = 'deploying'
+        end
       end
       state
     end
@@ -30,6 +36,11 @@ module Shipit
 
     def deploy_discouraged?
       stack.maximum_commits_per_deploy && index >= stack.maximum_commits_per_deploy
+    end
+
+    def blocked?
+      return @blocked if defined?(@blocked)
+      @blocked = super
     end
   end
 end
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -27,6 +27,7 @@ en:
   deploy_button:
     hint:
       max_commits: It is recommended not to deploy more than %{maximum} commits at once.
+      blocked: This commit range includes a commit that can't be deployed.
```
wvanbergen: > can't be deployed *automatically* ?

casperisfine: Nope, not deployed at all. It's akin to a build failure. You can use emergency mode if you want to ignore the status.

wvanbergen:
I wonder if we can also use this status in the case of too many "non-deployable" commits. E.g., shipit won't deploy if 8 (configurable) commits in a row are either failing CI, or are locked because of a revert. Currently in the UI is not clear that it will never end up deploying automatically in these cases, and manual intervention is required.

casperisfine:
You have a good point, it's something we need to improve, but it's actually totally orthogonal to this new `blocking` status.

When continuous delivery is blocked (and knowns it won't be unblocked without human intervention) we need to display something, as well as fire a hook.

app/models/shipit/undeployed_commit.rb
```diff
@@ -31,5 +37,10 @@ def deploy_disallowed?
     def deploy_discouraged?
       stack.maximum_commits_per_deploy && index >= stack.maximum_commits_per_deploy
     end
+
+    def blocked?
+      return @blocked if defined?(@blocked)
+      @blocked = super
```
DazWorrall: `@blocked ||= super`? Or is this not possible on older Rubies?

casperisfine: `||=` works since pretty much Ruby's inception.

However it's a shortcut for `foo = (defined?(foo) && foo) || something`.

So it can't be used to memoize `nil` and `false` values, which I need here.

PR #764, opened 2018-03-14
Distinguish deploy timeouts from deploy failures

casperisfine:
I don't have a designer at hand, so I had to reuse the `error` icon:

capture d ecran 2018-03-15 a 13 30 53

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+* Distinguish deploy timeouts from deploy failures.
 * Added blocking statuses. If they are missing or failing, they will prevent deploy even if they were reported on any of the commits in the deploy range.
 * Fix shipit.yml updates not being taken into account for the `fetch` command.
 * Fix failing membership webhooks for non fully downcase organization names.
--- a/app/assets/stylesheets/_pages/_commits.scss
+++ b/app/assets/stylesheets/_pages/_commits.scss
@@ -281,7 +281,9 @@
 }
 
 .status--error,
-[data-deploy-status='error'] {
+[data-deploy-status='error'],
+.status--timedout,
+[data-deploy-status='timedout'] {
   border-color: #333;
 
   .status__icon {
--- a/app/assets/stylesheets/_pages/_deploy.scss
+++ b/app/assets/stylesheets/_pages/_deploy.scss
@@ -93,7 +93,8 @@
   }
 
   &[data-status="failure"],
-  &[data-status="error"] {
+  &[data-status="error"],
+  &[data-status="timedout"] {
     .deploy-banner-status {
       background-color: $bright-red;
       width: 100%;
@@ -196,7 +197,8 @@
 
   &[data-status="failed"],
   &[data-status="failure"],
-  &[data-status="error"] {
+  &[data-status="error"],
+  &[data-status="timedout"] {
     border-color: $bright-red;
   }
 
--- a/app/jobs/shipit/perform_task_job.rb
+++ b/app/jobs/shipit/perform_task_job.rb
@@ -19,6 +19,9 @@ module Shipit
       checkout_repository
       perform_task
       @task.complete!
+    rescue Command::TimedOut => error
+      @task.write("\n#{error.message}\n")
+      @task.report_timeout!(error)
     rescue Command::Error => error
       @task.write("\n#{error.message}\n")
       @task.report_failure!(error)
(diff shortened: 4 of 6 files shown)


wvanbergen:
What kind of timeout is this? When a deploy command doesn't output anything for X minutes?

casperisfine:
> When a deploy command doesn't output anything for X minutes?

Yes, which is configurable so can be handled if expected. It was asked for by cloud people.

wvanbergen:
So if k8s-deploy stops waiting for a deploy to converge, it would still be considered an error rather than a timeout, right? Just want to make sure this is clear and what the intention was

casperisfine:
> So if k8s-deploy stops waiting for a deploy to converge, it would still be considered an error rather than a timeout, right?

If by that you mean the k8s command exit with some non-zero status, then yes.

dturn:
> If by that you mean the k8s command exit with some non-zero status, then yes.

The relevant PR on our side is: https://github.com/Shopify/kubernetes-deploy/pull/244 . The non-zero status we're targeting is `2`. But that's easy for us to change if you'd prefer something else.

> I don't have a designer at hand, so I had to reuse the error icon

Can we use font awesome? If so https://fontawesome.com/icons/hourglass-end?style=solid might be better.

casperisfine:
> The relevant PR on our side is: Shopify/kubernetes-deploy#244 . The non-zero status we're targeting is 2. But that's easy for us to change if you'd prefer something else.

Ok, so that's absolutely not what I understood from our discussion, and not what this PR is doing.

You are asking for a different status based on the exit code, we can probably work something out (no idea how at this point), but that's really not the same thing.

 > Can we use font awesome? 

Maybe ? I don't know. We need a `.svg`.

dturn:
> Maybe ? I don't know. We need a .svg.

It does SVGs. Its my goto pace for icons https://fontawesome.com/get-started.

>Ok, so that's absolutely not what I understood from our discussion, and not what this PR is doing.

Glad we caught this. Do you need more info from me?

casperisfine:
> Do you need more info from me?

I guess I get it now. 

You'd like to be able to configure your stack such as:

```
if exit_code == 2 then mark_as_timedout
```

So, in Shipit context:

  - The exact exit code have to be configurable in `shipit.yml`, because and exit code can mean anything depending on the program that is ran.
  - That configuration as to be defined for a specific command and not the full file or step list, because you can run other programs before or after (e.g. `bundler`, `ejson`, etc).


So the `shipit.yml` would have to look like this:

```yaml
deploy:
  override:
    - echo starting
    - my-deploy-command: { exit: { 2: timeout }}
    - echo done
```

or something like this:

```yaml
deploy:
  override:
    - echo starting
    - my-deploy-command: { timeout_on: [2] }
    - echo done
```

I'm trying to take a step back and figure out if it makes sense at all.

dturn:
cc: @KnVerey

KnVerey:
> I'm trying to take a step back and figure out if it makes sense at all.

Thanks, we really appreciate your input. A couple more details for you to consider:
- We want to do this because timeouts and failures have really different implications for Kubernetes deploys, and being able to easily make the distinction in the overview UI would be really helpful to our users. If more information on _why_ this is the case would be useful, I would be happy to go into the details.
- If we do have a visually distinct timeout status, I think also using it for Shipit-enforced timeouts like you have done in this PR make sense.
- This applies to all cloudplatform apps, which will be most everything at Shopify, so distributing either of the things in your comment would not be workable for us. Most of our users' `shipit.yml`s are using the special `kubernetes:` syntax though, so we could do something akin to that in the command Shipit constructs for that syntax. However, not all apps are able to do that (e.g. they're deploying to multiple targets or need extra steps), so ideally we'd like this feature to work either out of the box or as trivially as possible for those apps as well.

byroot:
> ideally we'd like this feature to work either out of the box or as trivially as possible for those apps as well.

To work out of the box, we could select an exit code that if very unlikely to happen in the wild, (e.g. `77`). Other than that I can't think of anything else.

KnVerey:
> To work out of the box, we could select an exit code that if very unlikely to happen in the wild, (e.g. 77). Other than that I can't think of anything else.

I'd be happy with that solution.

casperisfine:
👌 then, I'll get back at it first thing tomorrow.

casperisfine:
I'll merge this as is, and implement what we talked about in a followup.

PR #767, opened 2018-03-19
Enforce publishConfig settings for npm package

ismail-syed:
### What is this?
With the recent feature of private npm package support via [PackageCloud](https://packagecloud.io/l/npm-registry) (PC) we're updating all our isolated system to support this (token-rotator, pipa, shopify-build, dev etc.)

### How?
This PR enforces a [`publishConfig`](https://docs.npmjs.com/files/package.json#publishconfig) option to be present and properly configured in a `package.json` before publishing. The `publishConfig` option must be present for a few reason: 

1. To configure shipit to auth with PC for our private packages, we'll need to update the global `~/.npmrc` file with a valid config & token 
```
always-auth=true
@shopify:registry=https://packages.shopify.io/shopify/node/npm/
//packages.shopify.io/shopify/node/npm/:_authToken=<PC_TOKEN>
```
* All dependencies with the `@shopify/<package_name>` scope will be fetched from PC. For public `@shopify` name-spaced packages on the npm registry, PC will proxy the install through the public npm registry if they do not exist on PC.
* Since our global config for `@shopify` is set to point to PC, `publishConfig` settings in the `package.json` allows maintainer to override the registry being published to (private/public).This allows internal package maintainers to freely choose between deploying node.js packages publicly on the public npm registry and privately on our PC server.

Below is a summary of valid `publishConfig` options. This will also be documented and added to the engineering blog somewhere [here](https://development.shopify.io/guides/gems/)

### Public package publishes
```
  "publishConfig": {
      "access": "public"
  }
```

### Private package publishes
```
  "publishConfig": {
    "access": "restricted"
  }
```

### Next steps
* Add documentation to the engineering blog on how to deploy public/private packages
* Update the global `~/.npmrc` file on shipit.


app/models/shipit/deploy_spec/npm_discovery.rb
```diff
@@ -3,6 +3,15 @@
 module Shipit
   class DeploySpec
     module NpmDiscovery
+      # https://docs.npmjs.com/cli/publish
+      PUBLIC = 'public'
+      PRIVATE = 'restricted'
+      VALID_ACCESS = [PUBLIC, PRIVATE]
+
+      NPM_REGISTRY = "https://registry.npmjs.org/"
```
ismail-syed: Using the public npm registry for public publishes since yarnkpkg is just a reverse proxy to the npm registry.

app/models/shipit/deploy_spec/npm_discovery.rb
```diff
@@ -23,18 +32,23 @@ def discover_yarn_checklist
...
           [%(<strong>Don't forget version and tag before publishing!</strong> You can do this with:<br/>
             npm version <strong>&lt;major|minor|patch&gt;</strong>
-            && git push --follow-tags</pre>)]
+            && git push --follow-tags</pre>), publish_config_checklist.first]
         end
       end
 
+      def publish_config_checklist
+        [%(<strong>Don't forget publishConfig settings in your package.json!</strong> See
+          <a target="_blank" href="https://development.shopify.io/guides/gems/">docs</a>)]
```
ismail-syed: This link will need to be updated once I get the doc up

test/models/deploy_spec_test.rb
```diff
@@ -684,9 +684,138 @@ class DeploySpecTest < ActiveSupport::TestCase
...
+    test "#package_scoped_when_private? is true when packages are @shopify scoped" do
+      @spec.stubs(:package_name).returns("@shopify/polaris")
+      @spec.stubs(:publish_config_access).returns("restricted")
+      assert @spec.package_scoped_when_private?
+    end
+
     test '#publish_npm_package checks if version tag exists, and then invokes npm deploy script' do
       @spec.stubs(:npm?).returns(true)
       @spec.stubs(:package_version).returns('1.0.0')
+      @spec.stubs(:publish?).returns(true)
```
ismail-syed: stubbing `publish?` so these test's don't fail.

app/models/shipit/deploy_spec/npm_discovery.rb
```diff
@@ -79,7 +93,73 @@ def package_version
...
+
+      def valid_publish_config_access?
+        return false if publish_config_access.blank?
+        return false unless VALID_ACCESS.include?(publish_config_access)
+        true
+      end
+
+      def valid_publish_config_registry?
+        return false if publish_config_registry.blank?
+        return false unless VALID_REGISTRY.include?(publish_config_registry)
```
GoodForOneFare: Is this possible?
- Don't require registry info
- If registry is omitted && access === 'public': 
  - `npm publish --access public --registry https://registry.npmjs.org/`
- If registry is omitted && access === 'restricted': 
  - `npm publish --access restricted --registry https://packages.shopify.io/shopify/node/npm/`

ismail-syed: I don't see registry as a documented option here: https://docs.npmjs.com/cli/publish
Even if it works, we don't have a way to deploy scoped registry (`@shopify:registry`) via CLI - or at least I haven't come across one.
Since the global npmrc will point @shopify to pacakgeCloud, we can change that to the public npm registry via CLI.

ismail-syed: I recall testing this out via CLI and it didn't work. Lemme know if it happens to work for you.

GoodForOneFare: It's documented here: https://docs.npmjs.com/misc/registry#can-i-run-my-own-private-registry

> If you then want to publish a package for the whole world to see, you can simply override the --registry option for that publish command.

GoodForOneFare: If the global .npmrc is a problem, create one on the fly inside the checked out folder.

ismail-syed: It all works, great suggestion! 

This is what I've tested locally: 
* My global `~/.npmrc` had both a write token to PC and write token to our shopify scoped npm account. The tests worked as expected
```
always-auth=true
@shopify:registry=https://packages.shopify.io/shopify/node/npm/
//packages.shopify.io/shopify/node/npm/:_authToken=<PC_TOKEN>
//registry.npmjs.org/:_authToken=<NPM_TOKEN>
```

### Publishes to *public* registry with local .npmrc 
- [x] with  `@shopify` scope:
```
always-auth=true
@shopify:registry=https://registry.npmjs.org/
```

- [x] without scoped name:

```
always-auth=true
registry=https://registry.npmjs.org/
```
	
### Publishes to *private* registry with local .npmrc
- [x] with `@shopify` scope :

```
always-auth=true
@shopify:registry=https://packages.shopify.io/shopify/node/npm/
```

app/models/shipit/deploy_spec/npm_discovery.rb
```diff
@@ -79,7 +92,67 @@ def package_version
...
+      def registry
+        if publish_config_access == "public"
+          return scoped_package? ? "@shopify:registry=#{NPM_REGISTRY}" : "registry=#{NPM_REGISTRY}"
+        end
+
+        PACKAGE_CLOUD_REGISTRY
+      end
+
+      def generate_local_npmrc
+        File.delete(local_npmrc) if File.exist?(local_npmrc)
```
ismail-syed: Not sure if this is detrimental or not. The reason i'm doing this to ensure a cached npmrc from a previous deploy isn't used. We can hit the `git clear cache`, but if we forget to do this and a private deploy goes public we will be 😞

app/models/shipit/deploy_spec/npm_discovery.rb
```diff
@@ -79,7 +92,67 @@ def package_version
...
+
+      def registry
+        if publish_config_access == "public"
+          return scoped_package? ? "@shopify:registry=#{NPM_REGISTRY}" : "registry=#{NPM_REGISTRY}"
+        end
+
+        PACKAGE_CLOUD_REGISTRY
+      end
+
+      def generate_local_npmrc
```
ismail-syed: Had trouble mocking/testing this method so I just tested `npmrc_contents` and `registry` instead.

app/models/shipit/deploy_spec/npm_discovery.rb
ismail-syed: PR has been updated to reflect what we've discussed above.

app/models/shipit/deploy_spec/npm_discovery.rb
```diff
@@ -79,7 +92,67 @@ def package_version
...
+        File.delete(local_npmrc) if File.exist?(local_npmrc)
+        File.write(local_npmrc, npmrc_contents(registry))
+      end
+
       def publish_npm_package
+        if publish?
```
GoodForOneFare: Probably a Rubier way to do this:
```ruby
return ['misconfigured-npm-publish-config'] unless publish?

generate_local_npmrc
check_tags = 'asset-npm-version-tag'
```

app/models/shipit/deploy_spec/npm_discovery.rb
```diff
@@ -79,7 +92,67 @@ def package_version
...
+        private_package = publish_config_access == PRIVATE
+        return false if private_package && !scoped_package?
+        true
+      end
+
+      def local_npmrc
+        file(".npmrc")
+      end
+
+      def npmrc_contents(registry)
```
GoodForOneFare: It feels like we're smooshing a ton of methods into `DeploySpec`.  `PackageJson` and/or `NpmRc` helper classes would be more testable, and be less cognitive load for new devs.

I'll defer to Jean because I'm maybe just not embracing what Ruby does well ¯\\_(ツ)_/¯

app/models/shipit/deploy_spec/npm_discovery.rb
```diff
@@ -79,7 +92,67 @@ def package_version
         JSON.parse(package_json.read)['version']
       end
 
+      def publish_config
```
GoodForOneFare: It feels like we're smooshing a ton of methods into `DeploySpec`.  `PackageJson` and/or `NpmRc` helper classes could have isolated tests, and have less cognitive load for new devs.

I'll defer to Jean because I'm maybe just not embracing what Ruby does well ¯\\_(ツ)_/¯

app/models/shipit/deploy_spec/npm_discovery.rb
```diff
@@ -79,7 +92,67 @@ def package_version
...
+      def package_name
+        JSON.parse(package_json.read)['name']
+      end
+
+      def scoped_package?
+        package_name.start_with?('@shopify')
+      end
+
+      def publish?
+        return false if publish_config.blank?
```
GoodForOneFare: Requiring this config for every shipit install is maybe overkill.  Could be make this a per-install option that we'd enable for `shopify/shipit`?

app/models/shipit/deploy_spec/npm_discovery.rb
```diff
@@ -79,7 +92,67 @@ def package_version
...
+      end
+
+      def valid_publish_config_access?
+        return false unless VALID_ACCESS.include?(publish_config_access)
+        true
+      end
+
+      def package_scoped_when_private?
+        private_package = publish_config_access == PRIVATE
+        return false if private_package && !scoped_package?
```
GoodForOneFare: Isn't this the same as `private_package && scoped_package?`?

`package_scoped_when_private?` reads oddly.  `private_scoped_package?` makes more sense to me.

app/models/shipit/deploy_spec/npm_discovery.rb
```diff
@@ -79,7 +92,67 @@ def package_version
...
+      end
+
+      def publish?
+        return false if publish_config.blank?
+        return false unless valid_publish_config_access?
+        return false unless package_scoped_when_private?
+      end
+
+      def valid_publish_config_access?
+        return false unless VALID_ACCESS.include?(publish_config_access)
```
GoodForOneFare: I think this is a prolix equivalent of `VALID_ACCESS.include?(publish_config_access)`.

app/models/shipit/deploy_spec/npm_discovery.rb
ismail-syed: >  We can either leave it be, or try appending the correct scope info to it.

I don't know of a case where a project would required having a `.npmrc`. If it does the project is doing something very specific. We should defer the contents of the `.npmrc` to the project if it has an `.npmrc`

app/models/shipit/deploy_spec/npm_discovery.rb
```diff
@@ -3,6 +3,14 @@
 module Shipit
   class DeploySpec
     module NpmDiscovery
+      # https://docs.npmjs.com/cli/publish
+      PUBLIC = 'public'.freeze
+      PRIVATE = 'restricted'.freeze
+      VALID_ACCESS = [PUBLIC, PRIVATE].freeze
+
+      NPM_REGISTRY = "https://registry.npmjs.org/".freeze
+      PACKAGE_CLOUD_REGISTRY = "@shopify:registry=https://packages.shopify.io/shopify/node/npm/".freeze
```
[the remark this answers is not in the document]
ismail-syed: Good point - @byroot, any suggestions on how we handle these kind of cases in shipit/shipit-engine?

lib/snippets/misconfigured-npm-publish-config
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+echo -e "\033[1;31mpublishConfig setting in your package.json is misconfigured.\033[0m"
```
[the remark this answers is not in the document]
ismail-syed: Good suggestion, this way we don't even need to link an internal doc here.

app/models/shipit/deploy_spec/npm_discovery.rb
ismail-syed: > Requiring this config for every shipit install is maybe overkill.

Doesn't this only run on all npm package deploys?  From our conversation in the pod on Friday, I thought we were okay with doing this?

GoodForOneFare: If I said that, I've changed my mind ¯\\_(ツ)_/¯  Once publishConfig is fixed, this message is just noise.  

Once `misconfigured-npm-publish-config` displays actionable feedback, we probably have a Good Enough experience.

app/models/shipit/deploy_spec/npm_discovery.rb
ismail-syed: > package_scoped_when_private? reads oddly. private_scoped_package? makes more sense to me.

`package_scoped_when_private?` ensures that the package is scoped when it's private. `private_scoped_package?` reads as if the package must be private and scoped. 

> Isn't this the same as private_package && scoped_package??

`private_package && scoped_package?` will return false if the package is public and fail `publish?`. I've added an additional test to cover this case. 

I'll comment back once it's pushed.

ismail-syed: Actually, I can do something like this, and it reads better:

```
return true if publish_config_access == PUBLIC
publish_config_access == PRIVATE && scoped_package?
```

app/models/shipit/deploy_spec/npm_discovery.rb
ismail-syed: Yes, yes it is.
Looks like I was a bit high off some ruby here...🤣

app/models/shipit/deploy_spec/npm_discovery.rb
ismail-syed: Okay removing `publish_config.blank?` will publish packages privately to PC if they don't have an explicit `publishConfig`. Which is what we wanted the default behaviour to be. I'll add some tests to cover this too. 

--- 
`publish?` will now only test `valid_publish_config_access? && package_scoped_when_private?`

I think we should update the `misconfigured-npm-publish-config` with fixes for these potential misconfigurations

How does this sound?
```
Misconfigured package.json, please ensure the following has been met:

 * If your package is private, ensure your the package name is scoped (@organization/package)
 * Please add a publishConfig with an access setting in your package.json.  e.g.:

"publishConfig:" {
   "access": "<public|restricted>"
}
```

ismail-syed: Spoke with @GoodForOneFare offline about this

We'd want a way to enforce `publishConfigs` being present in `package.json`'s but without breaking other installs of shipit. We're thinking this can be done with an environment variable that set in shopify's instace of shipit and enforce `publishConfig` with something like:

```
return 'Missing publishConfig in package.json' if ENV['ENFORCE_PUBLISH_CONFIG'] && !has_publish_config?
```

lib/snippets/misconfigured-npm-publish-config
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+echo -e "\033[1;31m Misconfigured package.json, please ensure the following has been met:\n \033[0m"
```
ismail-syed: This writes
```
 Misconfigured package.json, please ensure the following has been met:

 * If your package is private, ensure your the package name is scoped (@organization/package)
 * Please add a publishConfig with an access setting in your package.json.  e.g.,

"publishConfig:" {
   "access": "<public|restricted>"
}
```

app/models/shipit/deploy_spec/npm_discovery.rb
ismail-syed: Removed this completely in favour of the `misconfigured-npm-publish-config` script

ismail-syed:
@GoodForOneFare I think this is ready for another round of reviews, excluding comments we're waiting to hear back from Jean about.

app/models/shipit/deploy_spec/npm_discovery.rb
casperisfine: Ultimately I'd like the `*Discovery` to be standalone classes instead of modules for the reasons you stated.

In the meantime it's fine to continue to follow the current way, otherwise it will make a refactor harder.

app/models/shipit/deploy_spec/npm_discovery.rb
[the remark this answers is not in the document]
casperisfine: This should be a config in `secrets.yml`, and then accessed on runtime: `Shipit.private_npm_registry`.

app/models/shipit/deploy_spec/npm_discovery.rb
casperisfine: Same than for the other question, I'd rather rely on a config than on an env var.

app/models/shipit/deploy_spec/npm_discovery.rb
```diff
@@ -79,10 +87,67 @@ def package_version
         JSON.parse(package_json.read)['version']
       end
 
+      def publish_config
+        JSON.parse(package_json.read)['publishConfig']
+      end
+
+      def publish_config_access
+        publish_config['access'] || PRIVATE
```
GoodForOneFare: This should default to public if a private scope/repo is not defined.

app/models/shipit/deploy_spec/npm_discovery.rb
```diff
@@ -79,10 +87,67 @@ def package_version
...
+        return false if ENV['ENFORCE_PUBLISH_CONFIG'] && publish_config.blank?
+        valid_publish_config_access? && package_scoped_when_private?
+      end
+
+      def valid_publish_config_access?
+        VALID_ACCESS.include?(publish_config_access)
+      end
+
+      # ensure private packages are scoped
+      def package_scoped_when_private?
```
GoodForOneFare: I still don't know what this means.  It makes it very difficult to understand the the descriptions.

test/models/deploy_spec_test.rb
```diff
@@ -684,10 +684,104 @@ class DeploySpecTest < ActiveSupport::TestCase
...
+      refute @spec.package_scoped_when_private?
+    end
+
+    test "#package_scoped_when_private? is true when private packages are scoped" do
+      @spec.stubs(:scoped_package?).returns(true)
+      @spec.stubs(:publish_config_access).returns("restricted")
+      assert @spec.package_scoped_when_private?
+    end
+
+    test "#package_scoped_when_private? is true when public packages are not scoped" do
```
GoodForOneFare: This highlights the incomprehensibility of the method name.  The name talks about private packages, but it also provides results for public packages?

test/models/deploy_spec_test.rb
```diff
@@ -684,10 +684,104 @@ class DeploySpecTest < ActiveSupport::TestCase
...
+      assert @spec.package_scoped_when_private?
+    end
+
     test '#publish_npm_package checks if version tag exists, and then invokes npm deploy script' do
       @spec.stubs(:npm?).returns(true)
       @spec.stubs(:package_version).returns('1.0.0')
-      assert_equal ['assert-npm-version-tag', 'npm publish --tag latest'], @spec.deploy_steps
+      @spec.stubs(:valid_publish_config?).returns(true)
+      @spec.stubs(:publish_config_access).returns('restricted')
+      @spec.stubs(:generate_local_npmrc).returns(true)
```
GoodForOneFare: Does this need to return something?

test/models/deploy_spec_test.rb
```diff
@@ -684,10 +684,104 @@ class DeploySpecTest < ActiveSupport::TestCase
...
+      assert_equal ['assert-npm-version-tag', 'npm publish --tag latest --access restricted'], @spec.deploy_steps
+    end
+
+    test '#npmrc_contents returns a private package configuration' do
```
GoodForOneFare: Under what circumstances?

test/models/deploy_spec_test.rb
```diff
@@ -684,10 +684,104 @@ class DeploySpecTest < ActiveSupport::TestCase
...
+      registry = "@shopify:registry=https://packages.shopify.io/shopify/node/npm/"
+      @spec.stubs(:publish_config_access).returns('restricted')
+      assert_equal "always-auth=true\n#{registry}", @spec.npmrc_contents(@spec.registry)
+    end
+
+    test '#npmrc_contents returns a public scoped package configuration' do
```
GoodForOneFare: Under what circumstances?

app/models/shipit/deploy_spec/npm_discovery.rb
ismail-syed: Okay, renamed to `private_scoped_package`

test/models/deploy_spec_test.rb
ismail-syed: Nope, just needed to mock `generate_local_npmrc` so the test would pass. 
What I actually needed to do was mock `registry` instead - updated pr.

test/models/deploy_spec_test.rb
ismail-syed: `#npmrc_contents returns a scoped private package configuration when the package is scoped and private`

PR updated

test/models/deploy_spec_test.rb
ismail-syed: `#npmrc_contents returns a public scoped package configuration when the package is scoped and public` - updated pr

lib/shipit.rb
```diff
@@ -140,6 +140,14 @@ def host
     secrets.host.presence
   end
 
+  def enforce_publish_config
```
ismail-syed: @casperisfine Is this approach acceptable?

Assuming this is correct, the next step would be to update `shipit` with these corresponding secrets.

ismail-syed:
@casperisfine Can I get another round of review please. 

I'd like to get this merged in soon and update `shipit`.

app/models/shipit/deploy_spec/npm_discovery.rb
```diff
@@ -79,10 +85,84 @@ def package_version
         JSON.parse(package_json.read)['version']
       end
 
+      def publish_config
+        JSON.parse(package_json.read)['publishConfig']
```
casperisfine: This could fail in many ways (and I've seen it fail in prod):

  - File could contain invalid JSON
  - The JSON document could be missing that key

Also you are re-reading and re-parsing the file in many places, it's quite wastful. Please factor this to read and parse it once and memoize it.

lib/shipit.rb
```diff
@@ -140,6 +140,14 @@ def host
...
+    secrets.enforce_publish_config.presence ? secrets.enforce_publish_config : nil
+  end
+
+  def private_npm_registry
+    secrets.private_npm_registry.presence ? secrets.private_npm_registry : nil
```
casperisfine: This is redundant, `#presence` returns `private_npm_registry` or `nil` already.

lib/shipit.rb
```diff
@@ -140,6 +140,14 @@ def host
     secrets.host.presence
   end
 
+  def enforce_publish_config
+    secrets.enforce_publish_config.presence ? secrets.enforce_publish_config : nil
```
casperisfine: Same here.

app/models/shipit/deploy_spec/npm_discovery.rb
```diff
@@ -79,10 +85,84 @@ def package_version
...
+        if publish_config_access == PUBLIC
+          return "#{prefix}=#{NPM_REGISTRY}"
+        end
+
+        "#{prefix}=#{Shipit.private_npm_registry}"
+      end
+
+      def generate_local_npmrc
+        contents = npmrc_contents(registry)
+        File.write(local_npmrc, contents) unless local_npmrc.exist?
```
casperisfine: That seem wrong. A discovery module shouldn't write content on disk. And it's not executed at the proper time & place anyway so it won't be persisted.

In short, when a commit is merged we clone the repo, run the discovery modules, cache the result as JSON and delete the repo.


So if a file has to be written, it should be done by one of the commands executed.

app/models/shipit/deploy_spec/npm_discovery.rb
```diff
@@ -79,10 +85,84 @@ def package_version
...
+      def local_npmrc
+        file(".npmrc")
+      end
+
+      def npmrc_contents(registry)
+        "always-auth=true\n#{registry}"
+      end
+
+      def registry
+        prefix = scoped_package? ? "@shopify:registry" : "registry"
```
casperisfine: Sorry but hardcoding `@shopify:registry` is not acceptable, because Shopify is not the sole user of Shipit.

I know very little about NPM & such, so if the way we're doing package publishing is specific to our organization, then this code shouldn't be inside `shipit-engine`, you can implement all that inside our private installation of Shipit.

If it's the way most people do package publishing, then go ahead but you'll have to make this org name configurable somehow.

ismail-syed: Oh yes, good point. I'll pull out the private scope into shipit as well.

lib/shipit.rb
ismail-syed: Neat, TIL!

app/models/shipit/deploy_spec/npm_discovery.rb
ismail-syed: > File could contain invalid JSON

If a npm library contains invalid json from a `package.json` it would have been flagged well before this line. (ie. CI from a failed `yarn install`) .

> The JSON document could be missing that key

All references of `publish_config` fail appropriately if `publish_config.blank?`. This is covered in tests too. 

I've memoized all the file reads introduced from this PR in https://github.com/Shopify/shipit-engine/pull/767/commits/94b1b9057c6ceb01696b758a458fddf7891e7cd7.

app/models/shipit/deploy_spec/npm_discovery.rb
ismail-syed: We spoke offline regarding this. 
This has been addressed with a shell script (https://github.com/Shopify/shipit-engine/pull/767/commits/be7ba76e950d71742a8785c0224e9a095485328f)

ismail-syed:
@casperisfine, ready for another 👀 - thanks!

Issue #766, opened 2018-03-16
Merge queue revalidate_after send the uses back at the end of the queue


tjoyal:
Waiting 60min to be asked if you are still staring to your screen to be sent at the end of the line after those that just arrived is not making me very happy.

If that something fixable?

casperisfine:
[Pull requests are ordered by `merge_requested_at asc` to be merged](https://github.com/Shopify/shipit-engine/blob/ba0b88e19e45331565731035adc6a3c2f2f3ae7b/app/models/shipit/pull_request.rb#L50), and [when you revalidate `merge_requested_at` is not modified](https://github.com/Shopify/shipit-engine/blob/ba0b88e19e45331565731035adc6a3c2f2f3ae7b/app/models/shipit/pull_request.rb#L96-L98). 

So the behavior you describe isn't expected.

However it has been reported before: https://github.com/Shopify/shipit-engine/issues/736, so I'm inclined to think there might be a bug or at the very least a poor UX that cause people to de-queue & re-queue instead of actually revalidating.

However all the reports I got so far didn't allow me to figure that out. If you remember which PR it is please open an issue on the private repo with the details so I try to figure out what happened.

tjoyal:
Reading some of the code I have a theory:
- 10 in queue, 5 need revalidation
- deploy finish, 5 first are asked to be re-validated and the next 5 gets auto merged

This would give the impression the fist 5 lost their priority?

That would mean either:
- Works as expected
or
- UX need a way to ask for "revalidation" slightly before or at expiry and not when it try to merge it?

dylanahsmith:
It does look like MergePullRequestsJob would skip over jobs that need to be revalidated, so I think @tjoyal's theory is right.  However, I have also been seeing the list of pull requests in the merge queue have their enqueued duration get updated to a short time ago after revalidation, so there might be two problems.

Could the pull requests be briefly transitioning to the rejected state then the pending state, which would update the `merge_requested_at` time?  I don't see why else `merge_requested_at` would be changing based on the code, but also nothing stands out as a reason why this would happen.

BoGs:
Is anyone looking at this? If not I am happy to.

wvanbergen:
Nobody from shipping pipeline is looking at this currently. Feel free to see if you can find the issue and post your findings here.

casperisfine:
Actually the problem have been identified already https://github.com/Shopify/shipit/issues/323

[...]

Issue #736, opened 2017-10-05
Merge queue: merge order is LIFO instead of FIFO


bdewater:
For a queue I was expecting pull requests to be merged in the order their merge was requested. This doesn't seem to happen.

IMO this is also a desirable property to have because eventually the `revalidate_after` logic will kick in if a PR keeps getting pushed back in line.

casperisfine:
It's supposed to be FIFO. If it's not, it's bug.

bdewater:
It seems that for some reason the revalidation message never arrived and this caused my PR to be stuck in the queue. Other PRs were observed to be merged in FIFO order.

bdewater:
Not sure if I'm just terribly unlucky but after requeueing I noticed the behaviour again 🤔

PR #773, opened 2018-03-22
Revalidate merge request as soon as they are expired

casperisfine:
Fix: https://github.com/Shopify/shipit-engine/issues/766
Fix: https://github.com/Shopify/shipit-engine/issues/736

As described here: https://github.com/Shopify/shipit-engine/issues/766#issuecomment-375257722

Previously we were only triggering the revalidation once the stack was allowing merges again. The idea was to avoid spamming users with revalidation requests in case a stack would be locked for a very long time.

However the downside is that on unlock old PRs would be asked for revaliation, and newer PRs would be merged instead, resulting in an unfair situation.

This PR change the behavior so that we ask for revalidation even if the stack is still locked.

Ideally we'd have some grave period so that if you respond under a minute or 2 your PR is never considered expired, but this should be a big improvement already.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+* Revalidate merge request as soon as they are expired, and not when they could be merged again.
 * Allow to configure Shipit to intepret some exit statuses as timeouts instead of regular failures.
 * Distinguish deploy timeouts from deploy failures.
 * Added blocking statuses. If they are missing or failing, they will prevent deploy even if they were reported on any of the commits in the deploy range.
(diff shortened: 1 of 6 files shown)


DazWorrall:
Test failure looks legit.

casperisfine:
@DazWorrall indeed, I forgot to run the entire test suite, I was focused on the job itself -_-.

I fixed it.

BoGs:
I think though it could still end up in the similar scenario if the stack is not locked. If it is locked it iterates over the PR and requests the revalidation which will be good. But I think that once `revalidate` the event goes to `spy` IRC pinging you a URL that calls the api/pull_requests that does a re-enqueue. 

Also there is the situation where if the stack is not locked and it has 10 PR as pending the iterator pings the first 2 the rest are fine then goes to merge them but there could be less then a second, causing the 2 PR to be dropped and re-enqueued.

casperisfine:
@BoGs 

Yes, if you are very unlucky there is still room for this to happen, that's why I mentioned a grace period could be implemented in followup.

However note that this job is triggered every minutes, so the window for this to happen is much smaller.

BoGs:
👍

PR #774, opened 2018-04-03
Fixing scope tag in lerna command

tzvipm:
For packages with names like `@scope/package-name`, we need to use `@scope` as the scope, whereas we currently use `@scope/package-name`.


tzvipm:
Sorry, it seems I need to dig into this further.

PR #775, opened 2018-04-03
Switching to npm for lerna publishing

tzvipm:
We (i.e. @ismail-syed) have been working on getting yarn publishing working for private packages, but have not got it working yet.

As such, we decided to switch to using `npm` for publishing with `lerna` as well, since we do have that working properly for mini-repos™️.

In doing this, we also realized that we're making use of utilities in the `lerna` package that have been changed in in `lerna 3.x`. As such, we pin the lerna version to `2.9.x`.

### DIFF
--- a/app/models/shipit/deploy_spec/lerna_discovery.rb
+++ b/app/models/shipit/deploy_spec/lerna_discovery.rb
@@ -41,7 +41,7 @@ module Shipit
         file('lerna.json')
       end
 
-      def lerna_version
+      def repo_version
         lerna_config = lerna_json.read
         JSON.parse(lerna_config)['version']
       end
@@ -55,12 +55,17 @@ module Shipit
       end
 
       def publish_lerna_packages
-        return publish_independent_packages if lerna_version == 'independent'
+        return publish_independent_packages if repo_version == 'independent'
         publish_fixed_version_packages
+      end
+
+      def check_lerna
+        'assert-acceptable-lerna-version'
       end
 
       def publish_independent_packages
         [
+          check_lerna,
           'assert-lerna-independent-version-tags',
           'publish-lerna-independent-packages',
         ]
@@ -69,7 +74,7 @@ module Shipit
       def publish_fixed_version_packages
         check_tags = 'assert-lerna-fixed-version-tag'
         # `yarn publish` requires user input, so always use npm.
-        version = lerna_version
+        version = repo_version
         publish =
           "node_modules/.bin/lerna publish " \
           "--yes " \
@@ -78,7 +83,7 @@ module Shipit
           "--force-publish=* " \
           "--npm-tag #{dist_tag(version)}"
 
-        [check_tags, publish]
+        [check_lerna, check_tags, publish]
       end
     end
   end
--- a/lib/snippets/assert-acceptable-lerna-version
+++ b/lib/snippets/assert-acceptable-lerna-version
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+VERSION=$(node --eval "console.log(require('./lerna.json').lerna);")
+echo -e "\033[0;32mTrying to publish using lerna $VERSION\033[0m"
+
+echo $VERSION | grep "^2\.9\.[0-9][0-9]*$" > /dev/null
+if [ $? != '0' ]; then
+  echo -e "\033[1;31mYou are using an unsupported version of lerna; only 2.9.x is supported.\033[0m"
+  exit 1
+fi
+
+echo -e "\033[0;32mAll clear!\033[0m"
+exit 0
--- a/lib/snippets/publish-lerna-independent-packages
+++ b/lib/snippets/publish-lerna-independent-packages
@@ -28,10 +28,12 @@ const packages = PackageUtilities.topologicallyBatchPackages(unsortedPackages, t
   .reduce((acc, packageGroup) => [...acc, ...packageGroup], [])
   .filter(({name}) => taggedPackages.includes(name));
 
-packages.forEach(({name, version}) => {
-  const command = `node_modules/.bin/lerna publish --yes --skip-npm=false --skip-git --force-publish=${name} --repo-version=${version} --scope=${name} --npm-tag=${npmTag(
-    version,
-  )}`;
+packages.forEach(({name, version, isPrivate}) => {
+  const access = isPrivate ? 'restricted' : 'public';
+
+  const command = `node_modules/.bin/lerna exec --scope=${
+    name
+  } -- npm publish --tag ${npmTag(version)} --access ${access}`;
 
   // eslint-disable-next-line no-console
   console.log(command);
--- a/lib/snippets/publish-lerna-independent-packages
+++ b/lib/snippets/publish-lerna-independent-packages
@@ -28,12 +28,10 @@ const packages = PackageUtilities.topologicallyBatchPackages(unsortedPackages, t
   .reduce((acc, packageGroup) => [...acc, ...packageGroup], [])
   .filter(({name}) => taggedPackages.includes(name));
 
-packages.forEach(({name, version, isPrivate}) => {
-  const access = isPrivate ? 'restricted' : 'public';
-
+packages.forEach(({name, version}) => {
   const command = `node_modules/.bin/lerna exec --scope=${
     name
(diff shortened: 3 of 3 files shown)


lib/snippets/publish-lerna-independent-packages
```diff
@@ -28,10 +28,12 @@ const packages = PackageUtilities.topologicallyBatchPackages(unsortedPackages, t
   .reduce((acc, packageGroup) => [...acc, ...packageGroup], [])
   .filter(({name}) => taggedPackages.includes(name));
 
-packages.forEach(({name, version}) => {
-  const command = `node_modules/.bin/lerna publish --yes --skip-npm=false --skip-git --force-publish=${name} --repo-version=${version} --scope=${name} --npm-tag=${npmTag(
-    version,
-  )}`;
+packages.forEach(({name, version, isPrivate}) => {
+  const access = isPrivate ? 'restricted' : 'public';
```
tzvipm: should I even bother with this, or should we just use the package.json config for `npm` publishing as specified here: https://development.shopify.io/guides/package_cloud_nodejs#?

IMO we should keep this, bc it seems to allow using the `private: true` or `private: false` flag in the package.json file which is more intuitive.

ismail-syed: where is `isPrivate` being set?

tzvipm: It comes from the `lerna` utility that we're using. That said, this is going to change in `lerna` version `3` 🤔: https://github.com/lerna/lerna/commit/75e3a4aa3ad879e8f26695b85c52dc931e1f4eb8

TheMallen: `packages` is an array of package descriptions generated by lerna's `PackageUtilities.topologicallyBatchPackages`

tzvipm: maybe we can just drop the `access` flag entirely and rely on the `publishConfig`?

tzvipm: I think the best approach is to just let lerna tell us using the current api and worry about upgrading when we upgrade. That said, I'm going to push a commit in a sec that pins the lerna version as part of the `lerna_discovery` script.

ismail-syed: The `private`, `package.json` option is very misleading. It doesn't dictate whether a package is published privately or publicly. It prevents deploys entirely - [docs](https://docs.npmjs.com/files/package.json#private)

TheMallen: 💯

ismail-syed:
To clarify/document, the `yarn publish` command is causing the issue here. It was working last week because shipit's default registry was the public npm registry.  

After the work for private npm package support was [introduced](https://github.com/Shopify/shipit-engine/pull/767), the default registry referenced by the `@shopify` scope has been [updated](https://github.com/Shopify/cookbooks/pull/16297/files#diff-6165c7bebe44d60671ef478a1e8a08adR2) to reference the PackageCloud registry.

Looking at the yarn issues, there are some open issue regarding the `publishConfig` setting not being respected by `yarn publish`. See https://github.com/yarnpkg/yarn/issues/5310#issuecomment-372084775. 

____

I see there's a [`--registry`](https://github.com/lerna/lerna#--registry-registry) flag in lerna - this may be worth testing out if you're interested. 
eg. `--registry "https://packages.shopify.io/shopify/node/npm/"`

app/models/shipit/deploy_spec/lerna_discovery.rb
```diff
@@ -41,7 +41,7 @@ def lerna_json
         file('lerna.json')
       end
 
-      def lerna_version
+      def repo_version
```
GoodForOneFare: I'd rather stick with `lerna_version`.  All of these methods [get munged into a single Ruby module](https://github.com/Shopify/shipit-engine/blob/f0dcc4aa4eac152bf437b562fc4857752ed6d4c8/app/models/shipit/deploy_spec/file_system.rb#L3), so debugging is easier when generic names aren't used.

This change will make sense once `Shipit::DeploySpec::FileSystem` is refactored.

lib/snippets/assert-acceptable-lerna-version
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+VERSION=$(node --eval "console.log(require('./lerna.json').lerna);")
+echo -e "\033[0;32mTrying to publish using lerna $VERSION\033[0m"
+
+echo $VERSION | grep "^2\.9\.[0-9][0-9]*$" > /dev/null
+if [ $? != '0' ]; then
+  echo -e "\033[1;31mYou are using an unsupported version of lerna; only 2.9.x is supported.\033[0m"
```
GoodForOneFare: Why isn't it supported?  What did 3.x remove?  Is this a temporary measure, or are we stuck permanently on 2.9?

lib/snippets/publish-lerna-independent-packages
```diff
@@ -29,9 +29,9 @@ const packages = PackageUtilities.topologicallyBatchPackages(unsortedPackages, t
   .filter(({name}) => taggedPackages.includes(name));
 
 packages.forEach(({name, version}) => {
-  const command = `node_modules/.bin/lerna publish --yes --skip-npm=false --skip-git --force-publish=${name} --repo-version=${version} --scope=${name} --npm-tag=${npmTag(
-    version,
-  )}`;
+  const command = `node_modules/.bin/lerna exec --scope=${
```
GoodForOneFare: `lerna publish --npm-client=npm` should also work.

ismail-syed: Will this then up running `npm publish` instead `yarn publish`?

The [docs](https://github.com/lerna/lerna#--npm-client-client) may be outdated, but it seems that setting npm as the client will defer dependency installs to `npm install`.

lib/snippets/assert-acceptable-lerna-version
TheMallen: lerna 3 is in beta, and it changes the `.isPublic()` method to just be a `.public` getter

lib/snippets/publish-lerna-independent-packages
GoodForOneFare: The flow is:
* [npm-client is read from the command line in `PublishCommand`](https://github.com/lerna/lerna/blob/2.x/src/commands/PublishCommand.js#L175)
* [config is passed to `publishTaggedInDir`](https://github.com/lerna/lerna/blob/2.x/src/commands/PublishCommand.js#L729)
* [client is used in an `exec` call](https://github.com/lerna/lerna/blob/a764945f7ebbac4e91dc9042b910085436daa55a/src/NpmUtilities.js#L155)

So I'm guessing it would work.  Haven't tried, though.

lib/snippets/assert-acceptable-lerna-version
GoodForOneFare: It sounds like that should be easy to support ¯\\_(ツ)_/¯  Searching lerna's code, I don't see any `public` references that look meaningful, so I can't say for sure.

TheMallen: I mean we could just do `package.isPublic && package.isPublic() ||  package.public`

lib/snippets/publish-lerna-independent-packages
TheMallen: Interesting! Maybe I'll put up a PR with that implementation, since Tzvi won't be here for the next couple days anyway.

ismail-syed: Ooo, excellent. 

It might be worth testing out `lerna publish --npm-client "npm" --registry "https://packages.shopify.io/shopify/node/npm/"`.

lib/snippets/assert-acceptable-lerna-version
tzvipm: I think the entire package structure that were relying on changed too, no?

lib/snippets/publish-lerna-independent-packages
tzvipm: Perfect! We should still look at pinning the lerna version, though, because the require tree changes too. They're switching to a monorepo afaict.

GoodForOneFare: In the future, can you document stuff like this in PR descriptions, please?

PR #776, opened 2018-04-04
change ci config to use shopify-build in the readme

doodzik:
This PR changes the configuration of the ci to use shopify-build instead of circle-ci in the README.
I propose this change as we want to encourage the usage of shopify-build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -385,7 +385,7 @@ For example:
 ```yml
 ci:
   require:
-    - ci/circleci
+    - buildkite/:pipeline_name
 ```
 
 **<code>ci.hide</code>** contains an array of the [statuses context](https://developer.github.com/v3/repos/statuses/) you want Shipit to ignore.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -403,7 +403,7 @@ For example:
 ```yml
 ci:
   allow_failures:
-    - ci/circleci
+    - buildkite/:pipeline_name
```
sander-m: maybe `buildkite/<pipeline_name>` is more clear?

doodzik: I prefer going with a colon (especially in routes) as it bandwagons the [rails routing param syntax](http://guides.rubyonrails.org/routing.html).
There are two places where the `<placeholder_var>` is used in the `README`.
If you think it is better to keep consistency with the rest of the document I will change this.

casperisfine:
> I propose this change as we want to encourage the usage of shopify-build.

The thing is `shipit-engine` is an open source project, and ShopifyBuild only available internally. So it doesn't really make sense, and would probably be confusing for the general public.

doodzik:
Makes sense 👍

PR #777, opened 2018-04-05
🛠️ use npm as the publish client for lerna

TheMallen:
Currently our lerna publishing is broken with private package clouds.

In [this PR](https://github.com/Shopify/shipit-engine/pull/775) it was suggested we use the `--npm-client` flag. This PR tries that.

I was able to run this script locally and it seemed to work and properly use `npm publish`.


GoodForOneFare:
I think it's safe to 🚢without a second review.  If it doesn't work, we can revert.

PR #781, opened 2018-04-10
Remove 'always-auth=true'

ismail-syed:
Remove the `always-auth=true` setting in the local `.npmrc` file that is generated on deploy. It's caused some troubles. 

Originally, we set this to true because PackageCloud [docs](https://packagecloud.io/docs) suggest the following: 

> The always-auth field is required when using Yarn with a private repository.

However, to support some project using npm on shopify-build, we've removed this [here](https://github.com/Shopify/shopify-build/pull/1056). Also, looking at yarn [source](https://github.com/yarnpkg/yarn/blob/d2302073e9e9554a2977f3cac2bf5a156859b50f/src/registries/npm-registry.js#L147) `auth-always` it seems that `auth-always=true` is required for non-scoped private packages. All our packages are all `@shopify` scoped so this shouldn't affect us. I ran some small tests locally, to verify this.

Another related issue was brought to my attention today. The `shopify/easdk` had trouble deploying privately.  [shipit logs](https://shipit.shopify.io/shopify/easdk/production)

```
ENEEDAUTH
auth required for publishing
You need to authorize this machine using 'npm adduser'
``` 

I'm not sure why this happened now, and not last week when we beta tested a number of private packages because the `auth-always=true` was set this whole time. One of my hunches to fix this was removing `always-auth=true` since it fixed our issue in  `shopify-build`. 

I was able to validate this by checking in a local `.npmrc` that doesn't `always-auth=true`. [shipit logs](https://shipit.shopify.io/shopify/npm-package-test/test)

@gf3, you can validate this (and unblock yourself, sorry!) by checking in an `.npmrc` into your repo with the following:
 
```
@shopify:registry=https://packages.shopify.io/shopify/node/npm/
```

### DIFF
--- a/app/models/shipit/deploy_spec/npm_discovery.rb
+++ b/app/models/shipit/deploy_spec/npm_discovery.rb
@@ -143,10 +143,6 @@ module Shipit
         file(".npmrc")
       end
 
-      def npmrc_contents(registry)
-        "always-auth=true\n#{registry}"
-      end
-
       def registry
         scope = Shipit.npm_org_scope
         prefix = scoped_package? ? "#{scope}:registry" : "registry"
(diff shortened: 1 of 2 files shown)


gf3:
thanks @ismail-syed

PR #782, opened 2018-04-11
Convert all command overrides to strings

### LINKED ISSUE #346: Command overrides should always be treated as string
``` yaml
deploy:
  override:
    - true
```

Will obviously be parsed as `true` not `"true"`, which breaks shipit.

cc @graemej

SeanPrashad:
Fixes #346 - Updated all override commands to be treated as strings. Also updated the respective documentation to reflect as such.


byroot:
That's not what #346 is about.

PR #783, opened 2018-04-13
Fix some Rails/InverseOf offenses

bdewater:
As author of the cop I felt I had to 😄 The remaining offenses are false positives which should go away when Rubocop 0.55 is released, see https://github.com/bbatsov/rubocop/pull/5726


bdewater:
Fixed a few more instance and changed a bad one, could you please take a quick second look @casperisfine

Issue #495, opened 2015-09-24
Finalize migration to organization hooks


byroot:
We're already using them for monitoring user memberships.

Using them for push and status event would simplify things quite a lot, we wouldn't have to setup new hooks for stacks.

It also open up interesting possibilities, like for example automatically creating stack when a shipit.yml is added.

One important concern here that I still ned to figure out is the migration path for existing installations.

cc @dwradcliffe

dwradcliffe:
For migration, I assume you have a task to create the org hook somewhere, and after it does that it can attempt to remove all hooks on individual repos.

dwradcliffe:
Also cc: @baldowl

byroot:
Update: We can't rely solely on Org hooks because there is no "User hooks". So repo hooks are the only way to track repositories owned by regular users.

Org hooks if available should be preferred though, simply because it's less state to track on GitHub side.

dwradcliffe:
:+1:

PR #784, opened 2018-04-13
Convert to a GitHub app

casperisfine:
Fix: https://github.com/Shopify/shipit-engine/issues/495

GitHub apps are the new recommended API, and offer adaptive rate limiting based on the number of users and repos.

This change will requires a relatively backward compatibility breaking migration. So before it's released we'll have to write some migration guide.

Issues:
  - [ ] `Shipit.user` is hard to replicate. Apps do have a "bot user", but I can't seem to find a good way to retrieve it. So far I've use the API to merge PRs and then grabbed the user id from there. Edit: seems like it was asked for before and nothing much happened: https://platform.github.community/t/obtaining-the-id-of-the-bot-user/2076
  - [ ] Ocotkit isn't really ready for bot users either, patch submitted upstream: https://github.com/octokit/octokit.rb/pull/1005
  - [ ] The setup docs have to be redone, as well as a migration guide.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -25,6 +25,9 @@ Rails/OutputSafety:
 Rails/DynamicFindBy:
   Enabled: false
 
+Rails/Delegate:
+  Enabled: false
+
 Rails/ApplicationRecord:
   Enabled: false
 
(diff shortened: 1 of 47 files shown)


casperisfine:
We've been running this branch successfully for a while now, so the plan is to polish the remaining issues for a 0.22 release.

PR #785, opened 2018-04-18
Don't allow k8s deployments to last more than 15 minutes

dturn:
In some circumstances kubernetes deploys can flap in a way that will cause kubernetes-deploy to believe they are making progress even though they will never succeeded. This defaults a hard time out of 15 minutes since that should be more than enough time for most use cases.

### DIFF
--- a/app/models/shipit/deploy_spec/kubernetes_discovery.rb
+++ b/app/models/shipit/deploy_spec/kubernetes_discovery.rb
@@ -25,10 +25,16 @@ module Shipit
 
       private
 
+      def timeout_duration
+        duration = kube_config.fetch('timeout', '900s')
+        Duration.parse(duration).to_i if duration.present? && duration != 'false'
+      end
+
       def discover_kubernetes
         return if kube_config.blank?
 
-        cmd = ["kubernetes-deploy", "--max-watch-seconds=#{60 * 15}"]
+        cmd = ["kubernetes-deploy"]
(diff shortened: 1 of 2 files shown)


casperisfine:
> This defaults a hard time out of 15 minutes since that should be more than enough time for most use cases.

Could we make this editable though ? I'm fine with opinionated default, but you must be able to override them in case they don't work for you.

e.g. something like:

```yml
kubernetes:
  namespace: my-app-production
  context: my-context
  timeout: 20m
```

dturn:
@casperisfine Added the config option. Is it possible to 🎩  here?

app/models/shipit/deploy_spec/kubernetes_discovery.rb
```diff
@@ -25,10 +25,16 @@ def discover_task_definitions
 
       private
 
+      def timeout_duration
+        duration = kube_config.fetch('timeout', '900s')
+        Duration.parse(duration).to_i if duration.present? && duration != 'false'
```
casperisfine: I think it would be `false` instead of `'false'`

dturn: `false.present? => false` I was worried someone would stick the string `'false'` into the yaml. If that's not a worry, I'll remove the second check.

casperisfine: It's definitely not a worry.

PR #786, opened 2018-04-24
Use stopwatch icon from polaris for timed out

dturn:
Use the stopwatch icon copied from the polaris-rails project for timed out status. I think this is a better fit than `!`.

Not really sure how to 🎩 to make sure this isn't actually a poor choice. (Also not sure if there are issues with just taking the icon from polaris)


casperisfine:
So on the top of my CSS skills I was about to make it look like this:

capture d ecran 2018-04-24 a 17 32 07

Still ugly if you ask me, but tolerable I'd say.

I pushed the change in this branch, if it works for you it works for me.

app/assets/stylesheets/_pages/_commits.scss
```diff
@@ -287,7 +287,9 @@
   border-color: #333;
 
   .status__icon {
```
KnVerey: Maybe I'm misunderstanding how CSS rules work, but won't this also change the icon for the actual error status?

byroot: Yeah you're right. We should split the rule for error & timedout, nice catch.

casperisfine: Fixed.

Issue #521, opened 2016-02-02
Allow creation of tag and / or releases on deploy.


byroot:
### Context:

For many reasons, mostly security, we don't want to let shipit push anything inside the source repository.

The big downside is that it prevents us from creating a tag on deploy. This is especially useful for libraries like gems for example. The `release` task provided by bundler try to create a tag and push it, which leads us to nasty hacks like this: https://github.com/Shopify/shipit-engine/blob/558a8c9aa6399f44b5bbc2f9d8d1b37a9db114cf/lib/snippets/assert-gem-version-tag#L8-L12
### Proposal

Maybe we need an alternative to shell commands. They are very versatile, but sometime limited.
If we could also define "macros" in Shipit and use them from the `shipit.yml`, it could allow us to expose more powerful possibilities, without exposing raw access to dangerous resources.

I'm thinking something like:

```
deploy:
  override:
    - shipit release-gem
```

`shipit` here not being a real binary, but calling a method inside the application. That way we can use Shipit's token to create a tag, without giving deploy script unlimited access to Shipit's token.

This would allow to replace entirely the "snippets": https://github.com/Shopify/shipit-engine/tree/558a8c9aa6399f44b5bbc2f9d8d1b37a9db114cf/lib/snippets

It might also be a first step for https://github.com/Shopify/shipit-engine/issues/297

I'll think about it for a while before doing anything though. Opinions are welcome.

cc @tahnok

PR #800, opened 2018-06-13
Set the true origin remote on working directory clones

casperisfine:
Fix: https://github.com/Shopify/shipit-engine/issues/521

As described in #521, we never wanted people to be able to push back to the repository, and were using a read only access. 

However it's getting clear that it is actually needed if we want to make gem and other packages releasing better. Most release tool tend to tag stuff & push these tags to the repo.

The push access should managed using GitHub permissions anyway, so it's up to each install to allow or disallow it either globally or on a per repo basis. And if Shipit doesn't have the permission to push, at least now the error message will be easier to understand.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+* Set the proper origin remote on working directory repositories.
+
 # 0.23.0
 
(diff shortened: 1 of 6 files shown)


casperisfine:
> A changelog entry is important for this I think 👍

Uh, I was sure I added one... I just forgot to push my amended commit...

Issue #454, opened 2015-07-10
Auto deploy + auto version increment for ruby gems


jasonhl:
We've got a project structured as a ruby gem. On a merge to master, a hook fires to Packer. It then builds a package which is installed using ark and `gem install` via a Chef cookbook. This works reasonably well -- except when you forget to bump the version in the gemspec.

I'd like to replace all this with Shipit. When a change is merged, CI runs and if it passes, the change is auto deployed and the version auto incremented. Ideally, the cookbook bit is also replaced by the deploy process which just installs the new gem on the target box(es).

fyi: @byroot  @csfrancis

dwradcliffe:
You could probably use https://github.com/svenfuchs/gem-release to bump the version

byroot:
I think the rubygem supports needs a bit of polish. One thing that really sucks with the current gem release script:

Since Shipit model is to do a full clone from the local git cache for each deploy, the  `origin` remote is actually the git cache, not github. So we can't easily push tags...

cc @etiennebarrie since we were talking about it this afternoon.

It would be nice to revamp the whole thing.

Issue #202, opened 2014-08-07
fetch-gem-version is not working properly


byroot:
It manage to map the last version on rubygems with a SHA on GitHub, but the problem is that `rake release` creates an annotated tag `git tag -a`, so that tag poing to a commit that is not on master, so shipit doesn't know that commit.

Not sure what we can do, I have no idea how I can map that annotated tag to an actual revision in master.

PR #801, opened 2018-06-13
Make the default gem release method CD compatible, and assume it can push tags to the origin repo

casperisfine:
Fix: https://github.com/Shopify/shipit-engine/issues/454
Closes: https://github.com/Shopify/shipit-engine/issues/202

This new snippet:

  - Extract the gem name and version from the `*.gemspec`
  - Check if that version was already published on rubygems
  - If it was simply exit 0 with a message
  - If it wasn't then `exec` to the supplied release command (or `bundle exec rake release` if none was supplied).

This means rubygems stack will now be able to turn on continuous delivery. Whenever they bump the version string in their gemspec, a new version will be released automatically.

I leave the old snippets in place in case some people were explicitly using them.

### DIFF
--- a/lib/snippets/release-gem
+++ b/lib/snippets/release-gem
@@ -1,0 +1,24 @@
+#!/usr/bin/env ruby
+
+require 'net/https'
+require 'uri'
+
+module RubygemsAPI
+  extend self
+
+  def published?(name, version)
+    uri = URI.parse("https://rubygems.org/api/v2/rubygems/#{name}/versions/#{version}.json")
+    Net::HTTP.get_response(uri).is_a?(Net::HTTPSuccess)
+  end
+end
+
+spec_path, *release_command = ARGV
+release_command = %w(bundle exec rake release) if release_command.empty?
+
+spec = Gem::Specification.load(spec_path)
+if RubygemsAPI.published?(spec.name, spec.version)
+  puts "#{spec.name} version #{spec.version} is already published."
+  exit 0
+else
+  exec(*release_command)
+end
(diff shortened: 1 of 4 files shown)


lib/snippets/release-gem
```diff
@@ -0,0 +1,46 @@
...
+  def published?(name, version)
+    uri = URI.parse("https://rubygems.org/api/v2/rubygems/#{name}/versions/#{version}.json")
+    Net::HTTP.get_response(uri).is_a?(Net::HTTPSuccess)
+  end
+end
+
+spec_path, *release_command = ARGV
+release_command = %w(bundle exec rake release) if release_command.empty?
+
+spec = FakeSpec.load(spec_path)
```
wvanbergen: Rather than parsing the gemspec, would it make sense to `gem build` and inspect the resulting `.gem`?

wvanbergen: `Gem::Package.new('pkg/mygem-2.2.0.gem').spec.version` will read the version from the compiled spec.

casperisfine: I'd rather not, because it's not guaranteed a gem will properly build with `gem build`, it might require previous step, like `ci-queue` does: https://github.com/Shopify/ci-queue/blob/0a884b216199e12f769292c17b2ad7ab16186c6a/ruby/Rakefile#L26

However that whole FakeSpec is a bit over the top, I cargo culted it from the previous script, but can't remember a justification for it.

I replaced it by a simple `Gem::Specification.load`. SHould be much cleaner now.

Issue #366, opened 2015-05-21
generate deploy keys per stack


dwradcliffe:
It would be nice if shipit generated a deploy key for each stack and added it to GitHub automatically. (Similar to Circle CI)

Not sure how the ssh part of this would work but it would make setup much easier.

Issue #360, opened 2015-05-20
Fetch public repos via HTTPS


byroot:
Public repos an be fetched by anyone, but still requires the SSH key of a github account.

For people deploying only public repos, using HTTPS would simplify things.

It mean adding a public column, and fetching that info.

cc @dwradcliffe

dwradcliffe:
If we do #366 we might not need this.

gmalette:
I think this is a good idea. It would allow deploying projects that the Shipit owner does not own, too. However, we would not be able to setup webhooks in that case, so we'd probably need something that automatically polls for status updates and such

casperisfine:
Some experimentation I did today.

Now that we fully converted to a GitHub App, we can actually use the App token for git operations, e.g: `git clone 'https://git:<token>@github.com/org/repo.git'`.

But off course we don't want to specify it like this, because it would be displayed in the command output and stored in the `.git/config`. So it would be both leaked, and would break once the token expires.

However you can tell git to ask the password to another program with the `GIT_ASKPASS` env var, which would allow to securely pass the token to git.

So thanks to this, I think we can 100% migrate to `https` cloning, which means we'd no longer need a real github user with SSH keys & all that. 

I'll work on this tomorrow, cc @DazWorrall, @wvanbergen

wvanbergen:
Neat, this is cool.

DazWorrall:
For bonus points: are the app tokens short lived? Currently we can completely hide the private key from deploy scripts, whereas `GIT_ASKPASS` will provide a mechanism to expose credentials. If they're time limited then that would greatly mitigate the risk of a leaked token.

wvanbergen:
When using a Github app, the tokens are valid for one hour.

PR #803, opened 2018-06-15
Use HTTPS protocol for git operations

casperisfine:
Fix: https://github.com/Shopify/shipit-engine/issues/360
Closes: https://github.com/Shopify/shipit-engine/issues/366

  - Public repos will be cloned without credentials
  - Private ones will be cloned and fetched using the GitHub App short-lived token
  - It means you can push tags back etc (ref: https://github.com/Shopify/shipit-engine/pull/801)
  - This makes the Shipit setup much simpler in most cases. You won't have to provide proper SSH access to github with a bot user etc.

### DIFF
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -292,7 +292,7 @@ module Shipit
     end
 
     def repo_git_url
-      "git@#{Shipit.github.domain}:#{repo_owner}/#{repo_name}.git"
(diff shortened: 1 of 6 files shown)


casperisfine:
> Does this have any implications for deploy scripts that run bundle install on the repo, potentially with git-based gems, as part of their deploy procedure?

It would have if we remove the SSH access from our installation, but it's orthogonal from this PR.

> Are there any docs that need updating to reflect the simplified setup?

No, but mostly because the setup docs are terrible :/

PR #806, opened 2018-06-25
Allow DeploySpec#task_discovery to be extensible

bbuchalter:
This change allows developers append to task lists without losing the tasks
defined by `KubernetesDiscovery`. The test case demonstrates how this can
be done, setting a new pattern for developers to follow.

### DIFF
--- a/app/models/shipit/deploy_spec/kubernetes_discovery.rb
+++ b/app/models/shipit/deploy_spec/kubernetes_discovery.rb
@@ -17,7 +17,7 @@ module Shipit
               'description' => "Simulates a rollout of Kubernetes deployments by using kubernetes-restart utility",
               'steps' => [kubernetes_restart_cmd],
             },
-          }
+          }.merge!(super)
         else
           super
         end
--- a/test/models/deploy_spec_test.rb
+++ b/test/models/deploy_spec_test.rb
@@ -350,9 +350,59 @@ module Shipit
       assert_equal ['foo'], definition.steps
     end
 
-    test "#task_definitions returns kubernetes commands as well as comands from the config" do
+    test "#task_definitions will always have definitions from shipit.yml take precedence over other modules" do
+      # Create a subclass of `FileSystem` which we can include another module in without
+      # affecting any other tests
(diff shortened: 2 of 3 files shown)


test/models/deploy_spec_test.rb
```diff
@@ -350,17 +350,45 @@ class DeploySpecTest < ActiveSupport::TestCase
...
           'context' => 'bar',
           'timeout' => '20m',
         },
       )
       tasks = @spec.task_definitions
-      assert_equal 2, tasks.size
+
+      # Assert we get tasks from all three sources: config, shipit-engine defined modules, and
+      # "third party" modules
+      assert_same_elements %w(restart config_task module_task), tasks.map(&:id)
```
casperisfine: Instead of adding a test helper:

```ruby
assert_equal %w(restart config_task module_task).sort, tasks.map(&:id).sort
```

app/models/shipit/deploy_spec/kubernetes_discovery.rb
```diff
@@ -16,8 +16,8 @@ def discover_task_definitions
               'action' => "Restart application",
               'description' => "Simulates a rollout of Kubernetes deployments by using kubernetes-restart utility",
               'steps' => [kubernetes_restart_cmd],
-            },
-          }
+            }
+          }.merge!(super)
```
casperisfine: I'd like a test that make sure that explicitly defined tasks inside the `shipit.yml` have precedence.

bbuchalter:
The failure seems to be a false positive:
https://travis-ci.org/Shopify/shipit-engine/builds/396508846?utm_source=github_status&utm_medium=notification

```
Errno::EACCES: Permission denied @ rb_sysopen -
/home/travis/build/Shopify/shipit-engine/vendor/bundle/ruby/2.4.0/cache/nokogiri-1.8.3.gem
An error occurred while installing nokogiri (1.8.3), and Bundler
cannot continue.
```

Is there a way I can trigger a retry?

casperisfine:
I triggered it.

PR #810, opened 2018-08-02
Stop storing hook deliveries in the database

casperisfine:
Based on production usage, it's clear that this method of dealing with webhooks is terribly inefficient.

Our production MySQL spend 80% of it's time on the `deliveries` table, even though we only have 3 registered webhooks.

capture d ecran 2018-08-02 a 17 24 44

There is probably a way to fix the contention and perf issue with MySQL, but ultimately besides some debugging capability once every blue moon, the `deliveries` table really isn't that useful.

This PR simply stop using it, and put all the hook attributes in the job argument, and rely on the job queue implementation for persistence.

Note that the old code is not removed yet for smoother rollout (in flight jobs), and will be removed in a future version.

### DIFF
--- a/app/jobs/shipit/deliver_hook_job.rb
+++ b/app/jobs/shipit/deliver_hook_job.rb
@@ -3,6 +3,7 @@ module Shipit
     queue_as :hooks
 
     def perform(delivery)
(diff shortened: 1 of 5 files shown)


casperisfine:
Tested this branch in production, it basically makes the database idle in comparison.

PR #811, opened 2018-08-20
Reword section to clarify the list contains PR and not commits

d1egoaz:


d1egoaz:
It doesn't seem there are tests affected by this change.

I couldn't run tests locally, I keep getting this error
```
Migrations are pending. To resolve this issue, run:

        bin/rails db:migrate RAILS_ENV=test
Coverage report generated for MiniTest to /Users/diegoalvarez/src/github.com/Shopify/shipit-engine/coverage. 1921 / 3820 LOC (50.29%) covered.

```

casperisfine:
> I keep getting this error

Sometimes sqlite3 cause this kind of issues. You can nuke the db with `bin/bootstrap`

app/views/shipit/stacks/show.html.erb
```diff
@@ -6,7 +6,7 @@
 <div class="wrapper">
   <section>
     <header class="section-header">
-      <h2>Undeployed Commits</h2>
+      <h2>Undeployed PRs</h2>
```
casperisfine: Its is actually commits. Sure most of the time you'll see merge commits from PRs, and they'll be deployed like this. But if someone push straight on master it won't be a PR.

d1egoaz: `Undeployed PRs/Commits` kind of :trollface: but not joking 🤷‍♂️

rafaelfranca: A PR is a commit. Can you explain why you think we need to make distinction here?

d1egoaz: It's about consistency, as this page is using PRs and Commits as they are the same thing, and IMHO they are different.

Most of the time a PR has a single commit, but you can find a PR containing multiple commits.

The `Undeployed Commits` list talks about PR names and not about commits. 
image

If you go into the PRs you'll find the commit messages are different:

image

image


That said, Shopify core is using the merge PR action, that will create an extra commit, and this seems it might be what this is actually talking about.

On the above case a `PR is a commit` **if and only if** the merge PR action is used.

d1egoaz:
the page talks about the merge commits.

PR #819, opened 2018-10-16
Prefer stacks with merge queue enabled in MergeStatusController

DazWorrall:
These stacks will see more utility from HCTW, but in the case where multiple stacks point to the same repo + branch, they may not be displayed by the plugin.

### DIFF
--- a/app/controllers/shipit/merge_status_controller.rb
+++ b/app/controllers/shipit/merge_status_controller.rb
@@ -60,7 +60,7 @@ module Shipit
       @stack ||= if params[:stack_id]
         Stack.from_param!(params[:stack_id])
       else
-        scope = Stack.order(id: :asc).where(
+        scope = Stack.order(merge_queue_enabled: :desc, id: :asc).where(
           repo_owner: referrer_parser.repo_owner,
           repo_name: referrer_parser.repo_name,
         )
--- a/test/controllers/merge_status_controller_test.rb
+++ b/test/controllers/merge_status_controller_test.rb
@@ -33,5 +33,20 @@ module Shipit
       assert_response :ok
       assert_predicate response.body, :blank?
     end
+
+    test "GET show prefers stacks with merge_queue_enabled" do
+      existing = shipit_stacks(:shipit)
+      Shipit::Stack.create(
+        repo_owner: existing.repo_owner,
+        repo_name: existing.repo_name,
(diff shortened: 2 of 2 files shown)


app/controllers/shipit/merge_status_controller.rb
```diff
@@ -60,7 +60,7 @@ def stack
       @stack ||= if params[:stack_id]
         Stack.from_param!(params[:stack_id])
       else
-        scope = Stack.order(id: :asc).where(
+        scope = Stack.order(merge_queue_enabled: :desc, id: :asc).where(
```
casperisfine: Is orderring on boolean standard SQL or is it a MySQLism ?

test/controllers/merge_status_controller_test.rb
```diff
@@ -33,5 +33,21 @@ class MergeStatusControllerTest < ActionController::TestCase
...
+      existing = shipit_stacks(:shipit)
+      Shipit::Stack.create(
+        repo_owner: existing.repo_owner,
+        repo_name: existing.repo_name,
+        environment: 'foo',
+        branch: existing.branch,
+        merge_queue_enabled: true,
+      )
+      existing.merge_queue_enabled = false
+      existing.save
```
casperisfine: 2 things:

  - You can do both using `update(merge_queue_enabled: false)`
  - In test suites I recommend to always use the bang versions `save! / update!`, this way you won't be tripped by a silent failure.

app/controllers/shipit/merge_status_controller.rb
casperisfine: I suppose CI will tell us.

DazWorrall: Good question. Reading about this, SQLite and MySQL store bools as ints, but Postgres doesn't, and when sorting ascending [true comes first](https://code.djangoproject.com/ticket/19726) (this bug is old, but looks like the facts are still correct. I havent tested myself). We chatted IRL and decided to do the sorting in Ruby and try the branch in prod - this endpoint is hit a lot and needs to be performant, but at Shopify the number of stacks configured with the same branch is very small, so this should be fine.

DazWorrall: Hmm. CI _did_ pass 🤔

PR #820, opened 2018-10-16
Show 0 changes if there are no additions or deletions

pawandubey:
In the UI, if a commit contains only additions or only deletions, no diff stats are shown.

E.g the second commit here:

screen shot 2018-10-16 at 4 01 06 pm

This PR fixes this visual (imo) bug.

**Why change in the UI instead of changing `commit.additions` to return `0` by default?**
I am unsure of the semantic meaning behind `nil` additions/deletions in the domain model, so I find this the easiest way to fix things. If it's acceptable to default additions/deletions to `0`, then I'm happy to update the PR.


pawandubey:
@casperisfine Thanks for the context about nullability. I have updated it with your suggestion.

PR #836, opened 2018-11-06
Make CheckRuns mark CI as enabled for a stack like Statuses do already

airhorns:
Right now, if you have only Github Checks on a stack, you get the "Heads Up your stack is not configured for CI banner", even if there are checks being sucked in and displayed with commits. Checks should mark CI as enabled just the same methinks.


casperisfine:
Merged as https://github.com/Shopify/shipit-engine/commit/a2d32f8cdcffa30f032cc4dc189c707a3e5e79b0

PR #838, opened 2018-11-08
Rely on Rails to bump updated_at when touching a continuously delivered stack

airhorns:
We were seeing a lot of these invalid statements sent to Heroku Postgres:

```
PG::SyntaxError: ERROR:  multiple assignments to same column "updated_at"
: UPDATE "stacks" SET "updated_at" = $1, "continuous_delivery_delayed_since" = $2, "updated_at" = $3 WHERE "stacks"."id" = $4
```

originating from `Stack#continuous_delivery_delayed!` . Turns out `ActiveRecord::Base#touch` already touches the `updated_at` column so there's no need to touch it explicitly. I suspect that Shopify didn't hit this issue because they're running MySQL in production for Shipit, not Postgres.

### DIFF
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -141,7 +141,7 @@ module Shipit
     end
 
     def continuous_delivery_delayed!
-      touch(:continuous_delivery_delayed_since, :updated_at) unless continuous_delivery_delayed?
+      touch(:continuous_delivery_delayed_since) unless continuous_delivery_delayed?
     end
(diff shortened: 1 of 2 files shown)


casperisfine:
Merged as https://github.com/Shopify/shipit-engine/commit/831a6652bbd6bcfb7e439640818b05fe6f5d0c32

airhorns:
Oops sorry bout the test fail, thanks Jean!

PR #839, opened 2018-11-13
Wait a bit longer than 0 seconds for the lock when checking continuous delivery

airhorns:
We were seeing a lot of `Redis::Lock::LockTimeout` exceptions from `Shipit::ContinuousDeliveryJob` on our rather modestly sized Shipit deploy on Heroku. I believe that because we have lots of github hooks arriving at the same time from many different CircleCI checks there is a bit of a stampede of `ContinuousDeliveryJobs` kicked off all at once. It seems correct to me they should execute serially, but, it doesn't seem correct that they should immediately exception with a `LockTimeout` if there is another one running already. The default `lock_timeout` for a `Shipit::BackgroundJob::Unique` is `0`, so, that was the behaviour before this change.

Instead, let's wait a small amount of time to see if the job can get the lock. The jobs are quick so I don't think we need to wait super long, so I chose 3 seconds.

A big question for me would be why doesn't Shopify hit this?

### DIFF
--- a/app/jobs/shipit/continuous_delivery_job.rb
+++ b/app/jobs/shipit/continuous_delivery_job.rb
@@ -2,6 +2,7 @@ module Shipit
   class ContinuousDeliveryJob < BackgroundJob
     include BackgroundJob::Unique
 
+    self.lock_timeout = 3 # seconds
     queue_as :default
 
     def perform(stack)


casperisfine:
> A big question for me would be why doesn't Shopify hit this?

I wondered as well, so I looked:

```ruby
# cat cat config/initializers/sidekiq.rb
    Bugsnag.configure do |config|
      config.ignore_classes += %w(
        Redis::Lock::LockTimeout
      )
    end
```

So we definitely do...

But the problem is not really to get the lock or not, the point of letting this exception raise is actually to have the job being retried rather than stuck waiting for a lock. It's much better if it's sitting on the queue rather than occupying a thread/process.

It's just that this exception should be properly handled in `shipit-engine` / ActiveJob, and more generally `BackgroundJob::Unique` would need to be improved because it's a bit too hacky right now.

airhorns:
Hmmm, I see, that makes sense, no need to have a worker occupied spinning on a lock if it can just go back in the queue. Ok, closing! I think it'd be great if those exceptions didn't bubble up though yes, but it does indeed seem like a bigger refactor.

PR #842, opened 2018-11-19
Upgrade faraday to 0.12 series for kubernetes-deploy > 0.6.0 compatibility

airhorns:
Before this change shipit-engine was locked to an older version of faraday which prevents dependency compatibility with `kubernetes-deploy`, which is a supported feature. `kubernetes-deploy` was held back to `0.6.0` which is real old, `0.22.0` was released a week or two ago. This upgrades `shipit-engine`'s faraday dependency to the same one `kubernetes-deploy` has so that `kubernetes-deploy`@`0.22.0` can be in the same bundle! Woop woop!

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -8,7 +8,7 @@ PATH
       autoprefixer-rails (~> 6.4.1)
       coffee-rails (~> 4.2)
       explicit-parameters (~> 0.3.0)
-      faraday (~> 0.9.1)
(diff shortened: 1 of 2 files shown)


casperisfine:
Thanks!

> so that kubernetes-deploy@0.22.0 can be in the same bundle!

Note that Shipit shells out with `Bundler.clean_env`, so that might not work as you expect.

PR #844, opened 2018-11-22
Return last deploy revision for stacks through api

JackTLi:
Part of Shopify's canary analysis tooling relies on getting the latest deploy revision (whether successful or not) to target for analysis. While we get this information from webhooks, the ability to pull this information can make it more robust in cases such as dropped webhooks.


app/serializers/shipit/stack_serializer.rb
```diff
@@ -5,7 +5,7 @@ class StackSerializer < ActiveModel::Serializer
     has_one :lock_author
     attributes :id, :repo_owner, :repo_name, :environment, :html_url, :url, :tasks_url, :deploy_url, :pull_requests_url,
                :deploy_spec, :undeployed_commits_count, :is_locked, :lock_reason, :continuous_deployment, :created_at,
-               :updated_at, :locked_since, :last_deployed_at, :branch, :merge_queue_enabled
+               :updated_at, :locked_since, :last_deployed_at, :branch, :merge_queue_enabled, :last_deploy_revision
```
byroot: IMO it would be better to have `last_deployed_commit`, and let the commit serializer do the job so that you have `stack.last_deployed_commit.sha`

app/models/shipit/stack.rb
```diff
@@ -255,6 +254,10 @@ def undeployed_commits
       scope.map.with_index { |c, i| UndeployedCommit.new(c, i) }.reverse
     end
 
+    def last_deploy
```
byroot: There is already a `last_deployed_commit` method. Basically all you need is to expose it in the serializer.

The rest of the diff shouldn't be required.

byroot: Oh, nvm, in your case you want the last deploy even if it's not successful.

In this case I think it would be better to add a `/tasks` endpoint and grab the last one.

There only so much stuff that can be put into the StackSerializer.

JackTLi: That works for me. What's considered a Task in shipit? just `Deploy`/`Rollback`?

byroot: `Task > Deploy > Rollback`

Straight `Task` are the custom commands you can trigger, but that don't change the deployed revision. e.g. failover, restart & such.

JackTLi:
@byroot simplified drastically 😄

app/controllers/shipit/api/tasks_controller.rb
```diff
@@ -12,6 +12,10 @@ def show
         render_resource stack.tasks.find(params[:id])
       end
 
+      def latest
+        render_resource stack.last_active_task
```
casperisfine: Oh, I didn't remember we had a `/tasks` endpoint already. In this case do we even need a change?

You could hit `/index` with `?page_size=1`

JackTLi: oh hm lemme see. would we be able to trust the ordering that this will always return the latest?

casperisfine: Yes, it should be in descending order by default. If it isn't it's easily fixable.

Be careful though as it will return tasks, deploys and rollback altogether. So you might want to take more than 1 and filter out tasks.

JackTLi: hm I don't really like that last part since I don't want to keep hitting the api and filtering the data until I get what I want. thinking about exposing a `latest_deploy` to just return the latest deploy/rollback task. thoughts?

casperisfine: Then I'd suggest to add a `/deploys` endpoint that filter tasks out.

It's much more versatile that a `/tasks/latest` endpoint and much more homogenous with the rest of the API.

JackTLi: works for me 👍

casperisfine: There is a `Stack#deploys_and_rollbacks` relation. So it's basically just `render_resources stacl.deploys_and_rollbacks` plus the route.

JackTLi:
@casperisfine another look?

PR #845, opened 2018-11-23
Introduce validating and faulty statuses for deploys

casperisfine:
Two new `Deploy#status` are introduced for canary stacks: `validating` and `faulty`.

When the deploy complete, instead of transitioning directly to `success` it goes into `validating`, and then whenever the release is marked as `faulty` or `healthy` it transition to `faulty` or `success` respectively.

This should impact stacks that don't use canaries in any way.
 

I'm not quite sure which symbol to give to the `faulty` status:

I'm thinking either a yellow cross:
capture d ecran 2018-11-23 a 15 23 29

Or a yellow checkmark:
capture d ecran 2018-11-23 a 15 42 32


DazWorrall:
Team yellow cross for me I think, but it's tricky, I can see the case for the check mark. The _deploy_ succeeded but the _release_ is faulty.

app/jobs/shipit/mark_deploy_healty_job.rb
```diff
@@ -0,0 +1,13 @@
+module Shipit
+  class MarkDeployHealthyJob < BackgroundJob
+    include BackgroundJob::Unique
+
+    queue_as :default
+
+    def perform(deploy)
+      return unless deploy.validating?
+
+      deploy.report_healthy!(description: "No issue were signaled after #{deploy.stack.release_status_delay}")
```
DazWorrall: grammar nit: `no issue was signalled` or `no issues were signalled`

app/models/shipit/deploy.rb
```diff
@@ -66,13 +66,14 @@ def trigger_rollback(user = AnonymousUser.new, env: nil, force: false)
     end
 
     # Rolls the stack back to the **previous** deploy
-    def trigger_revert
+    def trigger_revert(force: false)
```
DazWorrall: Doesn't need to be fixed here, but `trigger_rollback` would be clearer here, especially now Shipit deals with PRs directly.

casperisfine: Yes, unfortunately `trigger_rollback` is already used for something else :/

app/jobs/shipit/mark_deploy_healty_job.rb
casperisfine: Thanks. I fixed it.

PR #849, opened 2018-11-30
Write Task#ended_at when entering the `validating` status

casperisfine:
That property is used for `deploy.delay`. 

We should consider the deploy `ended` whenever it's no longer `running`.

### DIFF
--- a/app/models/shipit/task.rb
+++ b/app/models/shipit/task.rb
@@ -57,7 +57,7 @@ module Shipit
         task.started_at ||= Time.now.utc
       end
 
-      before_transition any => %i(success failed error timedout) do |task|
+      before_transition any => %i(success failed error timedout validating) do |task|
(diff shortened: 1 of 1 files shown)


app/models/shipit/task.rb
```diff
@@ -57,7 +57,7 @@ def current
         task.started_at ||= Time.now.utc
       end
 
-      before_transition any => %i(success failed error timedout) do |task|
+      before_transition any => %i(success failed error timedout validating) do |task|
```
doodzik: is there a reason why `aborted` isn't part of this transition?

casperisfine: Not all the status you list make sense.

However it would probably be better to take this from the other side:

`after_transition running => any` would be much better and won't break if we introduce more statuses.

PR #855, opened 2018-12-11
Add undeployed commits endpoint

JackTLi:
Add an api endpoint to be able to look up the undeployed commits for a given stack.

@grcooper is looking into building some tooling to be able to send messages to people who are waiting for their commits to be deployed. This will help enable that.

### DIFF
--- a/app/controllers/shipit/api/commits_controller.rb
+++ b/app/controllers/shipit/api/commits_controller.rb
@@ -6,6 +6,12 @@ module Shipit
       def index
         render_resources stack.commits.reachable.includes(:statuses)
       end
+
+      def undeployed
+        stack.undeployed_commits do |undeployed_commits|
+          render_resources undeployed_commits
+        end
+      end
     end
   end
 end
--- a/app/controllers/shipit/api/commits_controller.rb
+++ b/app/controllers/shipit/api/commits_controller.rb
@@ -4,13 +4,12 @@ module Shipit
       require_permission :read, :stack
 
       def index
-        render_resources stack.commits.reachable.includes(:statuses)
-      end
-
-      def undeployed
-        stack.undeployed_commits do |undeployed_commits|
-          render_resources undeployed_commits
+        commits = stack.commits.reachable.includes(:statuses)
+        if params[:undeployed]
+          commits = commits.newer_than(stack.last_deployed_commit)
         end
+
+        render_resources commits
       end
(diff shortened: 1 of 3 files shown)


grcooper:
Related PR: https://github.com/Shopify/spy/pull/4593

config/routes.rb
```diff
@@ -25,7 +25,9 @@
         resource :output, only: :show
       end
       resources :deploys, only: %i(index create)
-      resources :commits, only: %i(index)
+      resources :commits, only: %i(index) do
+        get :undeployed, on: :collection
```
byroot: IMO it would male more sense as a query param: `/commits?undeployed=1`

test/controllers/api/commits_controller_test.rb
```diff
@@ -15,6 +15,16 @@ class CommitsControllerTest < ActionController::TestCase
         assert_response :ok
         assert_json '0.sha', commit.sha
       end
+
+      test "#undeployed returns a list of undeployed commits" do
+        commits = @stack.undeployed_commits.map(&:sha)
```
byroot: nitpick: you can use `.pluck(:sha)` for that. It doesn't matter much because it's a unit test but...

app/controllers/shipit/api/commits_controller.rb
```diff
@@ -6,6 +6,12 @@ class CommitsController < BaseController
       def index
         render_resources stack.commits.reachable.includes(:statuses)
       end
+
+      def undeployed
+        stack.undeployed_commits do |undeployed_commits|
+          render_resources undeployed_commits
```
byroot: That block API isn't really meant to be used this way. You will get commits in reverse order etc.

You might be better to just get the query you want: https://github.com/Shopify/shipit-engine/blob/20f7c28874fc9b5e29cb3ca6efee2471048d5f0b/app/models/shipit/stack.rb#L255

```ruby
stack.commits.reachable.newer_than(stack.last_deployed_commit).includes(:statuses)
```

So basically the same thing already in `index` but with one more criteria.

grcooper: If we are already doing this in index then we probably should change it to add a param to index for deployed vs not deployed as you mentioned above.

JackTLi:
thanks @casperisfine , pr updated 😄

PR #864, opened 2019-01-13
Show an error message for the title when it would otherwise result in a 500

### LINKED ISSUE #862: ActionView::Template::Error in shipit/stacks#show when a bad task is specified in shipit.yml
in shipit.yml, it's possible to specify a task badly: 
```
tasks:
  reindex:
    action: "Reindex elastic search models"
    description: "Sometimes needed if data changes don't get propogated to Elastic Search."
    title: "Running Procfile command: %{COMMAND}"
    steps:
      - run service-production tier3 $reindex
    variables:
      - name: reindex
        title: "Reindex information for all Elastic Search Models"
```

Note the title mistakenly uses `COMMAND` instead of the intended `reindex` variable.

When this happens, the commit is deployed and the task is saved as a `TaskDefinition` in a `Task`. When the stack#show controller action is called, the `TaskDefinition` title is interpolated and an exception occurs, giving the user a 500. It's then not possible to use the interface to roll back the change.

I fixed the `TaskDefinition` manually, after unsuccessfully manually rolling back the change, but we should prevent this from happening in the first place. 

I'll attempt a fix to just catch the exception and render an error message for the broken task. Ideally we'd validate these tasks and fail CI when an invalid one is specified, but given the frequency of this happening, I don't think it's worth the extra effort.

mcgain:
Fixes #862.

I opted to do a very simple error message here instead of something more fancy since it doesn't seem worth the added complexity. I could also have put this fix in the TaskDefinition and delegated to description again, but that would just hide the problem.

### DIFF
--- a/app/models/shipit/task_definition.rb
+++ b/app/models/shipit/task_definition.rb
@@ -35,6 +35,8 @@ module Shipit
       else
         action
       end
+    rescue KeyError
+      "This task (title: #{@title}) cannot be shown due to an incorrect variable name. Check your shipit.yml file"
     end
 
     def allow_concurrency?
--- a/test/fixtures/shipit/tasks.yml
+++ b/test/fixtures/shipit/tasks.yml
@@ -209,3 +209,30 @@ canaries_running:
   created_at: <%= (60 - 7).minutes.ago.to_s(:db) %>
   started_at: <%= (60 - 7).minutes.ago.to_s(:db) %>
   ended_at: <%= (60 - 10).minutes.ago.to_s(:db) %>
+
+shipit_with_title_parsing_issue:
+  id: 105
+  user: walrus
+  since_commit_id: 2 # second
+  until_commit_id: 2 # second
+  type: Shipit::Task
+  stack: shipit
+  status: success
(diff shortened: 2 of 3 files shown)


app/models/shipit/task.rb
```diff
@@ -228,6 +228,8 @@ def supports_rollback?
 
     def title
       definition.render_title(env)
+    rescue KeyError
+      'This task cannot be shown due to an incorrect variable name. Check your shipit.yml file'
```
doodzik: \nit I would include the name of the task instead of referring to `this` (if that is possible of course)

casperisfine: I think I would have rescued in `render_title` itself.

mcgain:
I pushed the implementation lower and changed the error message as suggested. 

Catching an exception in the controller would halt the execution though, so other tasks won't show. Avoiding that will require a lot more code/complexity than this error warrants I think, so I'll keep it as is.

mcgain:
note, that deploy was a mistake. I didn't cut a new release so the deploy did not push to rubygems

Issue #841, opened 2018-11-16
Global/System Hook System


rbroemeling:
Hi Shopify,

We want to modify our local ShipIt install to track deployment metrics (how long are deploys taking? how many deploys are happening? Etc.) We track all metrics via DataDog, so what we want to do is report this information directly to DataDog somehow. We have a local dogstatsd instance that is available, so it would be straightforward to do this if we could figure out how to execute some code at specific points in the ShipIt process ("deploy begun", and "deploy complete", for instance).

I realize that we can do this with `post`/`pre` overrides in the ShipIt configuration of each repository, but that's pretty far from ideal because it means that we need to set that up as a part of every single repository, and that someone has to notice if it is missed, etc.

We'd much rather handle this at the system/ShipIt level, where every stack gets it by default, without any developer pain or requirement to ask for it or implement it.

I am further aware that ShipIt has web hook support (which is waiting for a UI from #364), and I briefly looked to see if it was flexible enough to actually post metrics to DataDog's HTTP metric submission API directly. It doesn't look like there is any ability to template or control the body of the sent data, though, so that doesn't appear viable.

So we're left in a situation where we need to implement an interstitial web daemon that just waits for hooks from ShipIt and then "translates" them into DataDog stats (probably submitting them to the local dogstatsd instance). That isn't really a desirable infrastructure from our point-of-view, however.

ShipIt doesn't currently seem to have a global hook script system (i.e., for each event run all of the executable scripts in a given directory).

This seems like it would be a significant win in terms of flexibility (it'd give a huge amount of flexibility to ShipIt, as people would be able to customize behaviours for the entire ShipIt system by just dropping scripts into this set of directories) -- is this a feature request that the ShipIt team at Shopify would be interested in, or is it undesirable for some reason?

Example: what I'm looking for is something similar to the Buildkite Agent hooks system (https://buildkite.com/docs/agent/v3/hooks).

rbroemeling:
I've been looking at a possible implementation of this.

One possible way would be to add a `LocalHookRunner` background job that takes an event, maps the event to a code directory, and then executes all of the scripts inside that directory in a deterministic order. Then, in `Shipit::Task` around line 285, add emission of a `LocalHookRunner` immediately after `Hook` is emitted, possibly with a guard to only emit the job if-and-only-if the directory exists (to prevent emitting a bunch of "no-op" background jobs).

Does that mechanism make sense? I'm willing to take on some of this work and to prepare a PR, if it would be welcomed.

[...]

PR #865, opened 2019-01-15
Allow the enclosing application to listen to hooks

mcgain:
I want to send hooks using an internal format that's not appropriate to support in this repo, so it occurred to me that it would be nice for the main app to be able to subscribe to hooks itself, instead of having to maintain another app to receive hooks and then send them on. This is the simplest way I can come up with to do that, but I wanted to push this up for feedback before I add tests, since someone might have a better approach.

Notably, this doesn't have nice features like event or stack filtering built in. 

The main application has to set 
```ruby 
Shipit.internal_hook_receivers = [MyHookReceiver]

module MyHookReceiver
  def self.deliver(event, stack payload)
    #this is the required interface
  end
end
```

### DIFF
--- a/app/models/shipit/hook.rb
+++ b/app/models/shipit/hook.rb
@@ -75,6 +75,13 @@ module Shipit
           stack_id: stack&.id,
           payload: coerce_payload(payload),
         )
+        deliver_internal_hooks(event, stack, payload)
+      end
+
+      def deliver_internal_hooks(event, stack, payload)
+        Shipit.internal_hook_receivers.each do |receiver|
+          receiver.deliver(event, stack, payload)
+        end
       end
 
       def deliver(event, stack_id, payload)
--- a/lib/shipit.rb
+++ b/lib/shipit.rb
@@ -58,6 +58,7 @@ module Shipit
   delegate :table_name_prefix, to: :secrets
 
   attr_accessor :disable_api_authentication, :timeout_exit_codes
+  attr_writer :internal_hook_receivers
 
   self.timeout_exit_codes = [].freeze
 
@@ -181,6 +182,10 @@ module Shipit
 
   def committer_email
     secrets.committer_email.presence || "#{app_name.underscore.dasherize}@#{host}"
+  end
+
+  def internal_hook_receivers
+    @internal_hook_receivers || []
   end
 
   protected
--- a/test/models/hook_test.rb
+++ b/test/models/hook_test.rb
@@ -42,5 +42,34 @@ module Shipit
       @hook.stack_id = 42
       assert @hook.scoped?
     end
+
+    test ".emit schedules an EmitEventJob" do
+      assert_enqueued_jobs(1, only: EmitEventJob) do
+        Hook.emit(:deploy, @stack, 'foo' => 42)
+      end
+    end
+
+    test ".emit calls #deliver on internal hooks" do
+      begin
+        original_receivers = Shipit.internal_hook_receivers
+        FakeReceiver = Module.new
+        FakeReceiver.expects(:deliver).with(:deploy, @stack, 'foo' => 42)
+
+        Shipit.internal_hook_receivers = [FakeReceiver]
+        Hook.emit(:deploy, @stack, 'foo' => 42)
+      ensure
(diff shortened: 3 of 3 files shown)


casperisfine:
Yeah, I'd be fine with such feature, and it has been requested before in https://github.com/Shopify/shipit-engine/issues/841

However, if we're handling this in process, it might be better to pass actual models rather than the serialized version. We could also wonder wether it should be called synchronously, or asynchronously.

app/models/shipit/hook.rb
```diff
@@ -73,13 +73,38 @@ def emit(event, stack, payload)
...
           stack_id: stack&.id,
-          payload: coerce_payload(payload),
+          payload: clean_payload(payload),
         )
       end
 
+      def clean_payload(payload)
+        cleaned_payload = payload.dup
+        payload.each do |key, value|
+          coerced_payload[key] =
```
doodzik: where is the `coerced_payload` defined?

mcgain:
@doodzik and I made an attempt at passing around proper models (well, at least the global ids) but there are a bunch of non-models being serialised (via json at the moment, which cannot be deserialised in a generic way at the other end). I'll figure out a better way to serialize them so they can be properly deserialized. 

I think these hooks should be called asynchronously, one-per-job since that isolates everything nicely.

casperisfine:
> I'm unsure whether I should keep going down the serialization road, or leave this as is. Thoughts?

Maybe the issue is that we try to make these async. 

What would be common usage for this kind of hook points? I'm thinking statsd metrics, and for those enqueuing a job is totally overkill.

If you need a job you can always manually enqueue one from the hook.

mcgain:
I imagine custom metrics are the primary use case. It's certainly mine and #841's

Firing the hooks synchronously will result in something as simple as that first commit I put up. I'll make the changes tomorrow.

mcgain:
Ok, I shifted the behavior right up to `emit` so internal hooks will be fired from wherever the event occurs, not in any job. This allows all the objects to be passed properly.

app/models/shipit/hook.rb
```diff
@@ -75,6 +75,13 @@ def emit(event, stack, payload)
           stack_id: stack&.id,
           payload: coerce_payload(payload),
         )
+        deliver_internal_hooks(event, stack, payload)
+      end
+
+      def deliver_internal_hooks(event, stack, payload)
+        (Shipit.internal_hook_receivers || []).each do |receiver|
+          receiver.deliver(event, stack, payload)
```
DazWorrall: What sort of error handling do we need here? Catch everything, log it, and carry on?

casperisfine: You can see it two ways. Either you consider theses internal hooks are part of the app and then if they are borken, it's like your app was broken and you'll fix it soon enough.

Or you can consider them as some kind of third party, and prevent them from breaking the app.

If you do the later, then yeah you'd want to rescue `StandardError`, and log it (and provide some kind of hook point to plug custom exception reporters).

Honestly I think solution 1 is ok, but I wouldn't mind 2.

app/models/shipit/hook.rb
```diff
@@ -75,6 +75,13 @@ def emit(event, stack, payload)
           stack_id: stack&.id,
           payload: coerce_payload(payload),
         )
+        deliver_internal_hooks(event, stack, payload)
+      end
+
+      def deliver_internal_hooks(event, stack, payload)
+        (Shipit.internal_hook_receivers || []).each do |receiver|
```
doodzik: I think it makes more sense for the internal hook receivers to return `[]` by default rather than checking for nil here.

mcgain: :+1:

app/models/shipit/hook.rb
mcgain: I wondered about this too. I decided not to add the error handling because logging is so often overlooked. Main app code can quite easily rescue exceptions from their `deliver` methods, but that's a decision they can make at the time. If it were likely that multiple teams would be adding a bunch of sophisticated hooks, then more error handling is justified, but I don't see the need for that right now.

PR #867, opened 2019-01-24
set ended_at when transitioning to an end statuses

doodzik:
This is a follow up PR on https://github.com/Shopify/shipit-engine/pull/849#discussion_r

Some statuses are missing the `ended_at` time in the internal hooks so we are adding them here. #865 

@casperisfine can you double check if the statuses that the `END_STATUSES` make sense?

### DIFF
--- a/app/models/shipit/task.rb
+++ b/app/models/shipit/task.rb
@@ -57,7 +57,7 @@ module Shipit
         task.started_at ||= Time.now.utc
       end
 
-      before_transition any => %i(success failed error timedout validating) do |task|
+      before_transition running: any do |task|
         task.ended_at ||= Time.now.utc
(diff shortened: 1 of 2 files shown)


app/models/shipit/task.rb
```diff
@@ -57,7 +58,7 @@ def current
         task.started_at ||= Time.now.utc
       end
 
-      before_transition any => %i(success failed error timedout validating) do |task|
+      before_transition any => END_STATUSES.map(&:to_sym) do |task|
```
byroot: Why not https://github.com/Shopify/shipit-engine/pull/849#discussion_r250537738 ?

It's much simpler, and much more forward compatible.

PR #868, opened 2019-02-04
Add descriptive title to the emergency mode button

### LINKED ISSUE #863: Explain Enable Emergency Mode button in UI
I've been working with Shipit for years and generally understand it as well as one needs to, but despite this, the `Enable Emergency Mode` button is still scary.

I don't use it often enough to remember exactly what it does, it's unclear whether it will produce a confirm dialogue explaining what will happen, and worst of all, these realizations generally happen during at least somewhat stressful scenarios. 

I think something as simple as the tooltip used on the commit lock button (screenshot below) could adequately address the issue.

screen shot 2019-01-11 at 13 25 28

I could contribute this if somebody could provide a succinct explanation for me to use.

gavinsmith:
Fixes #863 by adding a `title` attribute to the Emergency Mode button (link). I did try a tooltip, but the description is a bit too long to work well in that UI. This is fine.

I could not 🎩  this because I'm unclear on how to render stacks#index in development. I would appreciate it if somebody who has things running could check it out or show me the way to get things running myself.

### DIFF
--- a/app/views/shipit/stacks/show.html.erb
+++ b/app/views/shipit/stacks/show.html.erb
@@ -10,9 +10,9 @@
 
       <div class="commit-list-actions">
         <% if params[:force] %>
-          <%= link_to 'Exit emergency mode', stack_path(@stack) %>
+          <%= link_to t('emergency_mode.disable'), stack_path(@stack) %>
         <% else %>
-          <%= link_to 'Enable emergency mode', stack_path(@stack, force: 1) %>
+          <%= link_to t('emergency_mode.enable'), stack_path(@stack, force: 1), title: t('emergency_mode.enable_description') %>
         <% end %>
       </div>
     </header>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -60,6 +60,10 @@ en:
     create:
       error:
         deploy_in_progress: Rollbacks can't be triggered if a deploy is in progress
+  emergency_mode:
+    enable: Enable emergency mode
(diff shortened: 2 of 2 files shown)


casperisfine:
> I could not 🎩 this because I'm unclear on how to render stacks#index in development.

There's a `dev.yml`. You can simply `dev up`, and then the trick is to disable authentication: `SHIPIT_DISABLE_AUTH=1 dev s`, and then `dev open`.

gavinsmith:
@byroot thanks! I'm missing a few things though.

`Config: Shipit requires a GitHub App to authenticate users and perform API calls`
`Redis: Shipit needs a Redis server. Please configure the Redis URL in the secrets.yml file of your app, under the key redis_url.`
`Host: Shipit needs the host of the application before generating links in background jobs. Add the host name to the secrets.yml file, under the host key.`

casperisfine:
Oh, I forgot. You have to invoke: `dev bootstrap` first

gavinsmith:
got it, thank you! 🎩 ✅

casperisfine:
> I did try a tooltip, but the description is a bit too long to work well in that UI.

Is the rendering broken or something ? I feel like a tooltip would be more consistent with the rest of the UI.

gavinsmith:
It's just a bit big. The title attribute is expected for links where the intent isn't made evident by its contents anyways, it's what I expect every time I hover over it, uncertain of what it does.

screen shot 2019-02-04 at 10 18 11 am

screen shot 2019-02-04 at 10 19 37 am

PR #874, opened 2019-02-20
Create an endpoint to allow statuses to be set through the api

JackTLi:
This PR will add a `create_statuses` endpoint under deploys to allow commit statuses to be set programmatically on deploys. This will allow the automated canary analysis to publish faulty statuses when problems are detected.


config/routes.rb
```diff
@@ -24,7 +24,9 @@
       resources :tasks, only: %i(index show) do
         resource :output, only: :show
       end
-      resources :deploys, only: %i(index create)
+      resources :deploys, only: %i(index create) do
+        post 'create_status'
```
casperisfine: I'm not a fan of that endpoint. I'm no REST extremist, but it's not very consistent with the rest of the API.

A better API IMO would be:

`POST /.../deploys/:id/release_statuses` with `state = healthy / faulty`.

So in term of routes:

```ruby
resources :release_statuses, only: %i(create)
```

JackTLi:
@casperisfine another look? 😁

app/controllers/shipit/api/release_statuses_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+      end
+      def create
+        deploy = stack.deploys_and_rollbacks.find(params[:deploy_id])
+        case params[:status]
+        when 'success'
+          deploy.report_healthy!(user: current_user)
+        when 'failure'
+          deploy.report_faulty!(user: current_user)
+        else
+          param_error!(:status, 'Invalid status')
```
casperisfine: Instead of raising here, you can do:

```ruby
params do
  requires :status, String
  validates :status, inclusion: { in: %w(success failure) }
end
```

This way explicit-parameters will handle validation for you.

JackTLi: Ah cool!

PR #877, opened 2019-03-01
emit lock duration when stack is unlocked

doodzik:
This PR adds support for receiving the start and end time of how long a lock lasted. That can be used to save the data in order to query it later.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,6 +18,7 @@ end
 group :test do
   gem 'spy'
   gem 'mocha'
+  gem 'timecop'
   gem 'simplecov', require: false
 end
 
--- a/Gemfile
+++ b/Gemfile
@@ -18,7 +18,6 @@ end
 group :test do
   gem 'spy'
   gem 'mocha'
-  gem 'timecop'
   gem 'simplecov', require: false
 end
 
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -502,7 +502,12 @@ module Shipit
 
     def emit_lock_hooks
       return unless previous_changes.include?('lock_reason')
-      Hook.emit(:lock, self, locked: locked?, stack: self)
+
+      lock_duration = if previous_changes['lock_reason'].last.blank?
+        {from: previous_changes['locked_since'].first, until: Time.zone.now}
+      end
+
+      Hook.emit(:lock, self, locked: locked?, lock_duration: lock_duration, stack: self)
     end
 
     def emit_added_hooks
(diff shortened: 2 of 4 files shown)


test/models/stacks_test.rb
```diff
@@ -330,16 +330,20 @@ def setup
...
     end
 
     test "unlocking the stack triggers a webhook" do
+      Timecop.freeze
+      time = Time.current
       @stack.update(lock_reason: "Just for fun", lock_author: shipit_users(:walrus))
-      expect_hook(:lock, @stack, locked: false, stack: @stack) do
+      Timecop.travel(time + 5.minutes)
+      expect_hook(:lock, @stack, locked: false, lock_duration: {from: time, until: Time.current}, stack: @stack) do
         @stack.update(lock_reason: nil)
```
doodzik: This test fails as the asserted `lock_duration until` time until isn't equal with the one emitted in the ` def emit_lock_hooks`. It diverges by a couple of milliseconds. Commenting out `Timecop.return` doesn't solve the issue either.

One possibility could be [an issue with mysql](https://stackoverflow.com/a/42975145)

app/models/shipit/stack.rb
```diff
@@ -502,7 +502,12 @@ def schedule_merges_if_necessary
 
     def emit_lock_hooks
       return unless previous_changes.include?('lock_reason')
-      Hook.emit(:lock, self, locked: locked?, stack: self)
+
+      lock_duration = if previous_changes['lock_reason'].last.blank?
+        {from: previous_changes['locked_since'].first, until: Time.zone.now}
```
DazWorrall: The value of this doesn't match the name. Could we emit an integer here (`lock_duration_in_seconds`)? I think that would make the testing easier too (e.g. probably wouldn't need to add `Timecop`).

Gemfile
```diff
@@ -18,6 +18,7 @@ end
 group :test do
   gem 'spy'
   gem 'mocha'
+  gem 'timecop'
```
casperisfine: Instead of adding timecop it would be preferable to use Rails time helpers: https://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html#method-i-travel_to

app/models/shipit/stack.rb
casperisfine: `_in_seconds` is a bit verbose for no reason, so I'd say `lock_duration`, but 👍

Gemfile
doodzik: nice. I wasn't aware that this existed :)

app/models/shipit/stack.rb
doodzik: I wouldn't want to change that as we would loose the start and the end time of the lock. It is better to safe the raw data and do the calculation later on.

DazWorrall: nit: `lock_details` or similar would have been better in that case, the value of this field is not a duration.

doodzik: makes sense. I'll update that tomorrow.

Issue #778, opened 2018-04-05
Document NPM workflows


amrocha:
The current documentation doesn't explain shipit's support for NPM packages/apps. For example, this PR added NPM support but is not mentioned in the README at all: #631

PR #883, opened 2019-03-27
Add some docs about shipit+npm+lerna

iainmcampbell:
This is a start to addressing #778, combined with some recent learnings about Lerna and prereleases.

### DIFF
--- a/docs/npm-lerna.md
+++ b/docs/npm-lerna.md
@@ -1,0 +1,25 @@
+# Publishing to npm
+
+Shipit can be used to publish code to npm.
+If your project has a `package.json`, and doesn’t have any other deploy method set up, Shipit will publish it to npm.
+
+### Pre-releases
+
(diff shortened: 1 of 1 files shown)


docs/npm-lerna.md
```diff
@@ -0,0 +1,21 @@
+# Publishing to npm
+
+Shipit can be used to publish code to npm.
+If your project has a `package.json`, and doesn’t have any other deploy method set up, Shipit will publish it to npm.
+
+[Lerna](https://github.com/lerna/lerna) is a tool for managing projects with multiple packages.
```
TheMallen: We might want to split lerna and npm into their own documents, or have lerna be at the end so it's clear it's it's own thing

iainmcampbell: makes sense 👍 I put the Lerna info at the end.

PR #886, opened 2019-04-03
Add /task/:id to redirect to deploys/tasks/rollbacks

esnunes:
## Problem

At the moment the user needs to know the stack path in order to create a link to a `task`, `deploy` or `rollback`, even with the given resource `id` at hand.

## Solution

Create the endpoint `/task/:id` that with the given resource id, redirects the user to the correct resource url.

## Notes

`Rollback` extends `Deploy` but doesn't have a specific view/url, because of that the code transforms any subclass of `Deploy` to look like a `Deploy` using `ActiveRecord#becomes` method.

### DIFF
--- a/test/controllers/tasks_controller_test.rb
+++ b/test/controllers/tasks_controller_test.rb
@@ -123,5 +123,29 @@ module Shipit
       assert_response :success
       assert_json_keys %w(status output rollback_url)
     end
+
+    test ":lookup returns stack task url if the task is an instance of Task" do
+      @task = shipit_tasks(:shipit_restart)
+
+      get :lookup, params: {id: @task.id}
+
+      assert_redirected_to stack_task_path(@task.stack, @task)
+    end
+
+    test ":lookup returns stack deploy url if the task is an instance of Deploy" do
+      @task = shipit_tasks(:shipit)
+
+      get :lookup, params: {id: @task.id}
+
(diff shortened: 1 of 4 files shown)


test/controllers/tasks_controller_test.rb
```diff
@@ -123,5 +123,29 @@ class TasksControllerTest < ActionController::TestCase
...
+
+    test ":lookup returns stack deploy url if the task is an instance of Deploy" do
+      @task = shipit_tasks(:shipit)
+
+      get :lookup, params: {id: @task.id}
+
+      assert_redirected_to stack_deploy_path(@task.stack, @task)
+    end
+
+    test ":lookup returns stack deploy url if the task is a kind of Deploy" do
```
DazWorrall: nit `kind of Rollback`?

esnunes: actually not, the business logic I want to test is: given the instance from a subclass of `Deploy` 
it should redirect to the deploy resource url, in this example I'm providing a `Rollback` instance.

I can change to `:lookup returns stack deploy url if the task is an instance of a subclass of Deploy`, what do you think?

JackTLi: From the test perspective, we should forget that any subclassing exists. This test is explicitly testing that Rollbacks receive a deploy url, so we should probably label it like that. It's a single case we're testing here, not a generalization 🙂

PR #887, opened 2019-04-04
Flag commits scheduled to next deploy

esnunes:
## Problem

In a stack with continuous deployment enabled, by looking at the list of `Undeployed Commits` it is not clear which of them are going to be grouped and deployed.

## Solution

Based on the config value `deploy.max_commits` flag the commits that are supposed to be grouped and deployed.

## Notes

This PR was created on top of #890 .

Screenshot 2019-04-05 at 15 02 36

### DIFF
--- a/app/models/shipit/undeployed_commit.rb
+++ b/app/models/shipit/undeployed_commit.rb
@@ -35,12 +35,22 @@ module Shipit
     end
 
     def deploy_discouraged?
-      stack.maximum_commits_per_deploy && index >= stack.maximum_commits_per_deploy
+      maximum_commits_per_deploy_reached?
+    end
+
+    def deploy_scheduled?
+      stack.continuous_deployment && !maximum_commits_per_deploy_reached? && !active?
     end
 
     def blocked?
       return @blocked if defined?(@blocked)
       @blocked = super
+    end
+
+    private
+
+    def maximum_commits_per_deploy_reached?
+      stack.maximum_commits_per_deploy && index >= stack.maximum_commits_per_deploy
     end
   end
 end
--- a/app/views/shipit/commits/_commit.html.erb
+++ b/app/views/shipit/commits/_commit.html.erb
@@ -17,6 +17,11 @@
       <p class="commit-meta">
         <%= timeago_tag(commit.committed_at, force: true) %>
       </p>
+      <% if commit.deploy_scheduled? %>
+        <p class="commit-meta">
+	  <span class="scheduled">expected to be deployed next</span>
+        </p>
+      <% end %>
     </div>
     <div class="commit-lock" >
       <%= link_to stack_commit_path(commit.stack, commit), class: 'action-lock-commit', data: {tooltip: t('commit.lock')} do %>
--- a/test/models/undeployed_commits_test.rb
+++ b/test/models/undeployed_commits_test.rb
@@ -39,6 +39,46 @@ module Shipit
       assert_predicate @commit, :deploy_discouraged?
     end
 
+    test "#deploy_scheduled? returns true if the stack has continuous deployment enabled, commit index is lower then the maximum commits per deploy and commit is not part of active task" do
+      commit = UndeployedCommit.new(shipit_commits(:undeployed_4), 1)
+
+      assert_predicate commit.stack, :continuous_deployment
+      assert_equal 1, commit.index
+      assert_equal 3, commit.stack.maximum_commits_per_deploy
+      refute_predicate commit, :active?
+
+      assert_predicate commit, :deploy_scheduled?
+    end
+
+    test "#deploy_scheduled? returns false if the stack has continuous deployment enabled and commit index is equal or bigger then the maximum commits per deploy" do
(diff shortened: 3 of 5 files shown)


app/models/shipit/undeployed_commit.rb
```diff
@@ -35,12 +35,30 @@ def deploy_disallowed?
     end
 
     def deploy_discouraged?
-      stack.maximum_commits_per_deploy && index >= stack.maximum_commits_per_deploy
+      hit_maximum_commits_per_deploy?
```
casperisfine: It's weird to extract the method content, to finally just alias it.

app/models/shipit/undeployed_commit.rb
```diff
@@ -35,12 +35,30 @@ def deploy_disallowed?
...
+      return false unless stack.active_task?
+
+      active_task = stack.active_task
+
+      id > active_task.since_commit.id && id <= active_task.until_commit.id
+    end
+
+    private
+
+    def hit_maximum_commits_per_deploy?
```
casperisfine: Not a huge fan of that naming. I'm not a good english speaker, but somehow `maximum_commits_per_deploy_hit?` or even `maximum_commits_per_deploy_reached?` would feel more natural to me.

However I defer to native speakers.

app/models/shipit/undeployed_commit.rb
esnunes: I've extracted the method content to re-use it in the `deploy_scheduled?` predicate. In the context of `deploy_scheduled?`, the method name `deploy_discouraged?` does not express the conditions needed to validate the proposition.

app/models/shipit/undeployed_commit.rb
esnunes: I agree with you, I wasn't happy with the naming too, thanks!

JackTLi: I like `maximum_commits_per_deploy_reached?` 🙂

test/models/undeployed_commits_test.rb
```diff
@@ -39,6 +39,48 @@ class UndeployedCommitsTest < ActiveSupport::TestCase
...
     end
 
+    test "#deploy_scheduled? returns true if the stack has continuous deployment enabled, commit index is lower then the maximum commits per deploy and commit is not part of active task" do
+      commit = UndeployedCommit.new(shipit_commits(:undeployed_4), 1)
+      active_task = commit.stack.active_task
+
+      assert_predicate commit.stack, :continuous_deployment
+      assert_equal 1, commit.index
+      assert_equal 3, commit.stack.maximum_commits_per_deploy
+      refute commit.id > active_task.since_commit.id && commit.id <= active_task.until_commit.id
```
DazWorrall: This seems to be duplicating the logic of `active?`, so `refute_predicate commit, active?` would be much tidier.

test/models/undeployed_commits_test.rb
```diff
@@ -39,6 +39,48 @@ class UndeployedCommitsTest < ActiveSupport::TestCase
...
+    end
+
+    test "#deploy_scheduled? returns false if the commit is part of the active task" do
+      commit = UndeployedCommit.new(shipit_commits(:undeployed_3), 1)
+      active_task = commit.stack.active_task
+
+      assert_predicate commit.stack, :continuous_deployment
+      assert_equal 1, commit.index
+      assert_equal 3, commit.stack.maximum_commits_per_deploy
+      assert commit.id > active_task.since_commit.id && commit.id <= active_task.until_commit.id
```
DazWorrall: Ditto here (though the inverse).

app/views/shipit/commits/_commit.html.erb
```diff
@@ -17,6 +17,11 @@
       <p class="commit-meta">
         <%= timeago_tag(commit.committed_at, force: true) %>
       </p>
+      <% if commit.deploy_scheduled? %>
+        <p class="commit-meta">
+	  <span class="scheduled">scheduled to the next deploy</span>
```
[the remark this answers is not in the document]
esnunes: I guess `expected to be deployed next` fits better as we are not sure if it is going to really happen due to the checks and potential failure of the current running deployment.

app/models/shipit/undeployed_commit.rb
```diff
@@ -35,12 +35,30 @@ def deploy_disallowed?
...
       return @blocked if defined?(@blocked)
       @blocked = super
     end
+
+    def active?
+      return false unless stack.active_task?
+
+      active_task = stack.active_task
+
+      id > active_task.since_commit.id && id <= active_task.until_commit.id
```
doodzik: \nit using the `between?` method could make sense here https://ruby-doc.org/core-2.2.0/Comparable.html#method-i-between-3F

esnunes: I thought about it but `between?` uses a `>=` comparison instead of `>`. To be honest I will review this code, I was thinking about it and I think it will fail when the project has just one commit, I will add some tests to make sure how it is going to behave.

PR #888, opened 2019-04-04
Add ui for managing api clients

JackTLi:
This adds a UI for managing api clients

image

image

image

### DIFF
--- a/app/controllers/shipit/api_clients_controller.rb
+++ b/app/controllers/shipit/api_clients_controller.rb
@@ -1,0 +1,49 @@
+module Shipit
+  class ApiClientsController < ShipitController
+    include Pagination
+
+    def index
+      paginator = paginate(ApiClient.order(created_at: :desc).all)
+      @api_clients = paginator.to_a
+      @links = paginator.links
+    end
+
+    def new
+      @api_client = ApiClient.new
+    end
+
+    def create
+      @api_client = ApiClient.new(create_params.merge(creator_id: current_user.id))
+      unless @api_client.save
+        flash[:warning] = @api_client.errors.full_messages.to_sentence
+      end
+
+      respond_with(@api_client)
+    end
+
+    def show
+      @api_client = ApiClient.find(params[:id])
+    end
+
+    def update
+      @api_client = ApiClient.find(params[:id])
+      if @api_client.update(update_params)
+        options = {flash: {success: 'Successfully updated'}}
+      else
+        options = {flash: {warning: @stack.errors.full_messages.to_sentence}}
+      end
+
+      redirect_to(params[:return_to].presence || api_client_path(@api_client), options)
+    end
+
+    private
+
+    def create_params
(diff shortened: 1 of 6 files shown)


casperisfine:
Wow, this has been a long awaited feature, great idea.

Before I actually review the code, here's a little *questionnaire* to help you make sure every concern is accounted for:

- ApiClient have a `creator` field. Should this UI be global, hence list all clients from everyone, or scoped by user ?
- Should the auth token be displayed after the initial confirmation page ? If yes, should we use some kind of reveal button ?
- Shipit tokens are a bit special as in we can't regenerate them, they are derived from the record id using a private key. Hence to regenerate a token we need to delete and recreate the record. Is this something we'd like to facilitate ?
- `ApiClient` can optionally be scoped to a single stack via the `stack_id` attributes. Is that something we want to add to the UI.

Feel free to dismiss some of these concerns as suitable for a followup, but just want to makesure everything is accounted for.

JackTLi:
Thanks @casperisfine ! Appreciate you prompting the discussion

> ApiClient have a creator field. Should this UI be global, hence list all clients from everyone, or scoped by user ?

Good question, I thought about this and I believe a global list is better, since it makes it easier for collaboration or helping someone out (e.g. they are missing a permission on an existing key that one of their app is using). A filter or search functionality here could be pretty useful though 🤔 

> Should the auth token be displayed after the initial confirmation page ? If yes, should we use some kind of reveal button ?

I agree on the reveal button. I was thinking about adding that along with some auditing as well as a followup. Both for viewing the key, and also for when the key was last used.

> Shipit tokens are a bit special as in we can't regenerate them, they are derived from the record id using a private key. Hence to regenerate a token we need to delete and recreate the record. Is this something we'd like to facilitate ?

I was thinking about adding a `revoke` (soft-delete) feature as a followup-item. Assuming we could "unrevoke", I think this would have the same effect?

> ApiClient can optionally be scoped to a single stack via the stack_id attributes. Is that something we want to add to the UI.

Yeah that's a good point, I can probably patch that into this PR 👍

casperisfine:
> I was thinking about adding a `revoke` (soft-delete) feature as a followup-item. Assuming we could "unrevoke", I think this would have the same effect?

Not exactly the same effect. If you revoke / unrevoke you are making an existing token unusable.

Regenerating mean getting a new, valid token, while revoking the old one.

So basically for shipit it would mean creating a new record, will all identical attributes (except `id`), and then deleting the old one (or soft delete but WTV).

JackTLi:
Ahh I see what you mean. I _feel_ like we don't need a specific regenerate feature. 

The case I'm thinking of in my head is if a key needs to be rotated. In most key rotation cases, both keys need to be supported for a period of time anyways. So for a regeneration of key 1, key 2 needs to be created first, app moves to use key 2, and then key 1 is deleted, which lends itself better to a manual process.

PR #889, opened 2019-04-04
Track rollback safeness in shipit

JackTLi:
Button:

image

On undeployed commits list:

image

Shipit stack:

image

In this case, `add ergodox config` is the deploy which contains a unrollbackable commit. This means that we can rollback **to** this revision, but any rollback attempt on an older revision will result in the following screen:

image

Imo we should always give developers the option to rollback a commit even if it's marked as unsafe, because it's possible that it was marked as unsafe wrongly. At the same time, we should raise sufficient alarms so that these actions are performed with the right information. Any thoughts on the above ui/experience? Should we make it louder?

cc @Shopify/pipeline

### DIFF
--- a/app/assets/javascripts/merge_status.coffee
+++ b/app/assets/javascripts/merge_status.coffee
@@ -30,7 +30,7 @@ class MergeStatusPoller
 
   previousLastModified = null
   refreshPage: =>
-    @fetchPage window.location.toString(), (html, response) =>
+    @fetchPage @reloadUrl(), (html, response) =>
       @updateDocument(html, response)
       setTimeout(@refreshPage, POLL_INTERVAL)
 
@@ -41,6 +41,13 @@ class MergeStatusPoller
       if html && container = document.querySelector('[data-layout-content]')
         container.innerHTML = html
         @onPageChange()
+
+  reloadUrl: =>
+    rollback_checkbox = document.querySelector('input[name="rollbackable"]:checked')
+    if rollback_checkbox == null
+      return window.location.toString()
+    else
+      return window.location.toString() + "&rollbackable=" + rollback_checkbox.value
 
   isMergeQueueEnabled: =>
     document.querySelector('.merge-status-container .js-details-container')?.hasAttribute('data-queue-enabled')
--- a/app/assets/javascripts/merge_status.coffee
+++ b/app/assets/javascripts/merge_status.coffee
@@ -43,11 +43,8 @@ class MergeStatusPoller
         @onPageChange()
 
   reloadUrl: =>
-    rollback_checkbox = document.querySelector('input[name="rollbackable"]:checked')
-    if rollback_checkbox == null
-      return window.location.toString()
-    else
-      return window.location.toString() + "&rollbackable=" + rollback_checkbox.value
+    mergeForm =  document.querySelector('form#merge-form')
+    window.location.toString() + '&' + new URLSearchParams(new FormData(mergeForm)).toString()
 
   isMergeQueueEnabled: =>
     document.querySelector('.merge-status-container .js-details-container')?.hasAttribute('data-queue-enabled')
--- a/app/models/shipit/deploy.rb
+++ b/app/models/shipit/deploy.rb
@@ -111,11 +111,11 @@ module Shipit
     end
 
     def safe_to_rollback?
-      !commits.any?{|commit| commit.rollbackable == false}
+      !commits.any?(&:unsafe_to_rollback)
     end
 
     def safe_to_rollback_to?
-      !stack.deploys.newer_than(id).until(stack.last_completed_deploy.id).any? { |deploy| !deploy.safe_to_rollback? }
+      stack.deploys.newer_than(id).until(stack.last_completed_deploy.id).all?(&:safe_to_rollback?)
     end
 
     def currently_deployed?
--- a/app/models/shipit/deploy.rb
+++ b/app/models/shipit/deploy.rb
@@ -111,7 +111,7 @@ module Shipit
     end
 
     def safe_to_rollback?
-      !commits.any?(&:unsafe_to_rollback)
+      commits.none?(&:unsafe_to_rollback)
     end
 
     def safe_to_rollback_to?
--- a/app/models/shipit/deploy.rb
+++ b/app/models/shipit/deploy.rb
@@ -39,6 +39,16 @@ module Shipit
 
     delegate :broadcast_update, :filter_deploy_envs, to: :stack
 
+    def self.newer_than(commit)
+      return all unless commit
+      where('id > ?', commit.try(:id) || commit)
+    end
+
+    def self.until(commit)
+      return all unless commit
+      where('id <= ?', commit.try(:id) || commit)
+    end
+
     def build_rollback(user = nil, env: nil, force: false)
       Rollback.new(
         user_id: user&.id,
@@ -98,6 +108,14 @@ module Shipit
 
     def rollbackable?
       success? && supports_rollback? && !currently_deployed?
+    end
+
+    def safe_to_rollback?
+      !commits.any?{|commit| commit.rollbackable == false}
+    end
+
+    def safe_to_rollback_to?
+      !stack.deploys.newer_than(id).until(stack.last_completed_deploy.id).any? { |deploy| !deploy.safe_to_rollback? }
     end
 
     def currently_deployed?
--- a/app/views/shipit/commits/_commit.html.erb
+++ b/app/views/shipit/commits/_commit.html.erb
@@ -18,10 +18,10 @@
         <%= timeago_tag(commit.committed_at, force: true) %>
       </p>
       <p class="commit-meta">
-        <% if commit.rollbackable%>
-          <span class='rollbackable-yes'> Safe to rollback </span>
-        <% elsif commit.rollbackable == false %>
+        <% if commit.unsafe_to_rollback%>
           <span class='rollbackable-no'> Unsafe to rollback </span>
+        <% elsif commit.unsafe_to_rollback == false %>
+          <span class='rollbackable-yes'> Safe to rollback </span>
         <% else %>
           <!-- Do not display anything if rollbackable is nil -->
         <% end %>
--- a/app/views/shipit/commits/_commit.html.erb
+++ b/app/views/shipit/commits/_commit.html.erb
@@ -17,6 +17,15 @@
       <p class="commit-meta">
         <%= timeago_tag(commit.committed_at, force: true) %>
       </p>
+      <p class="commit-meta">
+        <% if commit.rollbackable%>
+          <span class='rollbackable-yes'> Safe to rollback </span>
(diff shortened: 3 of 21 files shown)


app/views/shipit/merge_status/_merge_queue_button.html.erb
```diff
@@ -9,8 +9,24 @@
 <% else %>
   <%= form_tag enqueue_pull_request_path(stack, pull_request_number), method: 'put', class: classes, data: {remote: true} do %>
     <%= hidden_field_tag 'referrer', params[:referrer] %>
-    <button type="submit" data-disable-with="Adding to merge queue…" class="btn btn-primary">
-      Add to merge queue
-    </button>
+
+    <div class='rollbackability'>
+      <%= label_tag :rollbackable do %>
+        Is this change <%= link_to('safe to rollback?', 'https://development.shopify.io/guides/shipping/safe_to_merge#Signs_your_PR_is_safe_to_rollback', target: :_blank) %>
```
JackTLi: Todo: Find a public source or make this link configurable, as the current link is only available to shopify employees I realized.

app/assets/javascripts/merge_status.coffee
```diff
@@ -42,6 +42,13 @@ class MergeStatusPoller
         container.innerHTML = html
         @onPageChange()
 
+  reloadUrl: =>
+    rollback_checkbox = document.querySelector('input[name="rollbackable"]:checked')
+    if rollback_checkbox == null
+      return window.location.toString()
+    else
+      return window.location.toString() + "&rollbackable=" + rollback_checkbox.value
```
JackTLi: The page refresh/poll here gets in the way a little bit, where after "no" is selected, a refresh can kick in which resets the radio buttons. '

This is the workaround, which will stick the current radio button value to the end of the refresh link as a query parameter, and then when we render the html.erb for the page, we will set the default checked radio depending on this param.

I spent way too long banging my head for a good solution to this, and this seems to be the least intrusive (least amount of js radio button manipulation). lmk if anyone has any suggestions

esnunes: one potential solution is: store the state of the `rollbackable` flag in the hash part of the url

JackTLi: I think you're describing the behaviour here - I'm storing the state in the url as something like:
`shipit.shopify.io/merge_status?commits=1&rollbackable=true`

JackTLi:
It'll render in the github UI exactly as the screenshot above, that page was rendered with github's stylesheet that we inject :) 

also yeah, I ran into a snag with `rollbackability` since it's already something that's used on `deploy`. I'm probably gonna change everything to `safe_to_deploy` instead.

app/assets/javascripts/merge_status.coffee
```diff
@@ -42,6 +42,13 @@ class MergeStatusPoller
         container.innerHTML = html
         @onPageChange()
 
+  reloadUrl: =>
+    rollback_checkbox = document.querySelector('input[name="rollbackable"]:checked')
```
casperisfine: So as you figured, this part of the UI really wasn't meant to hold state.

That being said, if you want to do this, a cleaner way could be to serialize the form in a more generic way, e.g. get the form element and iterate over the inputs and build the query string.

app/models/shipit/deploy.rb
```diff
@@ -100,6 +110,14 @@ def rollbackable?
       success? && supports_rollback? && !currently_deployed?
     end
 
+    def safe_to_rollback?
+      !commits.any?{|commit| commit.rollbackable == false}
```
casperisfine: `commits.all?(&:rollbackable)`

app/models/shipit/deploy.rb
```diff
@@ -100,6 +110,14 @@ def rollbackable?
...
+    end
+
+    def safe_to_rollback_to?
+      !stack.deploys.newer_than(id).until(stack.last_completed_deploy.id).any? { |deploy| !deploy.safe_to_rollback? }
```
casperisfine: This one too would be much clearer using `all?`

app/views/shipit/commits/_commit.html.erb
```diff
@@ -17,6 +17,15 @@
       <p class="commit-meta">
         <%= timeago_tag(commit.committed_at, force: true) %>
       </p>
+      <p class="commit-meta">
+        <% if commit.rollbackable%>
+          <span class='rollbackable-yes'> Safe to rollback </span>
```
casperisfine: For a boolean input, a checkbox would make much more sense IMO.

app/views/shipit/commits/_commit.html.erb
```diff
@@ -17,6 +17,15 @@
...
+        <% elsif commit.rollbackable == false %>
+          <span class='rollbackable-no'> Unsafe to rollback </span>
+        <% else %>
+          <!-- Do not display anything if rollbackable is nil -->
```
casperisfine: Tri-state boolean tend to be a smell. Why would a commit be `rollbackable=nil` ? Can't we default to either true or false ?

JackTLi: Although we can treat `nil` functionally the same as the `true` case, I don't want to surface it to the ui if it's not explicitly set. i.e. if a merge is done without going through the merge queue flow, we won't have this value, or for apps that don't have the merge queue enabled. I could maybe convert this to `unsafe_to_rollback` so that at least the `false` and `nil` cases can be used interchangeably for most of the logic?

app/views/shipit/commits/_commit.html.erb
JackTLi: hmm could you expand on this? I can't quite visualize what you're describing 😛

app/views/shipit/merge_status/_merge_queue_button.html.erb
```diff
@@ -1,16 +1,28 @@
...
-  <%= form_tag enqueue_pull_request_path(stack, pull_request_number), method: 'put', class: classes, data: {remote: true} do %>
+  <%= form_tag enqueue_pull_request_path(stack, pull_request_number), method: 'put', class: classes, data: {remote: true}, id: 'merge-form' do %>
     <%= hidden_field_tag 'referrer', params[:referrer] %>
-    <button type="submit" data-disable-with="Adding to merge queue…" class="btn btn-primary">
-      Add to merge queue
-    </button>
+
+    <div class='rollbackability'>
+      <%= label_tag :unsafe_to_rollback do %>
+        This change is <%= link_to('unsafe to rollback', 'https://development.shopify.io/guides/shipping/safe_to_merge#Signs_your_PR_is_safe_to_rollback', target: :_blank) %>
```
JackTLi: Todo: Find a public source or make this link configurable, as the current link is only available to shopify employees

JackTLi:
@DazWorrall @casperisfine @esnunes would appreciate another round of 👀

app/models/shipit/deploy.rb
```diff
@@ -39,6 +39,16 @@ def append_status(task_status)
 
     delegate :broadcast_update, :filter_deploy_envs, to: :stack
 
+    def self.newer_than(commit)
+      return all unless commit
+      where('id > ?', commit.try(:id) || commit)
```
casperisfine: I suppose you copy pasted this from `Commit`, because `id` in this context is a `deploy.id`, so that argument should probably be `deploy`

app/assets/javascripts/merge_status.coffee
```diff
@@ -42,6 +42,10 @@ class MergeStatusPoller
         container.innerHTML = html
         @onPageChange()
 
+  reloadUrl: =>
+    mergeForm =  document.querySelector('form#merge-form')
```
casperisfine: Ok, so you are going to hate me because I told you to do this.

Turns out the form contains these 3 hidden fields:

```
 <input name="utf8" type="hidden" value="✓">
 <input type="hidden" name="_method" value="put">
 <input type="hidden" name="authenticity_token" value="[REDACTED]">
```

The first two we could almost ignore, it's just ugly but fine.

But the 3rd one is different on every request, which mean we'd totally  break [the HTTP caching](https://github.com/Shopify/shipit-engine/blob/ff515fd8142b8d2d4b5fe7d5c83b5a67add52f50/app/controllers/shipit/merge_status_controller.rb#L14).

Also that same caching might cause you troubles, you'll need to include any form attribute you want to persist in there.

Really the HCTW iframe is not well suited for stateful inputs :/

I honestly have no idea how to implement that without lots of warts.

Maybe that data shouldn't be included in the URL, but reset by the javascript in the `fetchPage ` callback?

JackTLi: So I didn't go through your suggestion fully, I'm using this to serialize the forum state into the query string but I'm only using the single `unsafe_to_rollback` value to seed the initial checkbox state by. This is the page that the refresh mechanism polls, I'm not manipulating the url at all for the original page. 

The rest of the query params are actually just ignored, including the authenticity token.

&yeah I totally found that this wasn't meant for stateful inputs. Hopefully this isn't too nasty of a workaround though.

DazWorrall:
I'm really tired and my head is full of conference things so I apologise in advance for the half baked thoughts.

I'm worried about unintended consequences / perverse incentives with this change, and want to make sure we're aligned on what we want to achieve. The way I see it there are 2 big areas we want a better story on:

* When things are on fire, people dealing with the issue may not have all the context on all the changes. A quick signal on rollback safety allows them to move more quickly and more confidently at at time when they may really need to
* Have developers think about rollback safety, make a conscious decision either way, and then optionally / later have Shipit help them rollout unrollbackable changes in isolation (so that they're not bundled with other changes, which might break, and need rolling back)

If we're agreed there / I haven't missed anything, I wonder if we shouldn't actually have a default set either way, and have people deliberately pick - and refuse to merge without a choice. On the plus side it cant be ignored, and we sidestep the risk that we don't trust the boolean because people never set it. On the downside, shipping an unrollbackable change should be extremely exceptional, and so for the vast majority of PR's we're introducing unnecessary friction. I'm keen to get everyone's thoughts on this.

app/assets/javascripts/merge_status.coffee
mcgain: you could also use `window.localStorage` if you don't want the info in the url. More stuff like expiration to deal with then though.

JackTLi:
Thanks for the comments everybody, I think we can make the "Add to Merge Queue" a two step process like this:

> [Add to merge queue]
⬇️ 
 *click*
⬇️ 
 Is this change safe to rollback? [Yes] [No]
⬇️ 
*click yes or no*
⬇️ 
*change added to merge queue with flag set*

If we make it so that clicking the initial "Add to merge queue" button pauses the polling, then we can also sidestep the polling problem. Once Yes/No are selected, we can resume the polling.

doodzik:
I like that idea and that it is process driven. Maybe add this to the shipitnext document for experimentation.

JackTLi:
This has been put on hold since the interaction here is pretty janky and shipit-next might be a much better place to establish something like this

PR #890, opened 2019-04-05
Display deploying commits separately from undeployed commits

esnunes:
## Problem

When a developer look at the list of undeployed commits it is not clear enough which of the undeployed commits are currently being deployed.

## Solution

Display deploying commits separately from undeployed commits.

## Screenshot

Screenshot 2019-04-05 at 12 21 56

### DIFF
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -254,16 +254,15 @@ module Shipit
     def active_commits
       return [] unless active_task?
 
-      scope = commits.reachable.until(active_task.until_commit)
-      scope = if active_task.since_commit == active_task.until_commit
-                scope.since(active_task.since_commit)
-              else
-                scope.newer_than(active_task.since_commit)
-              end
+      scope = commits.reachable
+                     .since(active_task.since_commit)
+                     .until(active_task.until_commit)
 
       yield scope if block_given?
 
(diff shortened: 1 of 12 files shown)


app/models/shipit/stack.rb
```diff
@@ -251,8 +251,25 @@ def lock_reverted_commits!
       end
     end
 
-    def undeployed_commits
+    def active_commits
+      return [] unless active_task?
+
+      scope =
+        commits
```
doodzik: \nit commits should be on the same line as scope + the first method for readability.
```
scope = commits.reachable
                             .newer_than(active_task.since_commit)
                             .until(active_task.until_commit)

File: app/models/shipit/stack.rb [hunk 2]
```diff
@@ -251,8 +251,28 @@ def lock_reverted_commits!
       end
     end
 
-    def undeployed_commits
+    def active_commits
+      return [] unless active_task?
+
+      scope = commits.reachable.until(active_task.until_commit)
+      scope = if active_task.since_commit == active_task.until_commit
```
doodzik: reassigning to the same makes it a little bit hard to understand what each state holds.
I would change the name to communicate it better.

doodzik: very nitpicky feel free to ignore

doodzik:
Still LGTM 🚢

PR #891, opened 2019-04-12
Only mark deployable commits as expected to be deployed next

esnunes:
This PR modifies the logic used to display the message `expected to be deployed next`. From now on it is going to show this message based on the most recent deployable commit, considering the maximum commit to deploy limit.

### DIFF
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -261,7 +261,7 @@ module Shipit
       yield scope if block_given?
 
       scope.select(&:active?)
-           .map.with_index { |c, i| UndeployedCommit.new(c, i) }
+           .map.with_index { |c, i| UndeployedCommit.new(c, index: i, next_commit_to_deploy: next_commit_to_deploy) }
            .reverse
     end
 
@@ -273,7 +273,10 @@ module Shipit
       end
 
       yield scope if block_given?
-      scope.map.with_index { |c, i| UndeployedCommit.new(c, i) }.reverse
+
+      scope.map
+           .with_index { |c, i| UndeployedCommit.new(c, index: i, next_commit_to_deploy: next_commit_to_deploy) }
+           .reverse
     end
 
     def last_completed_deploy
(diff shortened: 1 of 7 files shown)


app/models/shipit/stack.rb
```diff
@@ -168,9 +168,13 @@ def schedule_merges
     end
 
     def next_commit_to_deploy
-      commits_to_deploy = commits.order(id: :asc).newer_than(last_deployed_commit).reachable.preload(:statuses)
-      commits_to_deploy = commits_to_deploy.limit(maximum_commits_per_deploy) if maximum_commits_per_deploy
-      commits_to_deploy.to_a.reverse.find(&:deployable?)
+      return @next_commit_to_deploy if defined?(@next_commit_to_deploy)
```
JackTLi: A nice ruby way to do this is to do:
```ruby
def next_commit_to_deploy
  @next_commit_to_deploy ||= begin
    ...
  end
end
```

😄

esnunes: I thought about using `||=` however in this case if there is no commit to deploy it will return `nil` and it won't be cached because `nil` is going to be false, that's why I decided to use `defined?`, by using it I can "cache" false values too.

I've noticed some tests failed because of this cache, I could change the signature of this method to `next_commit_to_deploy(cache: true)` but I'm not convinced it is a good approach, do you have a better suggestion?

app/views/shipit/stacks/show.html.erb
```diff
@@ -17,7 +17,7 @@
       </div>
     </header>
     <ul class="commit-list">
-      <%= render partial: 'shipit/commits/commit', collection: @undeployed_commits %>
+      <%= render partial: 'shipit/commits/commit', collection: @undeployed_commits, locals: {next_commit_to_deploy: @next_commit_to_deploy} %>
```
JackTLi: this makes sense to me 👍

app/controllers/shipit/stacks_controller.rb
```diff
@@ -21,6 +21,7 @@ def show
         scope.preload(:author, :statuses, :check_runs)
       end
       @active_commits = @stack.active_commits { |scope| scope.preload(:author, :statuses, :check_runs) }
+      @next_commit_to_deploy = @stack.next_commit_to_deploy
```
casperisfine: Did you try passing this to the `UndeployedCommit` decorator?

PR #894, opened 2019-04-17
Expire cache of commit partial when either commit or stack change

esnunes:
## Problem

Some `UndeployedCommit` predicates depend on the state of the stack to be computed. At the moment the cache expires only when the commit data changes not considering changes in the stack.

## Solution

Modify the cache key to be a composition of `commit` and `commit.stack`.

### DIFF
--- a/app/views/shipit/commits/_commit.html.erb
+++ b/app/views/shipit/commits/_commit.html.erb
@@ -1,5 +1,5 @@
 <li class="commit <%= 'locked' if commit.locked? %>" id="commit-<%= commit.id %>">
-  <% cache commit do %>
+  <% cache [commit, commit.stack] do %>
     <% cache commit.author do %>
       <%= render 'shipit/shared/author', author: commit.author %>
     <% end %>


casperisfine:
The problem with this is that it males that cache mostly useless.

Any change on a commit bubbles up to the stack: https://github.com/Shopify/shipit-engine/blob/62e80dad23c213316506311dc367fcb0a3ff2adf/app/models/shipit/commit.rb#L15

So in the end, this cache will be busted all the time.

So I see two options:

  - Simply remove it, but that will make `stacks#show` much slower.
  - Try to build a smarter cache key out of the stack.

esnunes:
> Simply remove it, but that will make stacks#show much slower.
What if we move the cache to `show.html.erb` and use only the stack as key? I understand it is not the best solution but it would be better than completely remove it.

> Try to build a smarter cache key out of the stack.
I will try to come up with a smarted cache key.

Thanks!

casperisfine:
> What if we move the cache to `show.html.erb` and use only the stack as key?

We already do that: https://github.com/Shopify/shipit-engine/blob/62e80dad23c213316506311dc367fcb0a3ff2adf/app/views/shipit/stacks/show.html.erb#L33

But that cache is busted quite often.

esnunes:
Yeah, we do it for tasks but not for commits.

casperisfine:
Oh right I missed that. Yeah, would be better than nothing.

esnunes:
I'm closing this PR in favour of: https://github.com/Shopify/shipit-engine/pull/893/files#diff-e923d385c81944d674ffa7ede97c8660R2

PR #895, opened 2019-04-17
Add rake deploy command

doodzik:
This command gives the ability to trigger a deploy through the command line interface when sshed into a running Shipit Instance. This is useful when the UI isn't responsive.

### DIFF
--- a/app/models/shipit/command_line_user.rb
+++ b/app/models/shipit/command_line_user.rb
@@ -1,0 +1,58 @@
+module Shipit
+  class CommandLineUser
+    def present?
+      false
+    end
+
+    def email
+      'command_line@example.com'
+    end
+
+    def login
+      'command_line'
+    end
+
+    def name
+      'CommandLine'
+    end
+
+    def avatar_url
+      'https://github.com/images/error/octocat_happy.gif'
+    end
+
+    def id
+    end
+
+    def github_id
+    end
+
+    def logged_in?
+      false
+    end
+
+    def authorized?
+      Shipit.authentication_disabled?
+    end
+
+    def stacks_contributed_to
+      []
+    end
+
+    def avatar_uri
+      User::DEFAULT_AVATAR.dup
+    end
+
+    def created_at
+      Time.at(0).utc
+    end
+    alias_method :updated_at, :created_at
+
+    def read_attribute_for_serialization(attr)
+      public_send(attr)
+    end
+
+    def github_api
+      Shipit.github.api
+    end
+  end
+end
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -124,10 +124,11 @@ module Shipit
       )
     end
 
-    def trigger_deploy(*args)
-      deploy = build_deploy(*args)
+    def trigger_deploy(*args, **kwargs)
+      run_now = kwargs.delete(:run_now)
+      deploy = build_deploy(*args, **kwargs)
       deploy.save!
-      deploy.enqueue
+      run_now ? deploy.run_now! : deploy.enqueue
       continuous_delivery_resumed!
       deploy
     end
@@ -399,6 +400,15 @@ module Shipit
 
     def to_param
       [repo_owner, repo_name, environment].join('/')
+    end
+
+    def self.run_deploy_in_foreground(stack:, revision:)
+      stack = Shipit::Stack.from_param!(stack)
+      until_commit = stack.commits.where(sha: revision).limit(1).first
+      env = stack.cached_deploy_spec.default_deploy_env
+      current_user = Shipit::CommandLineUser.new
+
+      stack.trigger_deploy(until_commit, current_user, env: env, force: true, run_now: true)
     end
 
     def self.from_param!(param)
--- a/app/models/shipit/task.rb
+++ b/app/models/shipit/task.rb
@@ -176,6 +176,11 @@ module Shipit
       PerformTaskJob.perform_later(self)
     end
 
+    def run_now!
+      raise "only persisted jobs can be run" unless persisted?
+      PerformTaskJob.perform_now(self)
+    end
+
     def write(text)
       chunks.create!(text: text)
     end
--- a/test/models/stacks_test.rb
+++ b/test/models/stacks_test.rb
@@ -133,6 +133,15 @@ module Shipit
       assert_instance_of Deploy, deploy
     end
 
+    test "#trigger_deploy doesn't enqueues a deploy job when run_now is provided" do
+      @stack.deploys.destroy_all
+      Deploy.any_instance.expects(:run_now!).once
+
+      last_commit = shipit_commits(:third)
+      deploy = @stack.trigger_deploy(last_commit, AnonymousUser.new, run_now: true)
+      assert_instance_of Deploy, deploy
+    end
+
     test "#trigger_deploy marks the deploy as `ignored_safeties` if the commit wasn't deployable" do
       last_commit = shipit_commits(:fifth)
       refute_predicate last_commit, :deployable?
@@ -261,6 +270,16 @@ module Shipit
     test "#active_task? is true if a rollback is ongoing" do
       shipit_deploys(:shipit_complete).trigger_rollback(AnonymousUser.new)
       assert @stack.active_task?
+    end
+
+    test ".run_deploy_in_foreground triggers a deploy" do
+      stack = Stack.create!(repo_owner: 'foo', repo_name: 'bar', environment: 'production')
+      commit = shipit_commits(:first)
+      stack.commits << commit
+
+      Stack.any_instance.expects(:trigger_deploy).with(anything, anything, env: {}, force: true, run_now: true)
+
+      Stack.run_deploy_in_foreground(stack: stack.to_param, revision: commit.sha)
     end
 
     test "#active_task? is memoized" do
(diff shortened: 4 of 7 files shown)


Rakefile
```diff
@@ -22,3 +22,31 @@ Rake::TestTask.new(:test) do |t|
...
+    raise ArgumentError.new('The second argument has to be a revision') if revision.nil?
+
+    module Shipit
+      class Task
+        def write(text)
+          p text
+          chunks.create!(text: text)
+        end
+      end
+    end
```
doodzik: Monkey path the write method so we can output the logs as they happen to the CLI. This monkey patch only affect this rake task.

test/models/stacks_test.rb
```diff
@@ -262,6 +271,17 @@ def setup
       assert @stack.active_task?
     end
 
+    test ".run_deploy_in_foreground triggers a deploy" do
+      stack = Stack.create!(repo_owner: 'foo', repo_name: 'bar', environment: 'production')
+      commit = shipit_commits(:first)
+      stack.commits << commit
+
+      Stack.any_instance.stubs(:trigger_deploy).with(anything, is_a(AnonymousUser), env: [], force: true, skip_queue: true)
```
doodzik: I wasn't able to replace the `anything` with the `is_a?(Commit)` object as it wasn`t the same class (Commit vs ActiveRecord_Relation).
I tried `.with { |actual| actual }` but the block argument didn't include the right number of passed arguments. 
https://www.rubydoc.info/github/floehopper/mocha/Mocha/Expectation#with-instance_method

test/models/stacks_test.rb
```diff
@@ -262,6 +271,17 @@ def setup
...
+
+      Stack.run_deploy_in_foreground(stack: "#{stack.repo_owner}/#{stack.repo_name}/#{stack.environment}",
```
JackTLi: `stack.to_param` should give you the stack in this format 😄

app/models/shipit/task.rb
```diff
@@ -171,9 +171,9 @@ def spec
       @spec ||= DeploySpec::FileSystem.new(working_directory, stack.environment)
     end
 
-    def enqueue
+    def enqueue(skip_queue: false)
```
JackTLi: perhaps `inline` vs `skip_queue` to highlight the fact that it doesn't just run immediately, but runs synchronously?

test/models/stacks_test.rb
```diff
@@ -262,6 +271,17 @@ def setup
...
+                                     revision: commit.sha)
+    end
```
JackTLi: It seems like this test is missing an assertion?

test/models/stacks_test.rb
JackTLi: Oh was this supposed to be an `expect` vs `stub` ? 

I'm fine with the `with` call, though really the only field we really care about is that it's using `skip_queue: true` right? If so I would be fine with putting `anything` for the rest of the fields so that we can really narrow the focus on this test (also becomes more resilient to other minor changes to this method's signature)

Rakefile
JackTLi: hm this is definitely a smell.

What if we add a property to `tasks` called `inline`/`skip_queue` that we pass here:
https://github.com/Shopify/shipit-engine/pull/895/files#diff-866796f172b542adeaf3b08f1d403511R128
so that we can just add a check to tasks#write instead? 😄

app/models/shipit/stack.rb
```diff
@@ -398,6 +398,15 @@ def to_param
       [repo_owner, repo_name, environment].join('/')
     end
 
+    def self.run_deploy_in_foreground(stack:, revision:)
+      stack = Shipit::Stack.from_param!(stack)
+      until_commit = stack.commits.where(sha: revision).limit(1)
+      env = stack.deploy_variables.map(&:name)
+      current_user = Shipit::AnonymousUser.new
```
DazWorrall: I wonder if we should create a `CLIUser` or similar, so it's explicit where the deploy was triggered.

app/models/shipit/task.rb
DazWorrall: Or even a new method - `run_now!` or `run_inline!` maybe.

app/models/shipit/stack.rb
```diff
@@ -124,10 +124,10 @@ def build_deploy(until_commit, user, env: nil, force: false)
       )
     end
 
-    def trigger_deploy(*args)
-      deploy = build_deploy(*args)
+    def trigger_deploy(until_commit, user, env: nil, force: false, skip_queue: false)
```
DazWorrall: This is a bit tidier, to avoid duplicating the `build_deploy` signature:

```suggestion
    def trigger_deploy(*args, **kwargs)
      skip_queue = kwargs.delete(:skip_queue)
      deploy = build_deploy(*args, **kwargs)
```

Rakefile
```diff
@@ -22,3 +22,31 @@ Rake::TestTask.new(:test) do |t|
 end
 
 task default: :test
+
+desc "Deploy from a running instance. "
+task :deploy, %i(stack revision) => :environment do |_t, args|
```
DazWorrall: Will this Rake task be bundled with the Shipit gem? Would also be nice if it was namespaced i.e. `shipit::deploy`

Rakefile
doodzik: I would prefer the job to be unaware of `inline`/`skip_queue`. The primary reason is that I don't want to introduce complexity that isn't necessary as `inline` is only going to be used for this command. If we had more use-cases for that option I would go that route.

test/models/stacks_test.rb
doodzik: This method kind of functions as a procedure, which makes it hard to test. The `with` expectation works as an expectation in this case. Suggestions on a better way to test this are welcome :)

app/models/shipit/task.rb
doodzik: I'll go with `run_now!` as it reflects more what is going on 👍

Rakefile
doodzik: yes, but we will have to load it in the shipit rakefile to use it

test/models/stacks_test.rb
doodzik: see the other comment

test/models/stacks_test.rb
JackTLi: I believe that with the `stubs.with`, it'll only fail if the method was called with the wrong arguments. Vs. if you go with `expects.with`, it'll fail if the method was not called at all, or if it was called with the wrong arguments

test/models/stacks_test.rb
```diff
@@ -263,6 +272,16 @@ def setup
       assert @stack.active_task?
     end
 
+    test ".run_deploy_in_foreground triggers a deploy" do
+      stack = Stack.create!(repo_owner: 'foo', repo_name: 'bar', environment: 'production')
+      commit = shipit_commits(:first)
+      stack.commits << commit
+
+      Stack.any_instance.stubs(:trigger_deploy).with(anything, anything, env: [], force: true, run_now: true)
```
JackTLi: (reposting comment on new diff)

I believe that with the stubs.with, it'll only fail if the method was called with the wrong arguments. Vs. if you go with expects.with, it'll fail if the method was not called at all, or if it was called with the wrong arguments

doodzik: I misunderstood you last time. I'll try it out. Thanks for clarifying it :)

Rakefile
```diff
@@ -22,3 +22,33 @@ Rake::TestTask.new(:test) do |t|
...
+            p text
+            chunks.create!(text: text)
+          end
+        end
+      end
+
+      Shipit::Stack.run_deploy_in_foreground(stack: stack, revision: revision)
+    rescue ArgumentError
+      p "Use this command as follows:"
+      p "bundle exec rake shipit::deploy\['shopify/shipit/production','593b1f07cec6c30df3f62d9e63b31dc0ff444098'\]"
```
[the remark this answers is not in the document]
doodzik: We could use `ARGV` instead. Maybe that would be a better way. Let me double check.

app/models/shipit/command_line_user.rb
```diff
@@ -0,0 +1,58 @@
+module Shipit
+  class CommandLineUser
```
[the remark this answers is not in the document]
doodzik: Yes, we could do that. 
IMO I would prefer to keep it a little bit less dry though. This way we can wait and see what kind of abstraction makes sense in the long term. I wouldn't want to lock ourselves into an abstraction that doesn't hold up when we have to add another user. (Not that we will, but who know)

app/models/shipit/stack.rb
```diff
@@ -401,6 +402,15 @@ def to_param
       [repo_owner, repo_name, environment].join('/')
     end
 
+    def self.run_deploy_in_foreground(stack:, revision:)
+      stack = Shipit::Stack.from_param!(stack)
+      until_commit = stack.commits.where(sha: revision).limit(1)
+      env = stack.deploy_variables.map(&:name)
```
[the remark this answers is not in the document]
doodzik: I think you are right. Nice catch.

looking at line 162 of the same file, It should be `env = stack.cached_deploy_spec.default_deploy_env` instead.

doodzik:
Screen Shot 2019-05-16 at 15 06 03

Ran it in production. It deployed successfully and it logged locally and on the web ui simultaneously.

PR #896, opened 2019-04-22
feat: add support for publishing packages with lerna 3

gf3:
Add support for new `publish` command syntax present in lerna 3

### DIFF
--- a/app/models/shipit/deploy_spec/lerna_discovery.rb
+++ b/app/models/shipit/deploy_spec/lerna_discovery.rb
@@ -3,6 +3,8 @@ require 'json'
 module Shipit
   class DeploySpec
     module LernaDiscovery
+      LATEST_MAJOR_VERSION = Gem::Version.new('3.0.0')
+
       def discover_dependencies_steps
         discover_lerna_json || super
       end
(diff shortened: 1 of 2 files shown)


app/models/shipit/deploy_spec/lerna_discovery.rb
```diff
@@ -21,15 +23,25 @@ def discover_review_checklist
...
-            You can do this with:<br/>
-            <pre>
-            lerna publish --skip-npm
-            && git add -A
-            && git push --follow-tags
-            </pre>
-          )]
+          if lerna_lerna >= LATEST_MAJOR_VERSION
+            [%(
+              <strong>Don't forget version and tag before publishing!</strong>
```
DazWorrall: nit: this can be de-duped with the `else` branch.

app/models/shipit/deploy_spec/lerna_discovery.rb
```diff
@@ -41,6 +53,11 @@ def lerna_json
         file('lerna.json')
       end
 
+      def lerna_lerna
+        lerna_config = lerna_json.read
+        Gem::Version.new(JSON.parse(lerna_config)['lerna'])
```
DazWorrall: Should/could we memoize this? Could this JS file every be large / expensive to parse?

app/models/shipit/deploy_spec/lerna_discovery.rb
gf3: done!

app/models/shipit/deploy_spec/lerna_discovery.rb
gf3: fixed!

PR #897, opened 2019-04-25
Move creation of UndeployedCommit to controller

esnunes:
This PR improves `stacks_controller#show` method by:

- Assigning block result to the query builder, enabling `preload` of `author`, `statuses` and `check_runs` (https://github.com/Shopify/shipit-engine/commit/6274d10110d3260584eae2c8b300804c734d5563#diff-866796f172b542adeaf3b08f1d403511R269);
- Refactor `Stack#next_expected_commit_to_deploy` to identify commit based on a list of commits (https://github.com/Shopify/shipit-engine/commit/6274d10110d3260584eae2c8b300804c734d5563#diff-866796f172b542adeaf3b08f1d403511R254);
- Moving creation of `UndeployedCommit` to `stacks_controller#show`, avoiding multiple calls to `Stack#next_expected_commit_to_deploy` (https://github.com/Shopify/shipit-engine/commit/6274d10110d3260584eae2c8b300804c734d5563#diff-f1a38892b030c28340444e0f9555ef71R34);

### DIFF
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -251,42 +251,24 @@ module Shipit
       end
     end
 
-    def active_commits
-      return [] unless active_task?
+    def next_expected_commit_to_deploy(commits: nil)
+      commits ||= undeployed_commits do |scope|
+        scope.preload(:statuses, :check_runs)
+      end
 
-      scope = commits.reachable
-                     .since(active_task.since_commit)
-                     .until(active_task.until_commit)
-
-      yield scope if block_given?
-
-      scope.select(&:active?)
-           .map.with_index { |c, i| UndeployedCommit.new(c, index: i) }
-           .reverse
+      commits_to_deploy = commits.reject(&:active?)
+      if maximum_commits_per_deploy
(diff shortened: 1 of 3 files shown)


app/models/shipit/stack.rb
```diff
@@ -251,42 +251,24 @@ def lock_reverted_commits!
...
-                     .since(active_task.since_commit)
-                     .until(active_task.until_commit)
-
-      yield scope if block_given?
-
-      scope.select(&:active?)
-           .map.with_index { |c, i| UndeployedCommit.new(c, index: i) }
-           .reverse
-    end
+    def next_expected_commit_to_deploy(commits: nil)
```
DazWorrall: I see no unit tests for this - should we add some?

esnunes:
> Shall we move our instance to this branch to validate the perf benefits before merging?

I think it is a good idea!

app/models/shipit/stack.rb
esnunes: Yes, I will add some, thanks!

esnunes:
We've tested these changes in production and they worked as expected, no performance issues anymore.

PR #898, opened 2019-04-25
update PyPI discovery to use twine to upload

tylerball:
I'm unable to deploy shopify_python_api to PyPI. The [Python Packaging Guide](https://packaging.python.org/tutorials/packaging-projects/) now recommends using the `twine` package to upload eggs to PyPI and testing locally this seems to work fine.

### DIFF
--- a/app/models/shipit/deploy_spec/pypi_discovery.rb
+++ b/app/models/shipit/deploy_spec/pypi_discovery.rb
@@ -29,7 +29,11 @@ module Shipit
       end
 
       def publish_egg
-        ["assert-egg-version-tag #{setup_dot_py}", 'python setup.py register sdist upload']
+        [
+          "assert-egg-version-tag #{setup_dot_py}",
+          'python setup.py register sdist',
+          'twine upload dist/*',
+        ]
       end
     end
   end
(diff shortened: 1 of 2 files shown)


dylanahsmith:
```
Shipit::DeploySpecTest#test_#publish_egg_first_check_if_version_tag_have_been_created_and_then_invoke_setup.py_upload [/home/travis/build/Shopify/shipit-engine/test/models/deploy_spec_test.rb:327]:
--- expected
+++ actual
@@ -1 +1 @@
-["assert-egg-version-tag /tmp/fake_setup.py", "python setup.py register sdist upload"]
+["assert-egg-version-tag /tmp/fake_setup.py", "python setup.py register sdist", "twine upload dist/*"]
```

Looks like you need to update a test as well

tylerball:
@dylanahsmith test updated!

test/dummy/db/schema.rb
```diff
@@ -239,7 +239,7 @@
     t.integer "additions", limit: 4, default: 0
     t.integer "deletions", limit: 4, default: 0
     t.text "definition", limit: 65535
-    t.binary "gzip_output"
+    t.binary "gzip_output", limit: 16777215
```
byroot: :/

I never quite understood why these schema changes come and go...

Issue #876, opened 2019-02-28
Better handle Redis connection error in PerformTaskJob


casperisfine:
### Context

Tasks being performed regularly update a key in Redis: https://github.com/Shopify/shipit-engine/blob/36e3cf096837d9fb8b2d01046f4be321f6b61e96/app/models/shipit/task.rb#L241-L247

The idea is that if the job was hard killed, the rest of the app can notice it once the key starts to get old and take action.

However if there's a Redis blip (e.g. a short service disruption on the service instance) setting the key would fail and bring the job down.

It would be much better to have some error handling here:

```
Redis::CannotConnectError: Error connecting to Redis on redis-xxxx.shipit-production:6379 (Redis::TimeoutError)
    gems/redis-3.3.5/lib/redis/client.rb:345:in `rescue in establish_connection'
    gems/redis-3.3.5/lib/redis/client.rb:330:in `establish_connection'
    ...
    gems/shipit-engine-f5f12a007caa/app/models/shipit/task.rb:242:in `ping'
    gems/shipit-engine-f5f12a007caa/app/jobs/shipit/perform_task_job.rb:78:in `block in capture!'
```

It probably should still give up after Redis is not responding for too long, but it should properly mark the task as failed in the same time so that it doesn't stay as running for too long.

esnunes:
Closed by #899

PR #899, opened 2019-04-26
Better handle Redis connection error in PerformTaskJob

esnunes:
This PR covers the issue #876 by:

- Preventing `async_refresh_deployed_version` to fail when trying to dispatch a job (when this method fails the task status transition to `error` is rolled back);
- Increasing the number of redis connections retries (from 1 to 3);
- Adding support to delayed redis connections retries (when using redis client version 4.0.3+);

### DIFF
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -182,6 +182,12 @@ module Shipit
     end
 
     def async_refresh_deployed_revision
+      async_refresh_deployed_revision!
+    rescue => error
+      logger.warn "Failed to dispatch FetchDeployedRevisionJob: [#{error.class.name}] #{error.message}"
+    end
+
+    def async_refresh_deployed_revision!
       FetchDeployedRevisionJob.perform_later(self)
     end
 
--- a/lib/shipit.rb
+++ b/lib/shipit.rb
@@ -75,7 +75,13 @@ module Shipit
   end
 
   def redis(namespace = nil)
-    @redis ||= Redis.new(url: redis_url.to_s.presence, logger: Rails.logger)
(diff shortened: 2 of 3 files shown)


app/models/shipit/stack.rb
```diff
@@ -182,6 +182,12 @@ def deployed_too_recently?
     end
 
     def async_refresh_deployed_revision
+      async_refresh_deployed_revision!
+    rescue => error
+      logger.warn "Failed to dispatch FetchDeployedRevisonJob: #{error.message}"
```
DazWorrall: Would be good to include the class name here: `[#{error.class.name}] #{error.message}`

app/models/shipit/stack.rb
```diff
@@ -182,6 +182,12 @@ def deployed_too_recently?
     end
 
     def async_refresh_deployed_revision
+      async_refresh_deployed_revision!
+    rescue => error
```
DazWorrall: Would be good to see a unit test trigger this code path.

lib/shipit.rb
```diff
@@ -75,7 +75,15 @@ def redis_url
   end
 
   def redis(namespace = nil)
-    @redis ||= Redis.new(url: redis_url.to_s.presence, logger: Rails.logger)
+    @redis ||= begin
+      Redis.new(
```
casperisfine: style: you can get rid of the begin/else:

```ruby
@redis ||= Redis.new(
  url: ....,
  foo: ...,
)
```

esnunes: good catch, thanks!

esnunes:
This closes #876

Issue #860, opened 2019-01-04
Show who locked a commit


davidcornu:
If you're looking at a stack page and see some locked commits, you can't tell who locked them.

This would be useful for

- Asking the person who locked the commit for more details
- Determining whether a commit was accidentally locked

----

The mechanism we use for `Stack` is pretty straightforward

https://github.com/Shopify/shipit-engine/blob/070da14b4f447f43f70377e095136edeea91af6f/test/dummy/db/schema.rb#L205

https://github.com/Shopify/shipit-engine/blob/070da14b4f447f43f70377e095136edeea91af6f/app/models/shipit/stack.rb#L369-L376

https://github.com/Shopify/shipit-engine/blob/070da14b4f447f43f70377e095136edeea91af6f/app/views/shipit/stacks/_banners.html.erb#L39

so we could likely do something similar for `Commit`.

casperisfine:
Indeed. PR Welcome.

davidcornu:
Implemented in #901

PR #901, opened 2019-05-02
Commit lock author

davidcornu:
Implements https://github.com/Shopify/shipit-engine/issues/860.

image

### DIFF
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -240,21 +240,23 @@ module Shipit
     end
 
     def lock_reverted_commits!
-      commits_to_lock = []
-
       backlog = undeployed_commits.to_a
+      affected_rows = 0
+
       until backlog.empty?
         backlog = backlog.drop_while { |c| !c.revert? }
-        if revert = backlog.shift
-          commits_to_lock += backlog.reverse.drop_while { |c| !revert.revert_of?(c) }
-        end
+        revert = backlog.shift
+        next if revert.nil?
+
+        commits_to_lock = backlog.reverse.drop_while { |c| !revert.revert_of?(c) }
+        next if commits_to_lock.empty?
+
+        affected_rows += commits
+                         .where(id: commits_to_lock.map(&:id).uniq)
+                         .lock_all(revert.author)
       end
 
-      unless commits_to_lock.empty?
-        if commits.where(id: commits_to_lock.map(&:id).uniq).update_all(locked: true) > 1
-          touch
-        end
-      end
+      touch if affected_rows > 1
     end
 
     def next_expected_commit_to_deploy(commits: nil)
--- a/db/migrate/20190502020249_add_lock_author_id_to_commits.rb
+++ b/db/migrate/20190502020249_add_lock_author_id_to_commits.rb
@@ -1,0 +1,5 @@
+class AddLockAuthorIdToCommits < ActiveRecord::Migration[5.2]
+  def change
+    add_column(:commits, :lock_author_id, :integer, limit: 4)
+  end
+end
--- a/test/dummy/db/schema.rb
+++ b/test/dummy/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_10_10_150947) do
+ActiveRecord::Schema.define(version: 2019_05_02_020249) do
 
   create_table "api_clients", force: :cascade do |t|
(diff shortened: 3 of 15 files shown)


test/dummy/db/schema.rb
```diff
@@ -76,6 +76,7 @@
     t.string "pull_request_title", limit: 1024
     t.integer "pull_request_id"
     t.boolean "locked", default: false, null: false
+    t.integer "lock_author_id", limit: 4
```
davidcornu: Done to match https://github.com/Shopify/shipit-engine/blob/5458ffe1cb8f76a505a22cc0784876c203977f48/test/dummy/db/schema.rb#L206

davidcornu:
Fixed the Rubocop violations

db/migrate/20190502020249_add_lock_author_id_to_commits.rb
```diff
@@ -0,0 +1,5 @@
+class AddLockAuthorIdToCommits < ActiveRecord::Migration[5.2]
+  def change
+    add_column(:commits, :lock_author_id, :integer, limit: 4)
```
casperisfine: `:integer`? 

Is `users.id` integer as well? If so we should prbably migrate to bigint someday, now Rails default.

casperisfine:
Wait, there is another code path, that might need change.

db/migrate/20190502020249_add_lock_author_id_to_commits.rb
davidcornu: I matched `lock_author_id` on `stacks`. Was also surprised.

casperisfine: Ok, disregard then.

casperisfine:
@davidcornu Here the other path that would need author handling: https://github.com/Shopify/shipit-engine/blob/f6b93d8a242962016e0f82e99c2eb223bdc75003/app/models/shipit/stack.rb#L254

app/models/shipit/stack.rb
```diff
@@ -239,7 +239,7 @@ def status
       :default
     end
 
-    def lock_reverted_commits!
+    def lock_reverted_commits!(user = AnonymousUser.new)
```
davidcornu: There's another code path that usees this https://github.com/Shopify/shipit-engine/blob/868398f243c63b6e5e08a07115c29e36d0048ba6/app/jobs/shipit/github_sync_job.rb#L15-L22

I left that one as commits being locked as a result of syncing with GH don't have an obvious lock author.

casperisfine: > syncing with GH don't have an obvious lock author.

I would have used the revert author, but both are fine.

davidcornu: 🤔 Given `appended_commits.any?(&:revert?)`, there could be multiple no?

casperisfine: In SyncJob yes, but here: https://github.com/Shopify/shipit-engine/blob/f6b93d8a242962016e0f82e99c2eb223bdc75003/app/models/shipit/stack.rb#L248-L249

It's not ambiguous.

app/models/shipit/stack.rb
```diff
@@ -240,21 +240,23 @@ def status
     end
 
     def lock_reverted_commits!
```
davidcornu: @casperisfine thoughts on this change? I refactored the logic to lock in batches rather than all at once so the `lock_author` can be set to the revert commit author.

Tests are gonna fail. Will fix after work.

casperisfine: LGTM. I'd use `affected_rows` rather than `affected` though /nitpick.

test/models/stacks_test.rb
```diff
@@ -729,5 +730,105 @@ def setup
 
       @stack.async_refresh_deployed_revision
     end
+
+    test "#lock_reverted_commits! locks all commits between the original and reverted commits" do
```
davidcornu: Ended up adding a bunch of test coverage here as there's some in https://github.com/Shopify/shipit-engine/blob/22403f4c7d51e70e9d3c0c075afac96c7e0d7584/test/jobs/github_sync_job_test.rb#L39-L87 and https://github.com/Shopify/shipit-engine/blob/22403f4c7d51e70e9d3c0c075afac96c7e0d7584/test/models/rollbacks_test.rb#L22-L56 but not enough to convince me that I wasn't breaking anything with the refactor.

I've also split the addition of these tests and the refactor into separate commits so it's more obvious what my changes actually did.

PR #903, opened 2019-05-06
Add stack deletion endpoint to API

lei-lo:
So bots can delete stacks too 🤖🔥

### DIFF
--- a/app/controllers/shipit/api/stacks_controller.rb
+++ b/app/controllers/shipit/api/stacks_controller.rb
@@ -2,7 +2,7 @@ module Shipit
   module Api
     class StacksController < BaseController
       require_permission :read, :stack, only: %i(index show)
-      require_permission :write, :stack, only: %i(create)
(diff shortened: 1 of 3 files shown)


app/controllers/shipit/api/stacks_controller.rb
```diff
@@ -25,6 +25,11 @@ def show
         render_resource stack
       end
 
+      def destroy
+        stack.schedule_for_destroy!
+        render_resources stacks
```
etiennebarrie: I don't think returning the list of stacks is the right API return value for a DELETE. Based on https://tools.ietf.org/html/rfc7231#section-4.3.5 it should probably be a 202 Accepted since the deletion hasn't happened yet, or maybe 204 if we know it will always work.

lei-lo:
> We'll need a test as well.
@etiennebarrie can you elaborate on what we need a test for?

PR #906, opened 2019-05-14
Report signalled commands

### LINKED ISSUE #882: The Command class doesn't handle signal exits properly
When the sub process is killed (e.g. by OOM killer) Shipit can't find the status.

It's because it's bit naive and simply access `Process::Status#exitstatus`.

Reality is a bit more complicated, here's a snippet from an internal app we should take inspiration from:

```ruby
    def exitstatus
      @exitstatus ||= status.exitstatus
    end

    def termsig
      status.termsig
    end

    def exited?
      status.exited?
    end

    def signaled?
      status.signaled?
    end

    def success?
      @status.success?
    end

    def running?
      !!pid && !@status
    end

    def terminated?
      !!pid && !!@status
    end

    def timedout?
      @timedout
    end

    def termination_status
      if running?
        "is running"
      elsif success?
        "terminated successfully"
      elsif timedout? && termsig
        "timed out and terminated with #{Signal.signame(termsig)} signal"
      elsif timedout?
        "timed out and terminated with exit status #{exitstatus}"
      elsif signaled?
        "terminated with #{Signal.signame(termsig)} signal"
      else
        "terminated with exit status #{exitstatus}"
      end
    end
```

cc @Shopify/pipeline

DazWorrall:
Closes #882 

The purpose of this change is to have `exit_message` show something meaningful when a command is signalled to exit.

### DIFF
--- a/lib/shipit/command.rb
+++ b/lib/shipit/command.rb
@@ -30,6 +30,7 @@ module Shipit
       @timeout = options['timeout'.freeze] || options[:timeout] || default_timeout
       @env = env
       @chdir = chdir.to_s
+      @timed_out = false
     end
 
     def with_timeout(new_timeout)
@@ -129,6 +130,10 @@ module Shipit
     end
 
     def timed_out?
+      @timed_out
+    end
+
+    def output_timed_out?
       @last_output_at ||= Time.now.to_i
       (@last_output_at + timeout) < Time.now.to_i
     end
@@ -149,7 +154,10 @@ module Shipit
           yield io.read_nonblock(MAX_READ)
           touch_last_output_at
         rescue IO::WaitReadable
-          raise TimedOut if timed_out?
+          if output_timed_out?
+            @timed_out = true
+            raise TimedOut
+          end
           IO.select([io], nil, nil, 1)
           retry
         end
@@ -243,6 +251,10 @@ module Shipit
         "is running"
       elsif success?
         "terminated successfully"
+      elsif timed_out? && signaled?
+        "timed out and terminated with #{Signal.signame(@status.termsig)} signal"
+      elsif timed_out?
+        "timed out and terminated with exit status #{exitstatus}"
       elsif signaled?
         "terminated with #{Signal.signame(@status.termsig)} signal"
       else
--- a/lib/shipit/command.rb
+++ b/lib/shipit/command.rb
@@ -23,7 +23,7 @@ module Shipit
       end
     end
 
-    attr_reader :out, :code, :chdir, :env, :args, :pid, :timeout
+    attr_reader :out, :chdir, :env, :args, :pid, :timeout
 
     def initialize(*args, default_timeout: Shipit.default_inactivity_timeout, env: {}, chdir:)
       @args, options = parse_arguments(args)
@@ -58,7 +58,7 @@ module Shipit
     end
 
     def exit_message
-      "#{self} exited with status #{@code}"
+      "#{self} #{termination_status}"
     end
 
     def run
@@ -107,16 +107,15 @@ module Shipit
       begin
         read_stream(@out, &block)
       rescue TimedOut => error
-        @code = 'timeout'
         yield red("No output received in the last #{timeout} seconds.") + "\n"
         terminate!(&block)
         raise error
       rescue Errno::EIO # Somewhat expected on Linux: http://stackoverflow.com/a/10306782
       end
 
-      _, status = Process.waitpid2(@pid)
-      @code = status.exitstatus
       self
+    ensure
+      reap_child!
     end
 
     def red(text)
@@ -183,18 +182,18 @@ module Shipit
       rescue TimedOut
       rescue Errno::EIO # EIO is somewhat expected on Linux: http://stackoverflow.com/a/10306782
         # If we try to read the stream right after sending a signal, we often get an Errno::EIO.
-        if status = Process.wait(@pid, Process::WNOHANG)
-          return status
-        else
-          # If we let the child a little bit of time, it solves it.
-          retry_count -= 1
-          if retry_count > 0
-            sleep 0.05
-            retry
-          end
+        if reap_child!(block: false)
+          return true
+        end
+        # If we let the child a little bit of time, it solves it.
+        retry_count -= 1
+        if retry_count > 0
+          sleep 0.05
+          retry
         end
       end
-      Process.wait(@pid, Process::WNOHANG)
+      reap_child!(block: false)
+      true
     end
 
     def kill(sig)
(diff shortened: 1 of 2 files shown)


lib/shipit/command.rb
```diff
@@ -215,5 +214,40 @@ def parse_arguments(arguments)
       end
       return args, options
     end
+
+    def running?
+      !!pid && !@status
+    end
+
+    def code
+      @status&.exitstatus
```
DazWorrall: The `&` guard is to maintain the same API that the previous `attr_reader` provided.

lib/shipit/command.rb
```diff
@@ -183,18 +182,18 @@ def kill_and_wait(sig, wait, &block)
...
-        if status = Process.wait(@pid, Process::WNOHANG)
-          return status
-        else
-          # If we let the child a little bit of time, it solves it.
-          retry_count -= 1
-          if retry_count > 0
-            sleep 0.05
-            retry
-          end
+        if reap_child!(block: false)
```
DazWorrall: This had no tests previously and is very difficult to test for - trying to reap a process not yet ready to exit. I've eyeballed this carefully and expect that I've preserved the existing behaviour, please check my working on this.

lib/shipit/command.rb
```diff
@@ -107,16 +107,15 @@ def stream(&block)
...
         terminate!(&block)
         raise error
       rescue Errno::EIO # Somewhat expected on Linux: http://stackoverflow.com/a/10306782
       end
 
-      _, status = Process.waitpid2(@pid)
-      @code = status.exitstatus
       self
+    ensure
+      reap_child!
```
DazWorrall: `stream` is where `start` is called, hence the blocking reap call.

lib/shipit/command.rb
```diff
@@ -58,7 +58,7 @@ def success?
     end
 
     def exit_message
-      "#{self} exited with status #{@code}"
+      "#{self} #{termination_status}"
```
wvanbergen: Is `self` end up calling `to_s`. Should we call `to_s` directly here?

lib/shipit/command.rb
wvanbergen: Will this work if `start` raises?

lib/shipit/command.rb
```diff
@@ -215,5 +214,40 @@ def parse_arguments(arguments)
...
+    def running?
+      !!pid && !@status
+    end
+
+    def code
+      @status&.exitstatus
+    end
+
+    def signaled?
+      @status.signaled?
```
wvanbergen: Do we need an & here like above?

lib/shipit/command.rb
```diff
@@ -215,5 +214,40 @@ def parse_arguments(arguments)
...
+    end
+
+    def reap_child!(block: true)
+      return @status if @status
+      return unless running? # Command was never started e.g. permission denied, not found etc
+      if block
+        _, @status = Process.waitpid2(@pid)
+      elsif res = Process.waitpid2(@pid, Process::WNOHANG)
```
wvanbergen: Does this return nil/false when the process is still running?

test/unit/command_test.rb
```diff
@@ -66,5 +66,54 @@ class CommandTest < ActiveSupport::TestCase
...
+      refute_predicate command, :running?
+      assert_predicate command.code, :nonzero?
+      assert_equal 'terminated with exit status 1', command.termination_status
+    end
+
+    test 'handles externally signalled commands correctly' do
+      command = Command.new('sleep 10', chdir: '.')
+      t = Thread.new do
+        signalled = false
+        20.times do
```
wvanbergen: why do we need to repeat this?

test/unit/command_test.rb
```diff
@@ -66,5 +66,54 @@ class CommandTest < ActiveSupport::TestCase
...
+      assert_equal 'terminated with KILL signal', command.termination_status
+    end
+
+    test 'reports timedout command correctly' do
+      command = Command.new('sleep 10', chdir: '.', default_timeout: 0.5)
+      assert_raises(Command::TimedOut) { command.run }
+      assert_predicate command, :signaled?
+      refute_predicate command, :running?
+      assert_nil command.code
+      assert_equal 'terminated with INT signal', command.termination_status
```
wvanbergen: Is this the message that eventually gets sent to the user?

I kind of like keeping track whether *we* sent the signal to the process, because that allows us to say that *we* killed the process (due to a timeout). Otherwise I am afraid this will raise questions from people.

lib/shipit/command.rb
DazWorrall: Yes, this is covered with 2 existing tests that cause permission denied and file not found exceptions.

lib/shipit/command.rb
DazWorrall: As this is a new API I decided to leave it unguarded, to catch anything in future calling this while the process is alive. No strong feelings if you prefer defensive here.

lib/shipit/command.rb
DazWorrall: Yes it returns `nil`.

test/unit/command_test.rb
DazWorrall: I want to avoid the thread spinning forever, this limits its life to 2 seconds (20x0.1).

test/unit/command_test.rb
DazWorrall: > Is this the message that eventually gets sent to the user?

It is, but it's not the only output - in the case of a timeout Shipit already prints "No output received in the last _n_ seconds", though I guess that too is ambiguous. Could append that with ", timing out command" or something.

lib/shipit/command.rb
wvanbergen: No, let's not introduce unnecessary defensiveness if it's not needed.

test/unit/command_test.rb
wvanbergen: Why do we need the loop in the first place? Would this not work if we only signal it once?

test/unit/command_test.rb
wvanbergen: For comparison, this is how Shopify Build does this:

- Timeout implementation: https://github.com/Shopify/shopify-build/blob/master/lib/shopify_build/command.rb#L202-L211
- Message: https://github.com/Shopify/shopify-build/blob/master/lib/shopify_build/command.rb#L170-L184

test/unit/command_test.rb
DazWorrall: This will only signal the command once - it breaks immediately after. I added this to protect myself when breaking something else in the code - mistyping a path or causing a loop etc. Without some kind of circuit breaker `thread.join` will never return in cases where the command is not started.

test/unit/command_test.rb
DazWorrall: The timeout model here is a bit different - it's not based on process execution time, but instead time without any output - so it's not directly applicable. However I think I can improve the messaging regardless 👍

lib/shipit/command.rb
DazWorrall: rubocop disagrees 👮‍♀

```
lib/shipit/command.rb:62:10: W: Lint/StringConversionInInterpolation: Use self instead of Object#to_s in interpolation.
      "#{to_s} #{termination_status}"
```

PR #907, opened 2019-05-14
fix(lerna): correctly set the dist tag on publish

gf3:
docs: https://github.com/lerna/lerna/tree/master/commands/publish#--dist-tag-tag

### DIFF
--- a/app/models/shipit/deploy_spec/lerna_discovery.rb
+++ b/app/models/shipit/deploy_spec/lerna_discovery.rb
@@ -83,15 +83,16 @@ module Shipit
 
       def publish_fixed_version_packages
         check_tags = 'assert-lerna-fixed-version-tag'
+        version = lerna_version
         publish = if lerna_lerna >= LATEST_MAJOR_VERSION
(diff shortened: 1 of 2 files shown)


app/models/shipit/deploy_spec/lerna_discovery.rb
```diff
@@ -83,15 +83,16 @@ def publish_independent_packages
 
       def publish_fixed_version_packages
         check_tags = 'assert-lerna-fixed-version-tag'
+        version = lerna_version
```
GoodForOneFare: Isn't this always going to be `fixed`?

GoodForOneFare: Oh, this is fixed version logic, eh.  Nevermind 😄

gf3: @GoodForOneFare with fixed versioning the `version` key is the actual current version for all packages (e.g. `"version": "1.3.1-alpha.2"`)

PR #909, opened 2019-06-11
Add lock_level property to stacks

mdeloupy:
This allows lock messages to be displayed as warnings in the banner without actually
locking deploys.

### DIFF
--- a/app/controllers/shipit/api/locks_controller.rb
+++ b/app/controllers/shipit/api/locks_controller.rb
@@ -5,21 +5,23 @@ module Shipit
 
       params do
         requires :reason, String, presence: true
+        accepts :lock_level, String
       end
       def create
         if stack.locked?
           render json: {message: 'Already locked'}, status: :conflict
         else
-          stack.lock(params.reason, current_user)
+          stack.lock(params.reason, current_user, lock_level: params.lock_level)
           render_resource stack
         end
       end
 
       params do
         requires :reason, String, presence: true
+        accepts :lock_level, String
       end
       def update
-        stack.lock(params.reason, current_user)
+        stack.lock(params.reason, current_user, lock_level: params.lock_level)
         render_resource stack
       end
 
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -23,6 +23,9 @@ module Shipit
     ENVIRONMENT_MAX_SIZE = 50
     REQUIRED_HOOKS = %i(push status).freeze
 
+    LOCK_LEVELS = %w(enforced advisory).freeze
+    enum lock_level: LOCK_LEVELS.zip(LOCK_LEVELS).to_h
+
     has_many :commits, dependent: :destroy
     has_many :pull_requests, dependent: :destroy
     has_many :tasks, dependent: :destroy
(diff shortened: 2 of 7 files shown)


app/controllers/shipit/api/locks_controller.rb
```diff
@@ -5,21 +5,23 @@ class LocksController < BaseController
 
       params do
         requires :reason, String, presence: true
+        accepts :lock_level, String
```
byroot: You can validates the strings directly here, this would generate better API errors.

app/models/shipit/stack.rb
```diff
@@ -386,16 +389,21 @@ def active_task
...
+    end
+
+    def soft_locked?
       lock_reason.present?
     end
 
-    def lock(reason, user)
-      params = {lock_reason: reason, lock_author: user}
+    def lock(reason, user, lock_level: 'enforced')
+      lock_level ||= 'enforced'
```
mdeloupy: What's the best way to set a default value for `lock_level` ? I used a keyword argument since it's a more optional argument, but could've used `def lock(reason, user, lock_level = 'enforced')` also. I also had to add the variable assignation above since the argument could either be present or sent as `nil`, and we don't want a `nil` value. But this looks redundant.

What do you think ?

test/dummy/db/schema.rb
```diff
@@ -185,9 +185,9 @@
     t.bigint "github_id"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.index ["commit_id", "github_id"], name: "index_deploy_statuses_on_commit_id_and_github_id"
-    t.index ["stack_id", "commit_id"], name: "index_deploy_statuses_on_stack_id_and_commit_id"
-    t.index ["user_id"], name: "index_deploy_statuses_on_user_id"
+    t.index ["commit_id", "github_id"], name: "index_release_statuses_on_commit_id_and_github_id"
+    t.index ["stack_id", "commit_id"], name: "index_release_statuses_on_stack_id_and_commit_id"
+    t.index ["user_id"], name: "index_release_statuses_on_user_id"
```
mdeloupy: I'm not sure what caused these changes 🤔.

byroot: Can't even tell what the diff actually is. Must be a minor change in Rails schema dump.

app/models/shipit/stack.rb
```diff
@@ -386,16 +389,21 @@ def active_task
     end
 
     def locked?
+      lock_reason.present? && lock_level != 'advisory'
```
byroot: I'd constantize that string, it kinda makes sure there's no typo.

app/models/shipit/stack.rb
byroot: Same here I'd reference a constant.

As for the keyword arg, yeah it's fine like this.

app/controllers/shipit/api/locks_controller.rb
mdeloupy: Your gem is awesome !

app/models/shipit/stack.rb
```diff
@@ -386,16 +391,21 @@ def active_task
     end
 
     def locked?
+      lock_reason.present? && lock_level != LOCK_LEVEL_ADVISORY
+    end
+
+    def soft_locked?
```
eapache: It's not clear to me without looking at the implementation whether `soft_locked?` means "locked, and level == advisory", or just "locked, and level is at least advisory".

mdeloupy: I'm going to rename the method to `advisory_locked?`.

PR #910, opened 2019-06-18
Ensure that the Abort & Rollback button does not rollback to faulty deploys.

ajshepley:
## Issue

Currently, when the `Abort & Rollback` button is used, it triggers a revert path that will roll back to the immediately previous deploy.

However, it's possible that the previous deploy(s) are themselves faulty. As a common use case of rolling back is to bail out of a deploy and to go back to a prior success, it's more desirable to roll back to the nearest prior successful deploy/commit.

This isn't an issue for manual rollbacks as we only expose the rollback button on valid prior deploys to begin with, and it's next to each individual deploy, making it unambiguous as to what you would be rolling back to.

---

## Changes

This PR changes the Abort & Rollback button (and its underlying trigger_revert method) to skip unsuccessful deploys when selecting the `until_commit` to roll back to.

To make the overall behaviour more obvious to the user, I've also added text to the task output page next to the Abort & Rollback button, which shows and links the `From...To` commit range that would be deployed if the button is pressed.

This should help to resolve https://github.com/Shopify/shipit/issues/577

### DIFF
--- a/app/controllers/shipit/deploys_controller.rb
+++ b/app/controllers/shipit/deploys_controller.rb
@@ -5,6 +5,7 @@ module Shipit
     before_action :load_stack
     before_action :load_deploy, only: %i(show rollback revert)
     before_action :load_until_commit, only: :create
+    helper_method :short_commit_sha
 
     def new
       @commit = @stack.commits.by_sha!(params[:sha])
@@ -40,6 +41,12 @@ module Shipit
       redirect_to rollback_stack_deploy_path(@stack, previous_deploy)
     end
 
+    def short_commit_sha(task)
+      if previous_successful_deploy(task)
+        @short_commit_sha ||= @previous_successful_deploy.until_commit&.short_sha
+      end
+    end
+
     private
 
     def load_deploy
@@ -56,6 +63,10 @@ module Shipit
 
     def deploy_params
       @deploy_params ||= params.require(:deploy).permit(:until_commit_id, env: @stack.deploy_variables.map(&:name))
+    end
+
+    def previous_successful_deploy(task)
+      @previous_successful_deploy ||= task&.previous_successful
     end
   end
 end
--- a/app/helpers/shipit/github_url_helper.rb
+++ b/app/helpers/shipit/github_url_helper.rb
@@ -55,5 +55,11 @@ module Shipit
       text = deploy.commit_range.map(&:short_sha).join('...')
       link_to(text, url, class: 'number')
     end
+
+    def link_to_github_range(stack, since_commit, until_commit)
+      url = github_commit_range_url(stack, since_commit, until_commit)
+      text = [since_commit, until_commit].map(&:short_sha).join('...')
+      link_to(text, url, class: 'number')
+    end
   end
 end
--- a/app/helpers/shipit/github_url_helper.rb
+++ b/app/helpers/shipit/github_url_helper.rb
@@ -55,11 +55,5 @@ module Shipit
       text = deploy.commit_range.map(&:short_sha).join('...')
       link_to(text, url, class: 'number')
     end
-
-    def link_to_github_range(stack, since_commit, until_commit)
-      url = github_commit_range_url(stack, since_commit, until_commit)
-      text = [since_commit, until_commit].map(&:short_sha).join('...')
-      link_to(text, url, class: 'number')
-    end
   end
 end
--- a/app/models/shipit/task.rb
+++ b/app/models/shipit/task.rb
@@ -47,6 +47,10 @@ module Shipit
         completed.last
       end
 
+      def previous_successful(deploy_id)
+        success.where("id < #{deploy_id}").last
+      end
+
       def current
         active.exclusive.last
       end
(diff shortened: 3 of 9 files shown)


app/controllers/shipit/deploys_controller.rb
```diff
@@ -40,6 +41,16 @@ def revert
       redirect_to rollback_stack_deploy_path(@stack, previous_deploy)
     end
 
+    def rollback_until_commit(task)
+      if previous_successful_deploy(task)
+        @rollback_until_commit ||= Commit.where(
+          stack_id: @previous_successful_deploy.stack_id,
+          id: @previous_successful_deploy.until_commit_id,
+        ).first
```
JackTLi: could you do `@previous_successful_deploy.until_commit` here?

ajshepley: I think that'd work, but I'm replacing this method altogether now.

app/controllers/shipit/deploys_controller.rb
```diff
@@ -40,6 +41,12 @@ def revert
       redirect_to rollback_stack_deploy_path(@stack, previous_deploy)
     end
 
+    def short_until_commit(task)
```
DazWorrall: Nit: `commit` is a real model, this returns the sha of a matched commit as a string so `short_commit_sha`?

app/models/shipit/task.rb
```diff
@@ -47,6 +47,10 @@ def last_completed
         completed.last
       end
 
+      def previous_successful(id)
+        success.where('id < ?', id).last
```
DazWorrall: ActiveRecord question: is `last` doing this in the SQL, or are we returning the whole dataset here and discarding everything but the last element? Also can we trust implicit ordering here?

app/models/shipit/task.rb
```diff
@@ -231,6 +235,10 @@ def supports_rollback?
       false
     end
 
+    def previous_success
```
DazWorrall: Are this  and `previous_successful` on the class too similarly named? Could they be the same?

app/models/shipit/task.rb
```diff
@@ -231,6 +235,10 @@ def supports_rollback?
...
+      Task.previous_successful(id)
```
casperisfine: Might want to use `self.class.previous...` here, so that `Deploy#previous_success` won't return raw `Task` instances.

app/models/shipit/task.rb
JackTLi: last should add a `DESC LIMIT 1` to the query 🙂

app/helpers/shipit/github_url_helper.rb
```diff
@@ -55,5 +55,11 @@ def link_to_github_deploy(deploy)
       text = deploy.commit_range.map(&:short_sha).join('...')
       link_to(text, url, class: 'number')
     end
+
+    def link_to_github_range(stack, since_commit, until_commit)
```
JackTLi: This looks unused now, can we 🔥 ?

app/models/shipit/task.rb
ajshepley: Ah, thank you for this. I am still a bit green on Ruby's class system and the specifics of its polymorphism.

app/models/shipit/task.rb
DazWorrall: Any idea about the ordering Jack? Don't want this to accidentally return a really old deploy.

JackTLi: I think this should work reliably, due to the nature of shipit creating new tasks on rollbacks/deploys. If anything, the success filter will give us a little more security as well. I don't anticipate any problems with this, and we've been using similar ID ordering in many other parts of shipit which have worked reliably

ajshepley: `.last` gets baked into the SQL as a limit clause, for example, this _previous_successful_ call on a Task with id 197:
```sql
SELECT  "tasks".* FROM "tasks" WHERE "tasks"."type" IN ('Shipit::Deploy') AND "tasks"."status" = ? AND (id < 197) ORDER BY "tasks"."id" DESC LIMIT ?  [["status", "success"], ["LIMIT", 1]]
```
So it should be fine in that respect.

The ordering question is interesting. The default ordering by id should be fine (since we're already using) but most best practices blogs I've found recommend being explicit about order, and often, recommend using created_at and similar fields. However, in this context I don't see created_at being better or worse than the incrementing id.

I think the approach of being consistent and explicit makes sense, so adding an explicit ordering could be fine. but there are other places where we use .first and .last without explicitly ordering, so I'm not sure if that would be consistent.

What are your thoughts?

DazWorrall: If it's working reliably elsewhere then we don't need to go against the grain 👍

app/models/shipit/task.rb
ajshepley: They can be named the same, yeah. I wasn't sure if there would be problems in having similarly named class and instance methods, but now that I have tests, it's simple to experiment and see that it works.

app/helpers/shipit/github_url_helper.rb
ajshepley: Weird, I could have sworn I deleted it. Maybe I ran a reset and took it with it. Will remove.

PR #912, opened 2019-06-20
Remove commit count warning

mctaylorpants:
Hello from Clio! 👋 

This PR removes the commit count warning that appears on the Merge Status views:

Screen Shot 2019-06-21 at 08 02 12

The link appears to lead to an internal document which is publicly inaccessible (I believe [this](https://gist.github.com/casperisfine/0356ab7e809f4dc632809727fcb952d0) is the content of the document). Since the policy appears to be related to Shopify itself, I believe it makes sense to remove this from the open-source version to avoid confusion.

### DIFF
--- a/app/helpers/shipit/merge_status_helper.rb
+++ b/app/helpers/shipit/merge_status_helper.rb
@@ -1,7 +1,0 @@
-module Shipit
-  module MergeStatusHelper
-    def too_many_commits?(commits)
-      commits > 4
-    end
-  end
-end
--- a/app/helpers/shipit/merge_status_helper.rb
+++ b/app/helpers/shipit/merge_status_helper.rb
@@ -1,0 +1,7 @@
+module Shipit
+  module MergeStatusHelper
+    def display_commit_count_warning?(commits)
+      commits > 4 && @stack.merge_queue_enabled?
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


JackTLi:
👋 Thanks for bringing this up!

What do you think about removing just the "Find out why" section and keeping the " Consider rebasing into a smaller number of commits containing atomic changes." copy?

mctaylorpants:
@JackTLi  That could work too! My reasoning for removing it entirely was because it might not apply to everyone's workflows - for example, we have squash & merge enabled as an option, and so the warning may be redundant if a developer is planning to squash commits after review.

If we decided to keep the message, I believe the more detailed explanation is still quite useful, but maybe we could update it to point to a public gist link instead. Let me know what you think!

JackTLi:
That's useful context! We don't use Squash & Merge on many projects so we haven't considered that case. 

Here's an idea, we can:
1. Remove the shopify specific gist (thanks again for catching this)
2. Display the message only if the merge queue is enabled (the merge queue only merges, no squashing for this guy)
```ruby
<%= render 'commit_count_warning' if too_many_commits?(params[:commits].to_i) && @stack.merge_queue_enabled? %>
```

Does that sound good to you?

mctaylorpants:
I like it! 👍 

d5f2391 removes the link to the gist, and refactors the helper method to check number of commits _and_ whether merge status is enabled.

PR #913, opened 2019-06-25
Fix incorrect stack not being skipped when selecting deploy commit for reverts.

ajshepley:
With PR https://github.com/Shopify/shipit-engine/pull/910, the Abort&Rollback button (eventually `Deploy.trigger_revert`) would try to skip unsuccessful tasks when searching for a commit to roll back to.

However, a critical case was missed. When multiple stacks are interleaving their deploys/commits, the codepath used was not correctly filtering out unrelated stacks. Unfortunately, the codepath used by the tests would end up with the correct context, and was correctly filtering.

This PR fixes that problem by removing the Task methods altogether, as it makes more sense for Deploys (and UI controllers) to ask the Stack about what the most recent successful deploy was on said stack prior to a given id. This implicitly filters to deploys _of_ that stack.

In addition to this, a UI bug was fixed where the prospective rollback sha was always shown, even when a rollback option was not present. Instead, the element will now be hidden where appropriate, as seen here:

deploy without sha

rollback with sha

I adjusted the controller test to verify that the text is shown at the correct times. I adjusted the span class to make this test as robust as possible.

Lastly, I've added a regression test that interleaves an unrelated stack's deploy and ensures that it (and its commits) are not selected when reverting.

### DIFF
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -285,7 +285,7 @@ module Shipit
     end
 
     def previous_successful_deploy(deploy_id)
-      deploys_and_rollbacks.previous_successful(deploy_id)
+      deploys_and_rollbacks.success.where("id < ?", deploy_id).last
     end
 
     def last_active_task
--- a/test/models/deploys_test.rb
+++ b/test/models/deploys_test.rb
@@ -538,6 +538,57 @@ module Shipit
       assert_equal commit_ids[1], test_stack.last_deployed_commit.id
     end
 
+    test "#trigger_revert skips deploys from other stacks" do
+      # The revert functionality should only consider Shipit::Deploy and Shipit::Rollback's from the same stack when selecting a target to roll back to for the user.
+      # But deploys and commits from other stacks can have ids between the current and the previous of this stack, so we want a test to ensure that the deploy
+      # from the correct stack is selected.
+      user_id = @user.id
+      test_stack = create_test_stack
+      test_stack.save
+      other_stack = create_test_stack
+      other_stack.repo_name += "_other"
+      other_stack.save
+      other_stack.reload
+      stack_id = test_stack.id
+
+      # Create valid commit history for the stack. We need several commits to deploy and roll back through.
+      commit_ids = generate_commits(amount: 4, stack_id: stack_id, user_id: user_id, validate: true)
+      commit_ids.each { |commit_id| create_test_status(commit_id: commit_id, stack_id: stack_id, state: "success").save }
+
+      # We want the following order of Deploys:
+      # 1. Success (commits 1-2)
+      # 2. Success, but belongs to a different stack (commits 2-3)
+      # 3. Running (commits 3-4)
+      # 4. Reversion of the running deploy to the last successful deploy of the same stack. (-> commits 1-2, i.e. the successful deploy.)
+      # If the revert functionality doesn't restrict to the correct stack, then commit 3 will be latest deployed when the reversion is done.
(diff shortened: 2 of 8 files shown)


ajshepley:
I'll squash the commits tomorrow, but they are fairly well separated for now - the first commit is likely the only redundant one.

test/models/deploys_test.rb
```diff
@@ -538,6 +538,57 @@ def generate_commits(amount:, stack_id:, user_id:, validate:)
       assert_equal commit_ids[1], test_stack.last_deployed_commit.id
     end
 
+    test "#trigger_revert skips deploys from other stacks" do
+      # The revert functionality should only consider Shipit::Deploy and Shipit::Rollback when selecting a target to roll back to for the user.
+      # But it is possible for other task types to be defined, so we want to ensure that they are properly skipped, as we can't know whether they are 'valid' to roll back to.
+      user_id = @user.id
+      test_stack = create_test_stack
+      test_stack.save
+      test_stack.reload
```
DazWorrall: Is this reload necessary?

ajshepley: Ah, I don't seem to re-use test_stack below this point, so it looks like it can be safely removed.

JackTLi:
Can we put the "to `12345678`" text into the button so that the whole button is "Abort and rollback to 12345678"? The separation is a little awkward since this is tied to the actions of that single button

app/models/shipit/stack.rb
```diff
@@ -285,7 +285,7 @@ def last_completed_deploy
     end
 
     def previous_successful_deploy(deploy_id)
-      deploys_and_rollbacks.previous_successful(deploy_id)
+      deploys_and_rollbacks.success.where("id < ?", deploy_id).last
```
JackTLi: this query will have the `stack` scope now?

ajshepley: I believe so. All of my testing seems to show it as such, since is executing within the context of the relation to this stack.
This is the query generated (both by page load and by manually calling this method):
`Shipit::Task Load (0.3ms)`
```sql
 SELECT  "tasks".* FROM "tasks" WHERE "tasks"."stack_id" = ? AND "tasks"."type" IN (?, ?) AND "tasks"."status" = ? AND (id < 199) ORDER BY "tasks"."id" DESC LIMIT ?  [["stack_id", 4], ["type", "Shipit::Deploy"], ["type", "Shipit::Rollback"], ["status", "success"], ["LIMIT", 1]]
```

ajshepley:
@JackTLi 

Having the sha in the button is what I originally started with, but it looks a bit weird (both with and without formatting) so I wasn't sure if it was what we wanted.

If either of these look better to you, I can go with them - let me know what you prefer:

Formatted sha:

in button 1

Unformatted sha:

in button 2

app/models/shipit/stack.rb
JackTLi: 👍

JackTLi:
I wonder if we could have the serif fonting without the background. Or, the unformatted sha is fine as well. Since this functionality is tooling that people use in emergencies, we should be as direct and reduce as much ambiguity as possible (having the seperation makes it unclear that the "to ..." only applies to the Abort and Rollback button

ajshepley:
I can try to style the text similarly without the background. The CSS seems nicely separated.

ajshepley:
@JackTLi It should be good to go now. The in-button SHA is now styled similarly but without color+background change:

adjusted button sha

PR #915, opened 2019-06-28
Optionally log command output

joeldrapper:
Optionally log command output to `Shipit.task_logger` with context on the current repo / deployment.

Because an output chunk can contain more or less than one line, we first buffer the output into whole lines and then format it with the current context before sending it on to the configured task logger.

### DIFF
--- a/app/models/shipit/task.rb
+++ b/app/models/shipit/task.rb
@@ -186,6 +186,7 @@ module Shipit
     end
 
     def write(text)
+      log_output(text)
       chunks.create!(text: text)
     end
 
@@ -332,6 +333,16 @@ module Shipit
 
     def abort_key
       "#{status_key}:aborting"
+    end
+
+    def log_output(text)
+      output_line_buffer.buffer(text) do |line|
+        Shipit.task_logger.info("[#{stack.repo_name}##{id}] #{line}")
+      end
+    end
+
+    def output_line_buffer
+      @output_line_buffer ||= LineBuffer.new
     end
   end
 end
--- a/lib/shipit/line_buffer.rb
+++ b/lib/shipit/line_buffer.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+module Shipit
+  class LineBuffer
+    SEPARATOR = "\n"
+
+    def initialize(queue = "")
+      @queue = queue.dup
+    end
+
+    def buffer(text, &block)
+      @queue << text
+      whole_lines.each(&block).tap { flush }
+    end
+
+    def empty?
+      @queue.empty?
+    end
+
+    private
+
+    def whole_lines
+      whole? ? lines : lines[0..-2]
+    end
+
+    def flush
+      whole? ? clear : @queue = lines.last
+    end
+
+    def whole?
+      @queue.end_with?(SEPARATOR)
+    end
+
+    def lines
+      @queue.split(SEPARATOR)
+    end
+
+    def clear
+      @queue.clear
+    end
+  end
+end
--- a/app/jobs/shipit/perform_task_job.rb
+++ b/app/jobs/shipit/perform_task_job.rb
@@ -18,6 +18,7 @@ module Shipit
       @task.run!
       checkout_repository
       perform_task
+      @task.write("\nCompleted successfully\n")
       @task.report_complete!
     rescue Command::TimedOut => error
       @task.write("\n#{error.message}\n")
--- a/lib/shipit.rb
+++ b/lib/shipit.rb
@@ -49,6 +49,7 @@ require 'shipit/environment_variables'
 require 'shipit/stat'
 require 'shipit/strip_cache_control'
 require 'shipit/cast_value'
+require 'shipit/line_buffer'
 
 SafeYAML::OPTIONS[:default_mode] = :safe
 SafeYAML::OPTIONS[:deserialize_symbols] = false
@@ -59,7 +60,7 @@ module Shipit
   delegate :table_name_prefix, to: :secrets
 
   attr_accessor :disable_api_authentication, :timeout_exit_codes
-  attr_writer :internal_hook_receivers
+  attr_writer :internal_hook_receivers, :task_logger
 
   self.timeout_exit_codes = [].freeze
 
@@ -193,6 +194,10 @@ module Shipit
 
   def internal_hook_receivers
     @internal_hook_receivers ||= []
+  end
+
+  def task_logger
+    @task_logger ||= Logger.new(nil)
   end
 
   protected
--- a/test/models/tasks_test.rb
+++ b/test/models/tasks_test.rb
@@ -17,5 +17,18 @@ module Shipit
       task = shipit_tasks(:shipit_with_title_parsing_issue)
       assert_equal 'This task (title: Using the %{WRONG_VARIABLE_NAME}) cannot be shown due to an incorrect variable name. Check your shipit.yml file', task.title
     end
+
+    test "#write sends line-buffered output to task logger" do
+      task = shipit_tasks(:shipit)
+
+      mock_task_logger = mock.tap do |m|
+        m.expects(:info).with("[shipit-engine#1] hello").once
+        m.expects(:info).never
+      end
+
+      Shipit.stubs(:task_logger).returns(mock_task_logger)
+
+      task.write("hello\nworld")
+    end
   end
 end
--- a/test/unit/line_buffer_test.rb
+++ b/test/unit/line_buffer_test.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+module Shipit
+  class LineBufferTest < ActiveSupport::TestCase
+    setup { @buffer = LineBuffer.new }
+
+    test "buffers partial lines" do
+      refute_predicate(@buffer.buffer("a"), :any?)
+      assert_equal(["a"], @buffer.buffer("\n").to_a)
+      assert_predicate(@buffer, :empty?)
+    end
+
+    test "splits up multiple lines" do
+      assert_equal(%w(a b), @buffer.buffer("a\nb\n").to_a)
+      assert_predicate(@buffer, :empty?)
+    end
+  end
+end


joeldrapper:
@DazWorrall @byroot I’m not sure how I should attach more context here, such as the repo / deployment id.

Also, @DazWorrall suggested this output should be optional, but I’m not sure how best to do that. Are there any other configurable options implemented here that I can look at?

casperisfine:
@joeldrapper I think for both of your question, the solution is the same.

Doing this logging in `Command` is not the best place because it's a quite low level class which doesn't know much about what it's doing.

Ultimately the output is registered here: https://github.com/Shopify/shipit-engine/blob/c73aaed110aef8fe08e7452f6e2dd8301671da9a/app/models/shipit/task.rb#L184-L186

That's a much better hook point in my opinion since your are inside the `Task`, so you know what stack & task you belong to.

As for making it opt-in / configurable, I'd say:

```ruby
def write(text)
  Shipit.task_logger.info(text)
  chunks.create!(text: text)
end
```

This way it's a config people can set in `application.rb` or `config/environments/*.rb`. It should default to `Logger.new(nil)`.

joeldrapper:
I originally looked at implementing this as a callback on `OutputChunks`. @DazWorrall suggested that these should be written by the actual job process and not any web process for a few reasons: 

> 1.  the chunks aren’t line buffered, which splunk cares about
> 2.  the web processes will interleave their own output
> 3.  if we echo it async we lose ordering guarantees

I guess we'd have the same problems using `Task#write`.

I love the idea of using a null logger by default. 👌

casperisfine:
> these should be written by the actual job process and not any web process

`Task#write` is only called from jobs, never from web workers.

As for the buffering etc concerns, that is a deployment concern. If you need log buffering, then simply set `Shipit.task_logger` to some buffered logger implementation.

DazWorrall:
@casperisfine I was concerned about _line_ buffering - chunks are (afaics) byte streams, and so could contain a partial line, or multiple lines. I didn't expect piping chunks straight to the logger would result in very legible output.

casperisfine:
> chunks are (afaics) byte streams, and so could contain a partial line, or multiple lines

Yes they are. And since Shipit use a PTY, many command will write partial lines, clear the line etc.

joeldrapper:
> Yes they are. And since Shipit use a PTY, many command will write partial lines, clear the line etc.

joeldrapper:
Can we just make a logger that handles line buffering?

casperisfine:
Yes probably, and it's likely the best solution.

joeldrapper:
I don't think it's obvious that `Shipit.task_logger` itself should line buffered, so I wonder if we should decorate it from `Shipit::Task`, i.e.

```ruby
def task_logger
  @task_logger ||= LineBufferedLogger.new(Shipit.task_logger)
end
```

casperisfine:
I'm fine with that.

app/models/shipit/task.rb
```diff
@@ -333,5 +334,17 @@ def status_key
     def abort_key
       "#{status_key}:aborting"
     end
+
+    def log_output(text)
+      output_log_line_buffer.buffer(text) do |line|
+        Shipit.task_logger.info(
+          "[#{stack.repo_name}] #{line}"
```
joeldrapper: I’m not sure what's the best way to get the deploy id here. Is it safe to assume that if the task doesn't have a `deploy`, then it _is_ a deploy? i.e. `deploy ? deploy.id : id`?

casperisfine: `Rollback < Deploy < Task`. So all `Deploy` are `Tasks` but the reverse isn't true.

But anyway, you don't really care wether it's a task or a deploy, in both case you want the record ID, so `id`.

joeldrapper: Thanks.

lib/shipit/line_buffer.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module Shipit
+  class LineBuffer
+    SEPARATOR = "\n"
+
+    def initialize(value = "")
+      @value = value
```
casperisfine: maybe `@buffer` would be more telling than `@value`?

lib/shipit/line_buffer.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+
+    def buffer(text, &block)
+      @value += text
```
casperisfine: Might not matter in the grand scheme of things, but mutating the string with `<<` is likely much more performant that constant concatenation.

lib/shipit/line_buffer.rb
```diff
@@ -0,0 +1,38 @@
...
+      whole_lines.each(&block).tap { flush }
```
casperisfine: Why the `.tap { }` ? Seems like flushing on the next line would be just fine.

lib/shipit/line_buffer.rb
```diff
@@ -0,0 +1,38 @@
...
+    def flush
+      whole? ? clear : @value = lines.last
+    end
+
+    def whole?
+      @value.end_with?(SEPARATOR)
+    end
+
+    def lines
+      @value.split(SEPARATOR).reject(&:empty?)
```
casperisfine: What's the reason for skipping empty lines? They might be somewhat meaningful.

lib/shipit/line_buffer.rb
```diff
@@ -0,0 +1,38 @@
...
+    def whole_lines
+      whole? ? lines : lines[0..-2]
+    end
+
+    def flush
+      whole? ? clear : @value = lines.last
+    end
+
+    def whole?
+      @value.end_with?(SEPARATOR)
```
DazWorrall: What will happen if the output does not contain a trailing newline (try `echo -n hello` in your shell). Will we lose the buffered output?

casperisfine: It should probably be entirely flushed when the task ends.

lib/shipit/line_buffer.rb
joeldrapper: In order to return the Enumerator.

lib/shipit/line_buffer.rb
joeldrapper: I didn't see the point in logging blank lines, but I can remove this if you think they would be meaningful.

lib/shipit/line_buffer.rb
joeldrapper: I thought the command output always ended with a new line. If it's interrupted, the half line would disappear with the line buffer when it's garbage collected. One option here if you think we're going to get important half lines is to flush it entirely when the task ends.

lib/shipit/line_buffer.rb
joeldrapper: I had that originally, but the class is a buffer, so it having a buffer value didn't feel right.

lib/shipit/line_buffer.rb
casperisfine: > if you think we're going to get important half lines

When a command crash, the last line is often the most important one.

lib/shipit/line_buffer.rb
casperisfine: > I didn't see the point in logging blank lines

They usually improve readability. I'd rather have that logger as dumb as possible, as in it should chose what or what not to log.

lib/shipit/line_buffer.rb
casperisfine: It's very common though. Up to you, but I generally find names like "value" or "data" very undescriptive.

lib/shipit/line_buffer.rb
joeldrapper: Good catch! Yes, I used `+=` because `@value` was a frozen string literal by default. Changing that by duplicating it in the initialiser.

lib/shipit/line_buffer.rb
joeldrapper: The problem is it's not a `@buffer`, it's a `String` that holds the current value of a `LineBuffer`, so it's very difficult to name. `value` felt better than `string` or `buffer` or `data`, but it still doesn't feel right.

joeldrapper: It's a queue! That's it. I’m going with `@queue`.

joeldrapper: Thanks for picking up on this @casperisfine. 🙏  `@queue` sparks joy.

lib/shipit/line_buffer.rb
joeldrapper: What's the best way to catch the task ending for any reason i.e. failed, aborted, success, etc?

casperisfine: It happens here: https://github.com/Shopify/shipit-engine/blob/73897c9268235ceb03ce8e335ebe5d931297275c/app/jobs/shipit/perform_task_job.rb#L17-L33

So the various `report_*` methods. Might want to call some flush in them.

joeldrapper: So it looks like all cases apart from `complete` already write log messages with new lines.

joeldrapper: There are a few options I’m considering here. I’d like to know what you think.

1. Write a final log message in all cases, ensuring a final whole line is sent to the buffer. Currently, `complete` is the only case missing a final log message. All the other cases are covered as far as I can tell.
2. Add an `ensure` clause to `PerformTaskJob#run` that reports it as `finished` i.e. the job is done, whether it's completed, timed out, erred, or anything else. Then connect that to a method that resolves the buffer.
3. Resolve the buffer directly from each case in `PerformTaskJob#run` (all the `report_*` methods @casperisfine mentioned).
4. Resolve the buffer on an `after_transition` callback in `Task`.

DazWorrall: We should try to keep the command output contiguous. Option 4 is harder to reason about in terms of ordering, 1-3 all seem fine. Would adding an ensure block [here](https://github.com/Shopify/shipit-engine/blob/73897c9268235ceb03ce8e335ebe5d931297275c/app/jobs/shipit/perform_task_job.rb#L76) also work?

```ruby
def capture!
...
ensure
  @task.write('\n') # move this down from the end of the main block
  # or @task.output_line_buffer.flush 
  # or something something
end
```

joeldrapper: Thanks @DazWorrall! We can't consistently resolve the buffer by simply writing `\n` since we no longer ignore blank lines. I’m going forward with option 1 for now, since it's the simplest. If we feel like that's not right, I’m more than happy to revisit option 2 or something like that another day.

PR #916, opened 2019-07-02
Use previous rollback behaviour when no previous deploy is found to roll back to.

ajshepley:
With PR #913, the Abort&Rollback button (eventually Deploy.trigger_revert) would try to skip unsuccessful tasks when searching for a commit to roll back to, and fix a bug where interleaved stack deploys would be used for the revert sha instead of the same stack.

Unfortunately, an edge case (no prior deploys found) was missed, resulting in a rare but blocking error.

This PR fixes the above problem by deferring to the old rollback behaviour whenever a previous-deploy could not be found. At the same time, I've made sure that the SHA shown in the button will be correct (showing the `current_deploy_task.until_commit` in the aforementioned scenario).

For context, see also:
https://github.com/Shopify/shipit-engine/pull/910
https://github.com/Shopify/shipit-engine/pull/913

### DIFF
--- a/app/models/shipit/task.rb
+++ b/app/models/shipit/task.rb
@@ -312,6 +312,16 @@ module Shipit
       Shipit::Engine.routes.url_helpers.stack_task_url(stack, self)
     end
 
+    def commit_to_rollback_to
+      previous_deployed_commit = stack.previous_successful_deploy_commit(id)
+
+      if previous_deployed_commit == Shipit::Stack::NoDeployedCommit
+        since_commit
+      else
+        previous_deployed_commit
+      end
+    end
+
     private
 
     def prevent_concurrency
--- a/app/controllers/shipit/deploys_controller.rb
+++ b/app/controllers/shipit/deploys_controller.rb
@@ -66,7 +66,7 @@ module Shipit
     end
 
     def previous_successful_deploy_commit(task)
-      @previous_successful_deploy_commit ||= @stack.previous_successful_deploy_commit(task&.id)
+      @previous_successful_deploy_commit ||= task.commit_to_rollback_to
     end
   end
 end
--- a/app/models/shipit/deploy.rb
+++ b/app/models/shipit/deploy.rb
@@ -87,14 +87,14 @@ module Shipit
 
     # Rolls the stack back to the most recent **previous** successful deploy
     def trigger_revert(force: false)
-      previous_successfully_deploy_commit = stack.previous_successful_deploy_commit(id)
+      previous_successful_commit = commit_to_rollback_to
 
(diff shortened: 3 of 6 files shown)


app/models/shipit/task.rb
```diff
@@ -312,6 +312,16 @@ def permalink
       Shipit::Engine.routes.url_helpers.stack_task_url(stack, self)
     end
 
+    def commit_to_revert_to
+      previous_deployed_commit = stack.previous_successful_deploy_commit(id)
+
+      if !previous_deployed_commit || previous_deployed_commit == Shipit::Stack::NoDeployedCommit
```
JackTLi: Any reason why we can't just do `previous_deployed_commit || since_commit` here?

ajshepley: previous_deployed_commit may be an instance of NoDeployedCommit.

I am amiable to the idea of just not using NoDeployedCommit for any of this stuff, if you think that's a good idea; it would simplify things to just nil-logic which is nicer IMO.

JackTLi: ah right.

Can we get rid of the `!previous_deployed_commit` case then? Or is this case expected sometimes too?

ajshepley: Ah, I see that NoDeployedCommit overrides `blank?` so it may work for the logic you described already.

ajshepley: If I remove NoDeployedCommit then this can become just
```ruby
def commit_to_revert_to
      stack.previous_successful_deploy_commit(id) || since_commit
end
```

ajshepley: Latest deploy simplifies the conditional to just checking for `NoDeployedCommit`. I've rebased that into the first commit.

app/models/shipit/task.rb
```diff
@@ -312,6 +312,16 @@ def permalink
       Shipit::Engine.routes.url_helpers.stack_task_url(stack, self)
     end
 
+    def commit_to_revert_to
```
DazWorrall: `commit_to_rollback_to`? `revert` implies git things.

ajshepley: We already use `trigger_revert` in deploy.rb so I was re-using the same wording. But `rollback_to` does reads a bit better.

ajshepley: @JackTLi What do you think makes more sense here?

ajshepley: I've gone with `_rollback` as it's a bit more explicit and `Task` has no notion of 'revert' inside it anywhere.

JackTLi: 👍 to rollback over revert

PR #917, opened 2019-07-02
Set a git ref when a deploy is successful.

ajshepley:
To support future features, we want to add a per-environment per-project git reference that will allow new tools and features to know which commit is currently deployed by making a Github `refs` api call.

See https://github.com/Shopify/shipit/issues/743 for more context.

The basic gist of these changes is that when a deploy finishes, its `-> success` state transition will trigger a call to its owning Stack, which will enqueue a job that will update or create a ref in github pointing to whatever the Stack thinks is its latest deployed commit.

For now, we've assumed that it is sufficient to abide by a ref set at the end of a deploy, and not add things like startup checks by the stacks the ensure the currently set ref is correct (and so on). The idea being that it is fine to let an assumedly-rare edge case of stale or incorrect last deployed refs self-correct via deploys or manual intervention for now.

This should serve to resolve https://github.com/Shopify/shipit/issues/743.

### DIFF
--- a/app/jobs/shipit/update_github_last_deployed_ref_job.rb
+++ b/app/jobs/shipit/update_github_last_deployed_ref_job.rb
@@ -5,7 +5,7 @@ module Shipit
     DEPLOY_PREFIX = 'shipit-deploy'.freeze
 
     def perform(stack)
-      stack_sha = stack.deploys.where(status: "success").last.until_commit.sha
+      stack_sha = stack.last_successful_deploy_commit&.sha
       return unless stack_sha
 
       environment = stack.environment
--- a/app/jobs/shipit/update_github_last_deployed_ref_job.rb
+++ b/app/jobs/shipit/update_github_last_deployed_ref_job.rb
@@ -1,0 +1,41 @@
+module Shipit
+  class UpdateGithubLastDeployedRefJob < BackgroundJob
+    queue_as :default
+
+    DEPLOY_PREFIX = 'shipit-deploy'.freeze
+
+    def perform(stack)
+      stack_sha = stack.deploys.where(status: "success").last.until_commit.sha
+      return unless stack_sha
+
+      environment = stack.environment
+      stack_ref = create_full_ref(environment)
+      client = Shipit.github.api
+
+      full_repo_name = stack.github_repo_name
+
+      update_or_create_ref(client: client, repo_name: full_repo_name, ref: stack_ref, new_sha: stack_sha)
+    end
+
+    private
+
+    def create_full_ref(stack_environment)
+      [DEPLOY_PREFIX, stack_environment].join("/")
+    end
+
+    def create_ref(client:, repo_name:, ref:, sha:)
+      client.create_ref(repo_name, ref, sha)
+    end
+
+    def update_or_create_ref(client:, repo_name:, ref:, new_sha:)
+      client.update_ref(repo_name, ref, new_sha)
+    rescue Octokit::UnprocessableEntity => e
+      error_msg = e.message
+      if error_msg.include? "Reference does not exist"
+        create_ref(client: client, repo_name: repo_name, ref: ref, sha: new_sha)
+      else
+        raise
+      end
+    end
+  end
+end
--- a/app/models/shipit/deploy.rb
+++ b/app/models/shipit/deploy.rb
@@ -8,6 +8,7 @@ module Shipit
       after_transition to: :success, do: :schedule_continuous_delivery
       after_transition to: :success, do: :schedule_merges
       after_transition to: :success, do: :update_undeployed_commits_count
+      after_transition to: :success, do: :update_latest_deployed_ref
       after_transition to: :aborted, do: :trigger_revert_if_required
       after_transition any => any, do: :update_release_status
       after_transition any => any, do: :update_commit_deployments
@@ -279,6 +280,10 @@ module Shipit
 
     def update_last_deploy_time
       stack.update(last_deployed_at: ended_at)
+    end
+
+    def update_latest_deployed_ref
+      stack.update_latest_deployed_ref
     end
   end
 end
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -458,6 +458,10 @@ module Shipit
       update(undeployed_commits_count: undeployed_commits)
     end
 
+    def update_latest_deployed_ref
+      UpdateGithubLastDeployedRefJob.perform_later(self)
+    end
+
     def broadcast_update
       Pubsubstub.publish(
         "stack.#{id}",
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -288,6 +288,10 @@ module Shipit
       deploys_and_rollbacks.last_completed
     end
 
+    def last_successful_deploy_commit
+      deploys_and_rollbacks.last_successful&.until_commit
+    end
+
     def previous_successful_deploy(deploy_id)
       deploys_and_rollbacks.success.where("id < ?", deploy_id).last
     end
--- a/app/models/shipit/task.rb
+++ b/app/models/shipit/task.rb
@@ -47,6 +47,10 @@ module Shipit
         completed.last
       end
 
+      def last_successful
+        success.last
+      end
+
       def current
         active.exclusive.last
       end
--- a/test/jobs/update_github_last_deployed_ref_job_test.rb
+++ b/test/jobs/update_github_last_deployed_ref_job_test.rb
@@ -1,0 +1,88 @@
+require 'test_helper'
+
+module Shipit
+  class UpdateGithubLastDeployedRefJobTest < ActiveSupport::TestCase
+    setup do
+      @stack = shipit_stacks(:shipit)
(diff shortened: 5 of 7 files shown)


app/jobs/shipit/update_github_last_deployed_ref_job.rb
```diff
@@ -0,0 +1,37 @@
...
+    def create_full_ref(stack_environment)
+      [DEPLOY_PREFIX, stack_environment].join("/")
+    end
+
+    def create_ref(client:, repo_name:, ref:, sha:)
+      client.create_ref(repo_name, ref, sha)
+    end
+
+    def update_or_create_ref(client:, repo_name:, ref:, new_sha:)
+      client.update_ref(repo_name, ref, new_sha)
```
ajshepley: When the request is valid (e.g. the ref exists), updating is Idempotent, to the extent that repeated Update calls will return 200 and not affect the ref (the SHA remains the same).
A 422 with a specific message is returned when the ref doesn't exist, so only in that case do we create the ref.

The most common case in Shipit will be updating a ref - creation will usually only happen for a completely new stack. Therefore, this is much more efficient overall than trying to check for existence in advance (i.e. by switching off of `GET /ref/`)

app/jobs/shipit/update_github_last_deployed_ref_job.rb
```diff
@@ -0,0 +1,37 @@
...
+    rescue Octokit::UnprocessableEntity => e
+      error_msg = e.message
+      if error_msg.include? "Reference does not exist"
+        create_ref(client: client, repo_name: repo_name, ref: ref, sha: new_sha)
+      end
```
DazWorrall: We should `raise` if we don't recognise the error message.

ajshepley:
Yeah, I was able to test with Shipit running locally using a stack based on a personal project branch. 

This led to me discovering the issue fixed with commit `c3b669c`. I may try to add a test for that case shortly, actually.

That said, even when 🎩 -ing I find there's a lot going on with Shipit and so it's easy (so far) for me to miss or not think of important cases. So if you're thinking I might have missed something, do let me know! :)

app/jobs/shipit/update_github_last_deployed_ref_job.rb
ajshepley: I'm re-raising in the other branch of the conditional now. Let me know if you meant something else.

ajshepley:
I realized there's a flaw in what I've written in [c3b669c](https://github.com/Shopify/shipit-engine/pull/917/commits/c3b669cf6e18272a9d6f7661d2aab959a74bd80c) - it could, in theory, retrieve a deploy other than the one that enqueued the job.

That alone isn't a problem and, in fact, ensures better behaviour of only updating with the latest sha. However, it's technically possible that a different, unsuccessful deploy could have occurred between the job being enqueued and the job firing. This would result in an incorrect ref being set.

I'm preparing a change that has the deploy provide the job with its own `until_commit` sha. This means that the job _could_ be stale by the time it executes, but in exchange it can never update the ref with the sha of a commit that wasn't successfully deployed.

However, something I'm now wondering about is if we want the ref to be updated for other Completed deploy statuses (such as Faulty, etc.). Those environments will technically be running new changes. But on the other hand, maybe we don't want to signal to other tools/features that they should be using the new commit in those cases? @JackTLi @DazWorrall what do you guys think?

JackTLi:
hm I'm a little lost on the change to include the `sha` as a job parameter - what's wrong with `stack_sha = stack.last_deployed_commit.sha` in the job? I would prefer this since it'll always give whatever is out right now. Ideally this value should be the one that's in sync with the ref

ajshepley:
For posterity, we discussed Jack's comment above offline:

In practice, `stack.last_deployed_commit` was somehow returning when running shipit locally / 🎩, but manually calling `stack.deploys.last` from the job correctly returned the most recent deploy.

But this ran the risk of failed deploys' SHAs being picked up, etc. Filtering seemed long-in-the-tooth compared to just passing the sha from the deploy triggering the job.

However, when discussing this it seemed that it was better to just make the filtered call (`stack_sha = stack.deploys.where(status: "success").last&.until_commit&.sha`) in the job, ensuring that it is correct even if it's a bit ugly, rather than run into potential issues due to staleness or ordering concerns. So I've now gone down that route, and added a test that checks that failed deploy SHAs are not sent to github.

As a sidenote, we also came to the conclusion that, due to rollbacks needing to update the ref, we'll let Octokit force the update_ref call via `force: true`. Github, by default, does the opposite (and so only allows fast-forwards) but Octokit true as the default.

app/jobs/shipit/update_github_last_deployed_ref_job.rb
```diff
@@ -0,0 +1,39 @@
+module Shipit
+  class UpdateGithubLastDeployedRefJob < BackgroundJob
+    queue_as :default
+
+    DEPLOY_PREFIX = 'shipit-deploy'.freeze
+
+    def perform(stack)
+      stack_sha = stack.deploys.where(status: "success").last&.until_commit&.sha
```
JackTLi: Are we expecting any `nil`s here? If so we should probably bail out of the job if `stack_sha` is nil

```ruby
return unless stack_sha
```

ajshepley: I think a nil is possible in a somewhat rare edge case: A new stack, where the first deploy was successful (and thus enqueues the job) but is immediately reverted.

In this situation the Rollback deploy could call `Rollback.update_release_status`, which would mark the only successful deploy as unsuccessful, and the Rollback deploy can still be pending/running/validating when the Job finally executes.

The stack technically doesn't need to be new, but merely needs all prior deploys to the above to not be 'success'. Not that it makes it any more likely.

This is fairly contrived, but not _too_ crazy, so I think it's OK to guard here. But if you think I'm being a bit too paranoid (or misread something) let me know. I've simply had bad luck with deploying faulty changes over the past couple of weeks 😉.

JackTLi:
```ruby
if stack_sha.can_be_nil?
  add('return unless stack_sha')
else
  remove('&').from_method_calls
end
```

Since it can be `nil`, sounds like we should do the first option 😉 
After that this should be good to 🚢

app/jobs/shipit/update_github_last_deployed_ref_job.rb
```diff
@@ -0,0 +1,41 @@
+module Shipit
+  class UpdateGithubLastDeployedRefJob < BackgroundJob
+    queue_as :default
+
+    DEPLOY_PREFIX = 'shipit-deploy'.freeze
+
+    def perform(stack)
+      stack_sha = stack.deploys.where(status: "success").last.until_commit.sha
```
casperisfine: You should put that query in a `Stack` method, so that it both have a meaningful name, and is easy to unit test in `stacks_test.rb`.

Also there's already a bunch of methods doing something very close.

There is also a scope for success status: https://github.com/Shopify/shipit-engine/blob/6422110b805c88ee5bd40936ae2b46b06a4e6b0f/app/models/shipit/task.rb#L29

ajshepley: Yeah, it's definitely worth extracting. I'll do so now.

Though I figure I should explain why I didn't re-use the existing scopes and methods:

---

I was previously trying to retrieve this information via `stack.last_deployed_commit` (which actually wasn't restrictive enough after talking to Jack, but that's a separate issue.)

That method is just calling `last_completed_deploy`, which itself is just `deploys_and_rollbacks.last_completed`. `last_completed` is just a class method returning the last record of the `completed` scope. That scope includes 'success'.

This worked fine in automated tests. However, I observed some odd behaviour when testing shipit itself locally.

When the Deploy transitioned to Success, it would trigger my method and enqueue the job. However, when the Job later called into the stack (via `last_completed_deploy`), it would return the deploy _prior_ to the one that triggered the job.

But  `stack.deploys` returns the latest deploy correctly.

---

I'm new to Rails, so I wasn't sure if this was due to how scopes work under the hood, or some other mechanism that I wasn't understanding.

I haven't gone so far as looking at the queries themselves to determine exactly what's going on here, but that was a reason I avoided using the existing scope here.

I have a bit more bandwidth now to test around and figure out this behaviour (or discover the flaw in my testing that led me to think this was happening).

ajshepley: I did some more research and talked to Jack about the stuff I described above.

It seems that state_machine's `after_transition` executes before the active record commits, and so the job is enqueued before the commit as well.

When running Shipit locally, this can result in the Job itself executing before the Deploy's change to 'success' has been committed, and so the scope filtering to the most recent deploy of 'success' will return the incorrect deploy.

This race condition explains why I was seeing the odd behaviour described above. This is actually an issue anywhere we're using after_transition to enqueue a job that accesses db state.

---

After talking to Jack, we've decided that this is unlikely to occur in production. So I'm going to defer to adding a Stack method that uses the aforementioned scope.

There are a few different overall approaches and workarounds that could be taken, but Jack and I agreed that they should be held off and done on a later change.

PR #922, opened 2019-07-19
Set email field using private emails if public ones are not found

ajshepley:
This PR looks to address https://github.com/Shopify/shipit/issues/775, wherein the EMAIL variable is not filled if the Github user has no public email addresses. We'd also like the ability to filter the emails selected by a specific org (as some users may be using a personal public email, or may have multiple emails including personal ones).

This is accomplished, in the case where they have no public email, by retrieving their emails from `/emails` and filtering it by a provided org domain, if any.

Some more specific tests are needed for this before merging, as well as (possibly) selection of the Primary email in the case where the user has multiple valid candidate emails.

I also need to :tophat: from a local running Shipit application itself, but I wanted to get some eyes on this approach first and whether it belongs here or if more things should be moved to the app integration side of things.

### DIFF
--- a/app/models/shipit/user.rb
+++ b/app/models/shipit/user.rb
@@ -131,18 +131,18 @@ module Shipit
     end
 
     def org_email?(email_address)
-      email_address.present? && email_address.split("@").last&.include?(org_domain)
+      return false if email_address.blank?
+
+      org_domains = Shipit.preferred_org_emails
+      return true if org_domains.blank?
+
+      org_domains.any? { |domain| email_address.end_with?("@#{domain}") }
     end
 
-    def filter_org_emails(emails)
-      emails
-        .select { |email| org_email?(email.email) }
-        .map(&:email)
-    end
-
-    # Domain to filter email addresses by. Override in integration application.
-    def org_domain
-      ""
+    def filter_org_emails(github_email_records)
+      org_email_records = github_email_records.select { |email_record| org_email?(email_record.email) }
+      email_record = org_email_records.find(-> { org_email_records.first }, &:primary)
+      email_record&.email
     end
   end
 end
--- a/app/models/shipit/user.rb
+++ b/app/models/shipit/user.rb
@@ -97,10 +97,16 @@ module Shipit
         github_user = github_user.rels[:self].get.data
       end
 
+      github_email = if org_email?(github_user.email)
+        github_user.email
+      else
+        filter_org_emails(github_api.emails)
+      end
+
       assign_attributes(
         github_id: github_user.id,
         name: github_user.name || github_user.login, # Name is not mandatory on GitHub
-        email: github_user.email,
+        email: github_email,
         login: github_user.login,
         avatar_url: github_user.avatar_url,
         api_url: github_user.url,
@@ -122,6 +128,21 @@ module Shipit
       update!(github_user: github_author)
     rescue Octokit::NotFound
       false
+    end
+
+    def org_email?(email_address)
+      email_address.present? && email_address.split("@").last&.include?(org_domain)
+    end
+
+    def filter_org_emails(emails)
+      emails
+        .select { |email| org_email?(email.email) }
+        .map(&:email)
+    end
+
+    # Domain to filter email addresses by. Override in integration application.
+    def org_domain
+      ""
     end
   end
 end
--- a/app/models/shipit/user.rb
+++ b/app/models/shipit/user.rb
@@ -97,16 +97,10 @@ module Shipit
         github_user = github_user.rels[:self].get.data
       end
 
-      github_email = if org_email?(github_user.email)
-        github_user.email
-      else
-        filter_org_emails(github_api.emails)
-      end
-
       assign_attributes(
         github_id: github_user.id,
         name: github_user.name || github_user.login, # Name is not mandatory on GitHub
-        email: github_email,
+        email: appropriate_email_for(github_user),
         login: github_user.login,
         avatar_url: github_user.avatar_url,
         api_url: github_user.url,
@@ -130,19 +124,26 @@ module Shipit
       false
     end
 
-    def org_email?(email_address)
-      return false if email_address.blank?
-
+    def email_valid_and_preferred?(email_address)
       org_domains = Shipit.preferred_org_emails
       return true if org_domains.blank?
+      return false if email_address.blank?
 
       org_domains.any? { |domain| email_address.end_with?("@#{domain}") }
     end
 
-    def filter_org_emails(github_email_records)
-      org_email_records = github_email_records.select { |email_record| org_email?(email_record.email) }
-      email_record = org_email_records.find(-> { org_email_records.first }, &:primary)
-      email_record&.email
+    def appropriate_email_for(github_user)
+      return github_user.email if email_valid_and_preferred?(github_user.email)
+
+      begin
+        github_api.emails
+                  .sort_by { |e| e.primary ? 0 : 1 }
+                  .map(&:email)
+                  .find { |e| email_valid_and_preferred?(e) }
+      rescue Octokit::NotFound
+        # If the user hasn't agreed to the necessary permission, we can't access their private emails.
+        nil
+      end
     end
   end
 end
--- a/test/models/commits_test.rb
+++ b/test/models/commits_test.rb
@@ -8,6 +8,8 @@ module Shipit
       @pr.message = "Merge pull request #31 from Shopify/improve-polling\n\nSeveral improvements to polling"
       @stack.reload
       @commit = shipit_commits(:first)
+
+      FakeWeb.register_uri(:get, "https://api.github.com/user/emails", status: %w(200 OK), body: {}.to_json, content_type: 'application/json')
     end
 
     test '.create_from_github handle unknown users' do
--- a/test/models/commits_test.rb
+++ b/test/models/commits_test.rb
@@ -9,7 +9,11 @@ module Shipit
       @stack.reload
       @commit = shipit_commits(:first)
 
-      FakeWeb.register_uri(:get, "https://api.github.com/user/emails", status: %w(200 OK), body: {}.to_json, content_type: 'application/json')
+      stub_request(:get, "https://api.github.com/user/emails").to_return(
+        status: %w(200 OK),
+        body: {}.to_json,
+        headers: {"Content-Type" => "application/json"},
+      )
     end
 
     test '.create_from_github handle unknown users' do
--- a/test/models/users_test.rb
+++ b/test/models/users_test.rb
@@ -3,6 +3,7 @@ require 'test_helper'
 module Shipit
   class UsersTest < ActiveSupport::TestCase
     setup do
+      @previous_preferred_org_emails = Shipit.preferred_org_emails
       @user = shipit_users(:walrus)
       @github_user = stub(
         id: 42,
@@ -12,6 +13,7 @@ module Shipit
         avatar_url: 'https://avatars.githubusercontent.com/u/42?v=3',
         url: 'https://api.github.com/user/george',
       )
+      @org_domain = "shopify.com"
       @minimal_github_user = stub(
         id: 43,
         name: nil,
@@ -21,6 +23,10 @@ module Shipit
         url: 'https://api.github.com/user/peter',
         rels: nil,
       )
+    end
+
+    teardown do
+      Shipit.preferred_org_emails = @previous_preferred_org_emails
     end
 
     test "find_or_create_from_github persist a new user if he is unknown" do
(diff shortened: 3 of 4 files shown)


app/models/shipit/user.rb
```diff
@@ -123,5 +129,20 @@ def identify_renamed_user!
     rescue Octokit::NotFound
       false
     end
+
+    def org_email?(email_address)
+      email_address.present? && email_address.split("@").last&.include?(org_domain)
```
casperisfine: I don't think `include?` is a good idea. It means if you set org domain to `example.com`, then `example.com.rogue-haxor.net` will match.

`email_address.end_with?("@#{org_domain}")` would be much safer.

app/models/shipit/user.rb
```diff
@@ -123,5 +129,20 @@ def identify_renamed_user!
...
+
+    def filter_org_emails(emails)
+      emails
+        .select { |email| org_email?(email.email) }
+        .map(&:email)
+    end
+
+    # Domain to filter email addresses by. Override in integration application.
+    def org_domain
+      ""
```
casperisfine: I don't like this way of configuring it. I'd rather see a property on the `Shipit` namespace. There is plenty of those already.

test/models/commits_test.rb
```diff
@@ -11,6 +11,8 @@ class CommitsTest < ActiveSupport::TestCase
     end
 
     test '.create_from_github handle unknown users' do
+      FakeWeb.register_uri(:get, "https://api.github.com/user/emails", status: %w(200 OK), body: {}.to_json)
```
casperisfine: That request shouldn't happen unless you've enabled the `org_email` property.

app/models/shipit/user.rb
```diff
@@ -123,5 +129,20 @@ def identify_renamed_user!
...
       false
     end
+
+    def org_email?(email_address)
+      email_address.present? && email_address.split("@").last&.include?(org_domain)
+    end
+
+    def filter_org_emails(emails)
+      emails
+        .select { |email| org_email?(email.email) }
```
casperisfine: `email.email` is a smell.

app/models/shipit/user.rb
ajshepley: Oh, that's much better.

app/models/shipit/user.rb
ajshepley: Yeah, this was on the list of things I need to revisit. Even ignoring that, 'emails' and 'email' are both kinda wrong terms for describing the records (and their collection) returned from the Github API. I'll adjust them.

test/models/commits_test.rb
ajshepley: We want to pull private emails even if they don't provide an org email; that is, the request would occur in any case where we the initial `github_user` does not have an email.

That said, a blank email _is_ different from a `null` email. I was assuming github would never return a blank email (as I don't know what that actually means) so I was treating blank email fields in these tests as equivalent to null, but maybe that's not correct.

app/models/shipit/user.rb
ajshepley: Would that best be done as a class variable on User (or some other class?), a module variable in a new file (itself just `def`ing Shipit), or as an overridden constant in one of the above?

I'd be interested in an example if you have one, as I'm not sure if the things I'm looking at are what you're talking about. I'm still a bit green on the way we do things with Ruby and Rails.

casperisfine: No, `Shipit` is better, as it's the app namespace, to it make sense to centralize this kind of configurations there.

As for the example, there is some here: https://github.com/Shopify/shipit-engine/blob/6bd83a31ad88ed81bc920d9d57c83055110bb815/lib/shipit.rb#L62

test/models/commits_test.rb
casperisfine: If github gives you an empty string as email, better coerce it to `nil`.

app/models/shipit/user.rb
ajshepley: Ah, thank you. This has led to some more questions though:

I noticed in testing that timeout_exit_codes as set below is actually nil at runtime.
```ruby
 attr_accessor :disable_api_authentication, :timeout_exit_codes, :preferred_org_emails
 self.timeout_exit_codes = [].freeze
// calling Shipit.timeout_exit_codes anywhere in the code returns nil, rather than an empty array.
```
Do you know if that's intended? `@timeout_exit_codes = [].freeze` also behaves the same, but `@@timeout_exit_codes = [].freeze` results in the var defaulting to an empty array at runtime.

I'm not sure if I'm just misunderstanding Rubys' runtime class structure or something like that.
It's overridden in the Shipit project and relevant unit test, so it doesn't appear to be an issue, but I think the intent here was to have a non-nil default value?

app/models/shipit/user.rb
ajshepley: I've adjusted it now.

test/models/users_test.rb
```diff
@@ -58,10 +59,58 @@ class UsersTest < ActiveSupport::TestCase
...
+      github_org_user = stub(
+        id: 42,
+        name: 'Jim Jones',
+        login: 'jim',
+        email: "jim@#{@org_domain}",
+        avatar_url: 'https://avatars.githubusercontent.com/u/42?v=3',
+        url: 'https://api.github.com/user/jim',
+      )
+
+      Shipit.preferred_org_emails = []
```
DazWorrall: I don't think it matters right now but to save us in future I think this should be set back to the previous value in an ensure block.

app/models/shipit/user.rb
```diff
@@ -123,5 +129,20 @@ def identify_renamed_user!
...
+
+      org_domains = Shipit.preferred_org_emails
+      return true if org_domains.empty? || org_domains.all?(&:blank?)
+
+      org_domains.any? { |domain| email_address.end_with?("@#{domain}") }
+    end
+
+    def filter_org_emails(github_email_records)
+      org_email_records = github_email_records.select { |email_record| org_email?(email_record.email) }
+      email_record = org_email_records.find(-> { github_email_records.first }, &:primary)
```
DazWorrall: 🤔I can't tell what this does, you're going to have to humour me - why do we go back to `github_email_records` here?

app/models/shipit/user.rb
casperisfine: Are you sure? I just checked and it works fine:

```
>> Shipit.timeout_exit_codes
=> []
>> Shipit.timeout_exit_codes.frozen?
=> true
```

If you get it `nil` it might be that something overwrites it.

app/models/shipit/user.rb
```diff
@@ -97,10 +97,16 @@ def github_user=(github_user)
         github_user = github_user.rels[:self].get.data
       end
 
+      github_email = if org_email?(github_user.email)
+        github_user.email
+      else
+        filter_org_emails(github_api.emails)
```
casperisfine: I'm not super kin on that email selection code, can't really put my finger on it but something rubs me the wrong way.

I think casting the Octokit response to a simple array of strings would make for s much simpler code:

```ruby
github_email = github_user.email if org_email?(github_user.email)
github_email ||= github_api.emails.map(&:email).find { |e| org_email?(e) }
github_email ||= github_user.email
```

app/models/shipit/user.rb
ajshepley: Whoops, that was a typo from when I was renaming these fields. I'll fix and add a test to cover the case (which is point 6 on the https://github.com/Shopify/shipit/issues/775).

test/models/commits_test.rb
```diff
@@ -11,6 +11,8 @@ class CommitsTest < ActiveSupport::TestCase
     end
 
     test '.create_from_github handle unknown users' do
+      FakeWeb.register_uri(:get, "https://api.github.com/user/emails", status: %w(200 OK), body: {}.to_json, content_type: 'application/json')
```
casperisfine: Again, I think the default config in the test suite should make it so that email filtering isn't enabled, hence that we don't need to fake these calls.

Either that or the calls should be stubbed globally so that we don't have to copy that over in dozens of tests.

app/models/shipit/user.rb
```diff
@@ -123,5 +129,20 @@ def identify_renamed_user!
     rescue Octokit::NotFound
       false
     end
+
+    def org_email?(email_address)
+      return false if email_address.blank?
+
+      org_domains = Shipit.preferred_org_emails
+      return true if org_domains.empty? || org_domains.all?(&:blank?)
```
casperisfine: Just `Shipit.preferred_org_emails.blank?` is enough. No need to be defensive about people putting blank strings in that array.

app/models/shipit/user.rb
ajshepley: Part of the complexity is from selecting the email marked as Primary if multiple valid org emails are found, and otherwise picking any of them if none are marked primary.
I could not think of a cleaner way of doing so.

casperisfine: > Part of the complexity is from selecting the email marked as Primary if multiple valid org emails are found

But the primary is accessible as `user.email`, so if it's a valid org email, it's already set prior to entering that sub routine.

app/models/shipit/user.rb
ajshepley: Ah, this may be some test specific thing.
Rails console returns an empty and frozen array, but calling `Shipit.timeout_exit_codes` and similar from Test entry points (either directly, or via a class like User making the call) results in nil.

I thought I had observed it while :tophat: testing the email logic, but perhaps I was mistaken.

app/models/shipit/user.rb
ajshepley: If they do have a public email, then it should be guaranteed to be their primary one, so if we've specified a disqualifying org email, then it's as simple as selecting the first/any result from github_api.emails that fits the org_email.

The case I'm concerned about is if the user has no public email (e.g. github_user.email will be blank), and multiple private org emails, with one of those private emails marked as primary, and no guarantee on the primary email's position in the set returned by `github_api.emails`.

Perhaps I'm misunderstanding what you're saying, though.

casperisfine: Ok, I didn't think of private emails. In this case:

```ruby
github_api.emails.sort_by { |e| e.primary ? 0 : 1 }.map(&:email).find { |e| org_email?(e) }
```

Also BTW, accessing private emails requires a specific permission: https://developer.github.com/v3/users/emails/#list-email-addresses-for-a-user `user:email`, does Shipit already ask for it, or do we have to add it?

ajshepley: Ah! Good catch.

I was assuming our apps already had User permissions as they were already pulling User information in the first place, but it turns out we're only accessing the public user information through this API and currently have no user permissions enabled for our apps (Shopify Shipit, Shipit Mobile Staging and Shipit Mobile I think - let me know if I've missed any.)

I believe we'd just need to enable the permission (`User Permissions -> Email Addresses -> Read Only`) on the apps.

Based on the [app user permissions documentation](https://developer.github.com/apps/building-github-apps/identifying-and-authorizing-users-for-github-apps/#user-level-permissions), it's not clear to me whether adding this permission would be graceful, or if we'd suddenly see a bunch of permission access prompts for our org users.

I don't think we have user-to-server tokens to worry about, but I could be mistaken.

@DazWorrall any thoughts?

test/models/commits_test.rb
ajshepley: Right, my apologies for not addressing this correctly earlier.

I wanted to avoid adding more branching logic on whether or not to make the `github_api.emails` request in User, as that begets further complexities and discussions about exposing control of that logic to integrations, so I'll go with a singular register call in the setup for the test here if that's OK.
A more global stub could be done in `test_helper.rb` but I don't think it's necessary / I think it's better for tests to catch any new unaccounted-for interaction with the API in other tests.

In the case of users_test, each test that interacts with the API is setting a specific return value germane to the test logic, so I'm going to leave that as is.

casperisfine: > I wanted to avoid adding more branching logic on whether or not to make the `github_api.emails` request in User

It really isn't that much more complexity, and would save useless requests if no org domain is set.

ajshepley: (sorry for the bricks of text, I just want to make sure I'm getting things right)

Just to make sure I'm reading things right, we want to make the request behaviour configurable/disable-able because the requests would be useless in the situation where:
1) No org domain is set for the integration.
2) The user(s) have no public email.
3) We specifically don't care about private emails for that project.

That is, in these tests, the (fake) request is made not because of filtering being enabled, but because the fallback logic is to check /emails for a valid email if the public /user did not provide one. Org domain filtering just makes it more likely that the public email is seen as invalid, but in the case of these tests, the email is invalid because it's blank as part of the test.

I can add a flag for projects to configure whether that fallback logic should be happen, but I'll need to figure out where and how far that flag would be exposed (e.g. in shipit.yml, env vars, etc).

I assumed that was too much extra complexity for this change, but maybe that's just because I haven't done that before.

casperisfine: Well, I had something much simpler in mind. Basically if you don't restrict emails to a specific domain, it mean you don't really care about private/public emails, hence you can simply take the primary public one directly accessible on the user record.

ajshepley: I see, I was unclear before and in the PR description.

The way I interpreted the source issue is that org domain filtering is an additional, optional constraint - and when we don't do that filtering, we still care about trying our best to get _some_ email, whether it's public or private.

Then in the case where we do want a specific org email, we'd rather have no email than a non-org domain email.

casperisfine:
Another thing to check: I'm not sure you can access private emails of an user that didn't grant the permission. It's fairly common to have commit authors which aren't users of the shipit instance, so we should double check that possible error scenarios and handle it properly.

ajshepley:
Without the permission, Github returns a 404 for /emails from what I've been able to test. I've handled that by logging a warning and returning nil (we should only be in that block once we've determined that we can't use their public email to begin with). I added a test for this case as well.

Beyond that and enabling the app permissions as described in https://github.com/Shopify/shipit-engine/pull/922#discussion_r306879403, do you see anything else needing to be done in regards to permissions?

casperisfine:
> by logging a warning

I wouldn't bother with the warning since it's expected that we'll frequently hit such 404.

> do you see anything else needing to be done in regards to permissions?

Not that I can think of.

ajshepley:
For the `preferred_org_emails` field, I've deferred to how we've done it with `internal_hook_receivers` and `task_logger` as it works correctly that way both in tests and on the gem integration/Shipit side.

In regards to permissions, I plan on changing the permission in Shopify Shipit (`User Permissions -> Email Addresses -> Read Only`) after merging this PR, but before updating the gem + deploying in Shipit.

Is there anything else we want to address here at the moment?

app/models/shipit/user.rb
```diff
@@ -123,5 +123,28 @@ def identify_renamed_user!
...
+      return true if org_domains.blank?
+
+      org_domains.any? { |domain| email_address.end_with?("@#{domain}") }
+    end
+
+    def appropriate_email_for(github_user)
+      return github_user.email if org_email?(github_user.email)
+
+      begin
+        github_api.emails
```
ajshepley: I figure I should mention,
I am more of a fan of this approach:
```ruby
github_api
  .emails
  .sort_by { |e| e.primary ? 0 : 1 }
  .map(&:email)
  .find { |e| org_email?(e) }
```
but it violates our rubocop settings.

I looked through various projects using a regex to find method chaining styles, and this ones was not uncommon.

app/models/shipit/user.rb
```diff
@@ -123,5 +123,28 @@ def identify_renamed_user!
     rescue Octokit::NotFound
       false
     end
+
+    def org_email?(email_address)
+      return false if email_address.blank?
+
+      org_domains = Shipit.preferred_org_emails
+      return true if org_domains.blank?
```
casperisfine: This check should be first, this way if the option is disabled, we won't fetch private emails even if the public one is blank.

casperisfine: Also since it return true when the option is disabled, the name is kinda weird, but no huge deal.

test/models/users_test.rb
```diff
@@ -58,10 +59,88 @@ class UsersTest < ActiveSupport::TestCase
...
+      Shipit.preferred_org_emails = [@org_domain]
+      expected_email = "myuser@#{@org_domain}"
+      FakeWeb.register_uri(:get, "https://api.github.com/user/emails", status: %w(200 OK), body: [{email: expected_email}].to_json, content_type: "application/json")
+
+      user = User.find_or_create_from_github(@github_user)
+      assert_equal expected_email, user.email
+    end
+
+    test "find_or_create_from_github selects private and primary org email for user when necessary" do
+      Shipit.preferred_org_emails = [@org_domain]
```
casperisfine: That's global state, it need to be properly reset after you test. Best way to do it is:

```ruby
teardown do
  Shipit.preferred_org_emails = nil
end

app/models/shipit/user.rb
ajshepley: We want to fetch private emails when the public one is blank, regardless of whether the org_emails filter is set or not.

test/models/users_test.rb
ajshepley: Right, I don't want to affect other tests. I used a `begin...ensure` block in another test to do this, but I'll strip it out in favor of setup/teardown to cover the whole test class.

app/models/shipit/user.rb
DazWorrall: I think if the `org_emails` filter is missing we can assume the operator doesn't care about this, and we can leave it blank.

ajshepley: Alrighty. I'll adjust (and update the source issue).

PR #925, opened 2019-07-25
Change the latest deployed ref job to create a proper refs/heads/ branch.

ajshepley:
We were creating the ref for the latest-deployed-commit under `refs/shipit-deploy/stack_env_name`.
However, this turned out to be unusable for Github's branch-based APIs, as Github only treats refs under `refs/heads/` as branches for those purposes.

As such, this PR changes the created refs to be under `refs/heads/shipit-deploy/stack_env_name`, and thus they show up as branches (`git fetch ; git branch -a`, the Branches view in Github, etc).

I 🎩locally with a test repo of mine:

The created ref `refs/heads/shipit-deploy/myenvironment` can be fetched, and will create the mapping `shipit-deploy/myenvironment -> origin/shipit-deploy/myenvironment` representing the remote `remotes/origin/shipit-deploy/devvy-myenvironment`.

Not sure how we want to handle cleanup of the existing refs.

### DIFF
--- a/app/jobs/shipit/update_github_last_deployed_ref_job.rb
+++ b/app/jobs/shipit/update_github_last_deployed_ref_job.rb
@@ -2,6 +2,7 @@ module Shipit
   class UpdateGithubLastDeployedRefJob < BackgroundJob
     queue_as :default
 
+    BRANCH_REF_PREFIX = 'refs/heads'.freeze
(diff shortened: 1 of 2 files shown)


DazWorrall:
> Not sure how we want to handle cleanup of the existing refs.

refs are cheap, I'm not worried about it personally. This never made it into a release either so we can assume the impact is limited to Shopify.

PR #926, opened 2019-07-25
fix CVE-2015-9284 per omniauth/omniauth#809

wenzowski:
The request phase of the OmniAuth Ruby gem is vulnerable to Cross-Site Request Forgery when used as part of the Ruby on Rails framework, allowing accounts to be connected without user intent, user interaction, or feedback to the user. This permits a secondary account to be able to sign into the web application as the primary account.

same fix as merged in Shopify/omniauth-identity#8
@see https://nvd.nist.gov/vuln/detail/CVE-2015-9284


wenzowski:
Corporate CLA signed on behalf of [Button Inc.](https://www.ic.gc.ca/app/scr/cc/CorporationsCanada/fdrlCrpDtls.html?corpId=10983128).

wenzowski:
@ajshepley any idea who to ping about the CLA?

wenzowski:
@casperisfine I don't know how to trigger CLA status check. It's been signed for both myself and my org. When I visit https://cla.shopify.com/ I see:
> GitHub username has already signed a CLA

casperisfine:
For the CLA: Must have been a dropped hook or something, triggering a re-run fixed it.

As for the patch I really doubt it works. You can't just add the gem, you also have to make sure the request is initiated with a POST, which I'm pretty sure it isn't right now, and making it so is actually quite hard.

wenzowski:
That's too bad. 

You're saying that since the `redirect_to` [when the user is not authorized](https://github.com/Shopify/shipit-engine/blob/master/app/controllers/concerns/shipit/authentication.rb#L26) is still an http `get` is the reason this is hard?

Presumably redirecting to a splash page with a "login with github" button that sends a `post` would do the trick? I don't mind taking a stab at that if that's all it would be.

casperisfine:
> You're saying that since the `redirect_to` [when the user is not authorized](https://github.com/Shopify/shipit-engine/blob/master/app/controllers/concerns/shipit/authentication.rb#L26) is still an http `get` is the reason this is hard?

Yes, we'd need to show a page with a button to trigger the auth with a POST.

>  I don't mind taking a stab at that if that's all it would be.

I'd like to make sure it's actually needed first. I don't fully understand the implications of the CVE, but it seems very mild to me.

> allowing accounts to be connected without user intent, user interaction, or feedback to the user. This permits a secondary account to be able to sign into the web application as the primary account.

That part I don't quite get, Shipit doesn't have roles or anything like that, and I can't think of any negative consequence about being authenticated "against your will".

@EiNSTeiN- I could use your help (or someone else from your team if you don't have time) to assert the criticity of this thing, and wether or not we should redirect to a form or not.

EiNSTeiN-:
@clayton-shopify would someone from your team be able to handle this?

wenzowski:
@casperisfine @EiNSTeiN- @clayton-shopify Would a splash page with a "login with github" button that sends a `post` be welcome here? Where did we land on figuring out if it's actually needed?

Being authenticated against your will would presumably allow an attacker to deploy commits that should otherwise be held back, no?

EiNSTeiN-:
> Where did we land on figuring out if it's actually needed?

that will indeed work to fix this issue

casperisfine:
> Being authenticated against your will would presumably allow an attacker to deploy commits that should otherwise be held back, no?

I don't think so no, the CSRF protection will prevent that. You'll only be able to generate GET requests.

PR #927, opened 2019-07-26
Remove 'refs/' prefix from latest deployed commit refs.

ajshepley:
This PR removes the 'refs/' prefix from the refs we are creating for the last-deployed-commit.

It turns out that Octokit's `[update_ref](https://github.com/octokit/octokit.rb/blob/master/lib/octokit/client/refs.rb#L68)` method will always add a `refs/` prefix, whereas `[create_ref](https://github.com/octokit/octokit.rb/blob/master/lib/octokit/client/refs.rb#L48)` does not.

So create_ref would pass, but then the update_ref on the next deploys would always fail.


ajshepley:
I'll try to put one up today. May come after the town hall though.

Issue #881, opened 2019-03-21
GitHub status page API is broken


casperisfine:
We record the GitHub status in https://github.com/Shopify/shipit-engine/blob/f2f6dbb672f3a5073d46e5353509705b25e55ec5/app/models/shipit/github_status.rb and use it to display a warning when there is an incident on GitHub.

It seems that they recently migrated from their in house solution to statuspage.io, so the API no longer exist.

We can either refactor the code to use statuspage.io's API at https://www.githubstatus.com/api/v2/summary.json

Or simply strip all that code.

cc @Shopify/pipeline

sciyoshi:
See also https://github.com/octokit/octokit.rb/issues/1107

ohsabry:
This was a big gotcha for us, and we had to make a rakefile with a commented out `Shipit::GithubStatus.refresh_status` to run shipit: 
https://github.com/TouchBistro/docker-shipit-engine/blob/master/lib/tasks/cronfix.rake#L6


Have you guys decided whether you want to switch to statuspage or to remove this code? we are running shipit here at our shop and would be happy to contribute with a PR

cc @itsWill sup

solojavier:
@peerkleio @casperisfine Since #928 is merged, I believe this issues can be marked as closed.

PR #928, opened 2019-07-31
Update Service Status check to use statuspage.io

peerkleio:
This is meant to address https://github.com/Shopify/shipit-engine/issues/881.  As Octokit does not presently support updates Statuspage.io this adds an api call to it directly, parses out the response for the API component and updates the UI call for the new key.

I had some trouble with Fakeweb and newer Ruby so i replace it with Webmock

cc/ @stantona @charles-ferguson @rbroemeling

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,7 @@ end
 
 group :development, :test do
   gem 'faker'
-  gem 'fakeweb'
+  gem 'webmock'
   gem 'rubocop', '~> 0.52.0'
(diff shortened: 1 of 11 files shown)


casperisfine:
The CI failure is legit, but other than that codes LGTM.

Just get the CI green as well as the CLI thing, and I'll happily merge this.

peerkleio:
@casperisfine done!

PR #931, opened 2019-08-12
Stack Statistics Page

peerkleio:
I created this page under the assumption that not everyone will have or want to use a metrics platform to send their deploy data too. In that case they can use this basic statistic page that shows some stats over rolling 7 day period. This is hardcoded for now for simplicity. Ideally it could be expanded to a selector with "Last 24h", "Last Week", "Last Month" or (if you are really ambitious) a custom date range.

This is how the page looks with some dummy data after manipulating the seed data.

Screen Shot 2019-08-12 at 2 14 24 PM

This PR is most certainly WIP. However, i wanted to get some upstream eyes on it before putting a bunch more work into it.  

TODO:
- [ ] tests
- [ ] CSS cleanup, not all classes are being used right now
- [ ] DRY, lots of duplication in the controller right now

### DIFF
--- a/app/models/shipit/deploy_stats.rb
+++ b/app/models/shipit/deploy_stats.rb
@@ -1,0 +1,55 @@
+module Shipit
+  class DeployStats
+    def initialize(deploys)
+      @deploys = deploys
+      @durations = @deploys.map { |d| d.duration.value }.compact
+    end
+
+    def count
+      @deploys.length
+    end
+
+    def average_duration
+      return if @durations.empty?
+      @durations.sum / @durations.length.to_f
+    end
+
+    def max_duration
+      @durations.max
+    end
+
+    def min_duration
+      @durations.min
+    end
+
+    def median_duration
+      return if @durations.empty?
+      (sorted_durations[(@durations.length - 1) / 2] + sorted_durations[@durations.length / 2]) / 2.0
+    end
+
+    def success_rate
+      return if @deploys.empty?
+      (@deploys.count(&:success?) / @deploys.length.to_f) * 100
+    end
+
+    def compare(compare_stats)
+      {
+        count: percent_change(compare_stats.count, count),
+        average_duration: percent_change(compare_stats.average_duration, average_duration),
+        median_duration: percent_change(compare_stats.median_duration, median_duration),
+      }
+    end
+
+    protected
+
+    def sorted_durations
+      @sorted ||= @durations.sort
+    end
+
+    def percent_change(from, to)
+      return if to.nil? || from.nil?
+      return to * 100 if from.zero?
+      ((to - from) / from.to_f) * 100
+    end
+  end
+end
--- a/app/models/shipit/task.rb
+++ b/app/models/shipit/task.rb
@@ -29,8 +29,11 @@ module Shipit
     scope :success, -> { where(status: 'success') }
     scope :completed, -> { where(status: COMPLETED_STATUSES) }
     scope :active, -> { where(status: ACTIVE_STATUSES) }
+    scope :not_active, -> { where(status: COMPLETED_STATUSES + UNSUCCESSFUL_STATUSES) }
     scope :exclusive, -> { where(allow_concurrency: false) }
     scope :unsuccessful, -> { where(status: UNSUCCESSFUL_STATUSES) }
+    scope :last_seven_days, -> { where("created_at > ?", 7.days.ago) }
+    scope :previous_seven_days, -> { where(created_at: 14.days.ago..7.days.ago) }
 
     scope :due_for_rollup, -> { completed.where(rolled_up: false).where('created_at <= ?', 1.hour.ago) }
 
--- a/app/assets/stylesheets/_pages/_stacks.scss
+++ b/app/assets/stylesheets/_pages/_stacks.scss
@@ -161,3 +161,71 @@
 .pagination {
   text-align: center;
 }
+
+.box {
+  border-radius: 5px;
+  width: 32%;
+  float: left;
+  padding: 5px;
+}
+
+.box__header {
+  padding: 15px 25px;
+  position: relative;
+}
+
+.box__header-title {
+  color: #333;
+  font-size: 18px;
+}
+
+.box__body {
+  padding: 0 25px;
+}
+
+.row {
+  @include clearfix;
+}
+
+/* STATS */
+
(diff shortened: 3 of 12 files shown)


app/controllers/shipit/stats_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+      @stats = {
+        average: 0,
+        min: 0,
+        max: 0,
+        median: 0,
+        success: 0
+      }
+      @diffs = {}
+
+      @previous_deploys = stack.deploys.where("created_at BETWEEN ? AND ?", 14.days.ago, 7.days.ago)
```
casperisfine: You can use `where(created_at: 14.days.ago..7.days.ago)` for better cross DB compatibility.

Also naming that scope something like `previous_week` would make the code a bit nicer.

app/controllers/shipit/stats_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+      prev_deploy_durations = @previous_deploys.map{|d| d.duration.value }
+      if prev_deploy_durations.size > 0
+        @prev_stats = {
+          count: @previous_deploys.size,
+          average: prev_deploy_durations.inject{ |sum, el| sum + el }.to_f / prev_deploy_durations.size,
```
casperisfine: `.inject{ |sum, el| sum + el }` can be replaced by `.sum`

app/controllers/shipit/stats_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+          min: prev_deploy_durations.min,
+          max: prev_deploy_durations.max,
+          median: (prev_deploy_durations.sort[(prev_deploy_durations.length - 1) / 2] + prev_deploy_durations.sort[prev_deploy_durations.length/2]) / 2.0,
+          success: (@previous_deploys.select{|d| d.success? }.size / @previous_deploys.size.to_f)*100
```
casperisfine: `.select{|d| d.success? }.size` can be replaced with `.count(&:success?)`

app/views/shipit/stats/show.html.erb
```diff
@@ -0,0 +1,85 @@
+<% subscribe events_path(channels: ["stack.#{@stack.id}"]), '#layout-content' %>
```
casperisfine: No need to subscribe to updates on this page.

app/controllers/shipit/stats_controller.rb
```diff
@@ -0,0 +1,54 @@
+module Shipit
+  class StatsController < ShipitController
+    def show
```
casperisfine: This could be `StacksController#statistics`.

app/controllers/shipit/stats_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+
+      @deploys = stack.deploys.where("created_at > ?", 7.days.ago)
+      @deploy_durations = @deploys.map{|d| d.duration.value }
+      if @deploy_durations.size > 0
+        @stats = {
+          count: @deploys.size,
+          average: @deploy_durations.inject{ |sum, el| sum + el }.to_f / @deploy_durations.size,
+          min: @deploy_durations.min,
+          max: @deploy_durations.max,
+          median: (@deploy_durations.sort[(@deploy_durations.length - 1) / 2] + @deploy_durations.sort[@deploy_durations.length/2]) / 2.0,
```
casperisfine: There seem to be some code duplication here, and the controller action is quite more complex than I'd like, as well as very procedural.

IMO it could benefit from being extracted as a model (not a db backed one, just some plain old ruby class).

Something like `DeployStats.new(deploys).count/average/etc`. It will also be much much easier to unit test.

peerkleio:
Thanks @casperisfine. I will patch this up and let you know when its ready for some more 👀

app/controllers/shipit/stats_controller.rb
peerkleio: Agreed this is very ugly.  Lots of duplication and unnecessary complexity for a controller.

peerkleio:
Updated based on feedback and added tests for the new model and controller action. Feel free to take another look at it when you are have time.

peerkleio:
Test have been fixed for all databases and rubocop errors too.

app/models/shipit/deploy_stats.rb
```diff
@@ -0,0 +1,55 @@
...
+      }
+    end
+
+    protected
+
+    def sorted_durations
+      @sorted ||= @durations.sort
+    end
+
+    def percent_change(new_value, old_value)
```
casperisfine: Nitpick: not fan of `percent_change(new_value, old_value)` as a method signature.

I think `percent_change(from, to)` would be better.

app/models/shipit/task.rb
```diff
@@ -29,8 +29,11 @@ class Task < ActiveRecord::Base
     scope :success, -> { where(status: 'success') }
     scope :completed, -> { where(status: COMPLETED_STATUSES) }
     scope :active, -> { where(status: ACTIVE_STATUSES) }
+    scope :not_active, -> { where(status: COMPLETED_STATUSES + UNSUCCESSFUL_STATUSES) }
     scope :exclusive, -> { where(allow_concurrency: false) }
     scope :unsuccessful, -> { where(status: UNSUCCESSFUL_STATUSES) }
+    scope :current_week, -> { where("created_at > ?", 7.days.ago) }
+    scope :previous_week, -> { where(created_at: 14.days.ago..7.days.ago) }
```
casperisfine: These `ago` means "last week" is a sliding window. I guess it's fine, but as always with stats, it might throw people off.

app/models/shipit/deploy_stats.rb
```diff
@@ -0,0 +1,55 @@
...
+      @durations = @deploys.map { |d| d.duration.value }.compact
+    end
+
+    def count
+      @deploys.length
+    end
+
+    def average_duration
+      return if @durations.empty?
+      (@durations.sum / @durations.length.to_f)
```
casperisfine: the parentheses are not useful here.

app/models/shipit/deploy_stats.rb
peerkleio: Updated

casperisfine:
@DazWorrall do you want to take a look as well?

PR #932, opened 2019-08-13
Bumps nokogiri for CVE-2019-5477

peerkleio:
As per https://github.com/sparklemotion/nokogiri/issues/1915, bumps nokogiri to protect against said vulnerability

@casperisfine @EiNSTeiN-


casperisfine:
I'll merge but there is little benefits to this kind of PRs. That Gemfile is only used for CI & dev basically, it's up to the host apps to update this kind of dependencies.

PR #934, opened 2019-08-16
Improve 'Commits not include in this deploy' presentation

paracycle:
## Problem

In the current state, commits not included in the deploy are presented inside a section with the header that says `Commits included in this deploy`, which creates the wrong impression on the user.

Moreover, these not-included commits are listed above the `Create Deploy` button, giving the impression that they are a part of the commits that will be included in the action.

## Solution

This PR moves the `Commits not included in this deploy` section outside and below the `Create Deploy` button so that it is visually separate from where the action is expected to happen. The information is still being displayed in a de-emphasized manner, and the title of the new section is very clear, so it should not create extra user confusion.

### Before

Screenshot 2019-08-16 at 18 51 06

### After

Screenshot 2019-08-16 at 18 49 08

### DIFF
--- a/app/assets/stylesheets/_structure/_main.scss
+++ b/app/assets/stylesheets/_structure/_main.scss
@@ -142,8 +142,9 @@ pre {
 .less-important {
     color: $grey;
     margin-bottom: 0em;
-    margin-top: 1em;
+    margin-top: 5em;
     p {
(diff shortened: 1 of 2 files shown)


ajshepley:
Looks good. Adding a not-included header is a big readability improvement alone.

PR #936, opened 2019-08-21
Use --recursive option when cloning repo from cache

wenzowski:
Addition of the `--recursive` flag enables deployment of repositories with git submodules.

Further discussion welcome.


wenzowski:
This brings the clone from cache in line with the `git_clone` method first used (and already implemented).

PR #937, opened 2019-08-28
Increase max fetch to 25

JackTLi:
Our new merge-queue can fast-forward up to 24 commits at a time right now, so the github sync here should be increased to be able to process and sync more commits to ensure things are not dropped.


DazWorrall:
Is this definitely necessary? I presume any commits > the limit will be fetched in subsequent runs, are there any downsides to this?

casperisfine:
That limit was there mostly for newly created stack. We don't want to fetch the entire git history each time you create a new stack, or if you force push the tracked branch etc, so there must be a limit. But granted that 10 was quite conservative.

Issue #935, opened 2019-08-21
Octokit cache gets huge


josacar:
Hi,

`shipit-engine` uses `Octokit` with its [cache](https://github.com/Shopify/shipit-engine/blob/master/lib/shipit.rb#L111-L117) but we are facing these cache gets bigger and bigger, in 1 week time we get 1MB objects in memcached, which later takes a lot of CPU time deserializing those megabytes every Octokit request.

This is the output log:

```
E, [2019-08-21T11:42:26.269820 #26691] ERROR -- : [ActiveJob] [Shipit::GithubSyncJob] [ae2bc425-5a2f-47e6-8685-da12cd931af7] Value for 4e73b0f9615276ddb2ab1769d50c6a00df581a5e over max size: 5242880 <= 5260143 - this value may be truncated by memcached
```

I don't know if this could be caused by the `StripCacheControl` [middleware](https://github.com/Shopify/shipit-engine/blob/master/lib/shipit/strip_cache_control.rb)

casperisfine:
Did you look at what these objects look like? Without an indication of where the growth is coming from it will be hard to figure it out.

josacar:
I'm getting the two biggest objects right now"

```
Bytes                           Cache Key
--------------------------------------------------------------------------------
1411767                 2b852c01abf5104ff9b24cc662e67b928765db20
1916999                 4e73b0f9615276ddb2ab1769d50c6a00df581a5e
```

It's composed by array of Octokit requests:

```
irb(main):002:0> Rails.cache.fetch('2b852c01abf5104ff9b24cc662e67b928765db20').class
=> Array
```

And then each request has the request and the response:

```
irb(main):004:0> Rails.cache.fetch('2b852c01abf5104ff9b24cc662e67b928765db20').first.class
=> Array
```

Checking which URLs are being cached I see this in one object:

```
irb(main):026:0> Rails.cache.fetch('2b852c01abf5104ff9b24cc662e67b928765db20').map { |x| x.first[:url] }.uniq
[...]

PR #938, opened 2019-08-29
Improve the faraday cache middleware

casperisfine:
Fix: https://github.com/Shopify/shipit-engine/issues/935

Since it's quite brittle, I explained the decision in the code comments.

Ultimately I think GitHub is right to add `Vary: Authorization` since different tokens might mean different users with potentially different accesses. However in our case we know that we're always using the same user (actually app) so we can safely disregard that directive.

The only place where we switch users is for [commit deployments](https://github.com/Shopify/shipit-engine/blob/7bcaceaefdd83c4245db02109654cdc81c10fa84/app/models/shipit/commit_deployment.rb#L24), but first it's a `POST` request to it shouldn't be cached, and second if it fails it will be with a permission error which wouldn't be cached either. So really we're fine, the only worry I have is for a potential future feature, and we might be caught off guard by this, but Shipit is relatively stable this days so the risk is low enough IMO.

On another note, it could be interesting to report this to `faraday-http-cache`, it's not exactly a bug, but I expect that other HTTP cache implementations have a way to deal with such cases, and there's no reason faraday-http-cache couldn't do it.

@josacar would you mind deploying this branch on your installation just to make sure I haven't overlooked anything? If it works for you I'll merge and cut a new release ASAP.


josacar:
@byroot I'll try this now, thanks a lot!

josacar:
No duplicated cache entries so far.

josacar:
Still working great, thanks for your help!

byroot:
👌 I'll cut a release in a bit.

casperisfine:
@josacar I released it as part of `0.28.1`.

josacar:
thanks for your help. You are such a great maintainer!

PR #948, opened 2019-10-10
Fix GitHubApp for Enterprise installations

MXfive:
Fix GitHubApp for Enterprise installations.  Options hash was not actually being modified to add in the `api_endpoint`.

### DIFF
--- a/lib/shipit/github_app.rb
+++ b/lib/shipit/github_app.rb
@@ -132,7 +132,7 @@ module Shipit
       if enterprise?
         options = options.reverse_merge(
           api_endpoint: api_endpoint,
-          web_endpoint: web_endpoint
+          web_endpoint: web_endpoint,
         )
       end
       client = Octokit::Client.new(options)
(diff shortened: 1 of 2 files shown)


MXfive:
CLA Signed, not sure if it's waiting for manual review or just isn't updating here 🤔.

aquartier:
@MXfive do a force push. 😄

lib/shipit/github_app.rb
```diff
@@ -125,7 +125,7 @@ def enterprise?
     end
 
     def new_client(options = {})
-      options.reverse_merge(api_endpoint: api_endpoint) if api_endpoint
+      options.reverse_merge!(api_endpoint: api_endpoint) if api_endpoint
```
casperisfine: `options = options.reverse_merge` would be preferable, as to avoid mutating the argument.

casperisfine:
@MXfive thanks for the patch, could you try adding a regression test as well?

MXfive:
@casperisfine Good to go 😄

PR #949, opened 2019-10-11
Added note about new task visibility to the tasks section.

ajshepley:
Normally it shouldn't be an issue, but I think it's worth mentioning in case a user doesn't see their tasks. This also will help them lean towards thinking something is wrong if they refresh and still don't see their new tasks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -522,6 +522,8 @@ tasks:
       - name: POD_ID
 ```
 
+New tasks will become available once the stack has updated and refreshed. You can press the `Refresh statuses & commits` button to manually trigger this refresh.
+
 <h3 id="custom-links">Custom Links</h3>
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -522,6 +522,8 @@ tasks:
       - name: POD_ID
 ```
 
+New tasks will become available once the stack has updated and refreshed. You can press the `Refresh statuses & commits` button to manually trigger this refresh.
```
DazWorrall: Is this correct? I thought tasks were loaded from the last successfully deployed commit. Not 100% sure here, but if I'm right then triggering a refresh will not be sufficient.

ajshepley: Perhaps I should have write "Once merged to Shipit and deployed, new tasks..."?

PR #959, opened 2019-11-27
Set initial status when GH deploy is created

JackTLi:
Right now, when we create a CommitDeployment, we don't set an initial status so the first status that gets pushed to GitHub is the _next_ status that the commit deployment receives after creation. This is a small tweak that will immediately push the current status when the commit deployment is created.

### DIFF
--- a/app/models/shipit/commit_deployment_status.rb
+++ b/app/models/shipit/commit_deployment_status.rb
@@ -45,6 +45,7 @@ module Shipit
       client.create_deployment_status(
         commit_deployment.api_url,
         status,
+        accept: 'application/vnd.github.flash-preview+json',
         target_url: url_helpers.stack_deploy_url(stack, task),
         description: description,
       )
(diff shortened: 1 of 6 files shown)


DazWorrall:
🎩 'd locally, LGTM:

Screenshot 2019-11-28 at 10 22 38

JackTLi:
Ty for the 🎩 , still dealing with osx upgrade 💀

DazWorrall:
As a follow up to this: I would like us to set the first status with state `in_progress` rather than `pending`, so we can treat _deployments_ in the `pending` state (including those with no statuses) as not yet active. This will require [adding a header](https://developer.github.com/v3/repos/deployments/#create-a-deployment-status) to opt in though:

> To use the `in_progress` and `queued` states, you must provide the `application/vnd.github.flash-preview+json` custom media type.

JackTLi:
👍 should be able to do so by changing the mapping here https://github.com/Shopify/shipit-engine/blob/master/app/models/shipit/deploy.rb#L17

JackTLi:
@DazWorrall would you be able to help me 🎩 again? I don't have a good local dev setup to 🎩 this effectively 😅

DazWorrall:
Progress @JackTLi :

Screenshot 2019-12-02 at 17 50 15

Looks like there's a translation to add, and some tests to fix!

JackTLi:
🙏 will fix, wanted to see if the accept header was used properly. thanks!

PR #961, opened 2019-12-12
Allow registering additional webhook handlers

benlangfeld:
Allows wrapping applications to extend Shipit functionality based on triggers from Github.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+* Allow applications to register extra handlers for Github webhooks. (#961)
+
 # 0.29.0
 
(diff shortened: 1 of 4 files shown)


benlangfeld:
I'm thinking the restriction on re-registration and extending the default handled types makes this approach unfeasible, and it may be better to simply allow registering procs which will be invoked for all web hooks in addition to the default ones, and force routing logic into the consumer code. I'll prep a commit to switch to that approach for consideration.

casperisfine:
LGTM. Can you squash your commits please?

benlangfeld:
@casperisfine Ready for merge.

PR #964, opened 2019-12-13
show absolute time of older tasks instead of "X days ago" relative time

kevinodotnet:
When debugging issues that were potentially caused by a deployment of code, it can be hard to determine which task/deploy happened within a given time range. The default time display can easily fold an entire day of deployments into the same "X days ago" window.

This PR updates the existing use of `timeago_tag` to have it print out the absolute time the task was created after `12.hours` have elapsed. For recent deployments, this will maintain current behaviour. If a user is looking at the timeline beyond 12 hours, the absolute time will be shown instead, which is probably more useful.

image

I thought about making this a configurable Stack settings option, but decided to put a smaller/simpler PR up for review first instead, to see how the community feels about this.

Local display example:

### DIFF
--- a/app/views/shipit/tasks/_task.html.erb
+++ b/app/views/shipit/tasks/_task.html.erb
@@ -24,7 +24,7 @@
         <% if read_only %>
           <span class="utc-timecode">on <%= task.created_at.strftime('%Y-%m-%d %H:%M:%S') %> UTC</span>
         <% else %>
-          <%= timeago_tag(task.created_at, force: true) %>
+          <%= timeago_tag(task.created_at, force: false, date_only: false, limit: 12.hours.ago, format: '%Y-%m-%d %H:%M %Z') %>
         <% end %>
       </p>
     </div>


casperisfine:
That's a hard one. It's not the first time I hear complaints about this. 

In some cases (typically after the fact investigation) a full, UTC timestamp is indeed much better.

In some others `15 minutes ago` is much more readable than a UTC timestamp, especially if your timezone is far from UTC.

We used to have both: https://github.com/Shopify/shipit-engine/pull/157, I don't remember why it's not the case anymore, probably because it takes a lot of space in the UI.

I'm honestly really unsure about what's best here, I'm totally torn, however I don't think changing the display after 12 hours like this is a good solution.

Right now the full timestamp is accessible as a `title` property, but it means the browser search function won't work.

Maybe some toggle button, like the emergency mode one ?

kevinodotnet:
> Right now the full timestamp is accessible as a title property, but it means the browser search function won't work.

TIL!

I'll back this out and try a toggle. Thanks!

casperisfine:
Not sure if you are working on this or no, but after some thoughts I think the toggle is the way to go.

Implementation wise, I think it should be handled purely client side, as otherwise we'd hit lots of cache troubles.

UI wise, I think clicking on a timestamp should  toggle between relative and absolute, and the last state should be stored in `localStorage` (or similar).

PR #965, opened 2019-12-16
Drops duplicate Continuous Delivery jobs

benlangfeld:
This job does not explicitly set the lock timeout, and so uses the default of `0`. It also defaults to `on_duplicate :retry`, despite being executed every minute by the cron job. It is highly likely that many of these jobs are duplicates that are being retried, though this is entirely unnecessary because the job runs every minute to catch up anyway, causing a very high job load and many reported exceptions to exception tracking services.

In our instance of Shipit, we're seeing thousands of this exception occurring every week.


benlangfeld:
That CI failure is unrelated to this work.

PR #968, opened 2019-12-18
Fix NoMethod error `stacks` in WebHooksController

indiebrain:
When receiving messages for repositories we've not yet present in the
shipit datastore, `Repositorty.from_github_repo_name` will be nil -
resulting in a NoMethod error:

```
NoMethodError undefined method `stacks' for nil:NilClass
```


test/models/shipit/unknown_repository_test.rb
```diff
@@ -0,0 +1,29 @@
...
+    test "name" do
+      assert_equal "unknown-name", @unknown_repository.name
+    end
+
+    test "stacks" do
+      assert_equal Stack.none, @unknown_repository.stacks
+    end
+
+    test "http_url" do
+      assert_equal "https://github.com/#{@unknown_repository.owner}/#{@unknown_repository.name}", @unknown_repository.http_url
```
benlangfeld: Isn’t this test a little useless, being identical to the implementation?

app/models/shipit/unknown_repository.rb
```diff
@@ -0,0 +1,23 @@
...
+    def stacks
+      Stack.none
+    end
+
+    def http_url
+      Shipit.github.url("#{owner}/#{name}")
+    end
+
+    def git_url
+      "https://#{Shipit.github.domain}/#{owner}/#{name}.git"
```
casperisfine: I'm not sure a NullObject is warranted here. I really like that pattern, but in this case the null values might lead to weird errors, such as trying to clone https://github.com/unknown-owner/unknown-name (which by the way could totally be registered by a malicious user and lead to arbitrary code being executed on your Shipit install).

I'd rather have `Handler#repository` return `nil` or raise an error for now.

If some pattern emerge, and a NullObject would streamline the implementation, then it will always be time to reconsider.

indiebrain: How about something a little less clever than the null object? https://github.com/Shopify/shipit-engine/pull/968/commits/fda7be6c60b2439f84df3cdde8b03bc378ee7d50

test/models/shipit/unknown_repository_test.rb
indiebrain: Removed in https://github.com/Shopify/shipit-engine/commit/fda7be6c60b2439f84df3cdde8b03bc378ee7d50

indiebrain:
As always, thanks for the conversation and review!

PR #969, opened 2019-12-19
Permit archival of stacks

benlangfeld:
Archiving a stack hides it from the list of stacks in the index view (but not from API clients) and locks the stack for deployment.

This is useful in itself for uncluttering the UI while retaining data, but is also intended to form part of the future addition of an ability to auto-provision stacks based on open PRs; when a PR is closed, its stack would be archived, and resources allocated to deploying the stack (eg Kubernetes namespaces) would be destroyed at this point.


test/dummy/db/schema.rb
```diff
@@ -218,6 +218,8 @@
     t.boolean "merge_queue_enabled", default: false, null: false
     t.datetime "last_deployed_at"
     t.integer "repository_id", null: false
+    t.datetime "archived_since"
```
indiebrain: Had you considered `archived_on` to keep with rails conventions for date/time stamps?

benlangfeld: I considered it but maintained consistency with fields like `locked_since` instead. I can do whatever @byroot prefersn

benlangfeld: I considered it but maintained consistency with fields like `locked_since` instead. I can do whatever @byroot prefers.

app/models/shipit/stack.rb
```diff
@@ -58,6 +58,10 @@ def self.repo(full_name)
       Repository.find_by(owner: repo_owner, name: repo_name).stacks
     end
 
+    def self.not_archived
```
indiebrain: I know we're modeling behavior of _archival_ and we've been using that nomenclature in our stories, but had you considered language which provides a clear affirmative/negative case; `active` vs `inactive` perhaps? I can't think of a word to describe something that _isn't_ archived other than `unarchived` which - in all the ambiguity of the English language is more likely to mean an object which was _once_ archived, but is no longer.

Inverted attribute names tend to trip me up when reading code - in general, I find it much easier to grok what _is_ rather than what _isn't_ - especially when we start applying additional predicates.

benlangfeld: I could do hidden/visible if you think that’s better.

viniciusgama: also, this could be turned into a scope

test/dummy/db/schema.rb
casperisfine: "archived_since" is fine.

app/models/shipit/stack.rb
benlangfeld: Scopes are an anti-pattern.

casperisfine: Not sure what you mean by that, but regardless they're used across the codebase, so might as well stay consistent.

benlangfeld: Done. I can explain later in more detail, but scopes are pointless at a minimum. Anywho, it’s now a scope :)

benlangfeld:
@casperisfine Is there anything else you need here for this to be acceptable?

benlangfeld:
Happy new year @casperisfine !

I wonder if you might have time in the near future to integrate this, and a few other PRs which we have open and I believe are mergeable?

* https://github.com/Shopify/shipit-engine/pull/970

casperisfine:
@benlangfeld apologies, I was in vacation for the last 2 weeks. I'll go over all these PRs today.

benlangfeld:
Currently the assumption is that if one must recover a stack, one either knows the URL or would essentially perform maintenance at a rails console.

I’m happy to implement something more, like a listing of archived stacks. Would the new stack form behaviour you described block merging this, or would an “archived stacks” listing be sufficient?

casperisfine:
> Would the new stack form behaviour you described block merging this, or would an “archived stacks” listing be sufficient?

A listing would be enough for me to merge. I'm just worried about users being confused by the UI telling them the stack already exist, and them not seeing it in the main listing.

benlangfeld:
@casperisfine I believe I've addressed your concern and this is ready for review once again.

app/assets/stylesheets/_pages/_stacks.scss
```diff
@@ -49,6 +49,15 @@
       padding-left: 30px;
       margin-left: -30px;
     }
+    .archived {
+      // From FontAwesome, license @ https://fontawesome.com/license
```
casperisfine: This leads to the private license, did you mean to link to https://fontawesome.com/license/free ?

I'd like to be sure there's no licensing issue here.

benlangfeld: Sorry, yes. Clarified. They say this:

> Downloaded Font Awesome Free files already contain embedded comments with sufficient attribution, so you shouldn't need to do anything additional when using these files normally.

But that's incorrect, the SVG doesn't include a comment.

casperisfine:
Thanks.

PR #972, opened 2019-12-23
Extract webhook handlers

indiebrain:
This moves the Webhook `Handler`s out of the `WebhooksController`; providing several benefits:

1. Webhook `Handler`s now live in their own module - they're not tied to the Webhooks Controller anymore.
2. The Webooks registry is out of the Rails autoloader path. In development it was quite often the case the the Rails reloader would re-initialize the `extra_handlers` attribute of the `WebhooksController` to an empty array - effectively un-registering all of the custom handlers. This avoids the case where a user would expect a custom handler to fire in development, but it has been effectively unregistered by the autoloader. 
3. The interface between the defaut, `shipit-engine` supplied, and user defined handlers has been unified. A handler need only implement the `call`able interface.
4. Since handlers need only respond to `call` registering handlers becomes more flexible. Any PORO which implements a `call` method, however any other `call`able Ruby object will work just fine: e.g a block, a proc, a lambda, etc. 
5. The registry allows for multiple handlers to be registered for the same event.


lib/shipit/webhooks.rb
```diff
@@ -0,0 +1,32 @@
...
+        'membership' => [Handlers::MembershipHandler],
+        'check_suite' => [Handlers::CheckSuiteHandler],
+      }
+    end
+
+    def self.handlers
+      @handlers ||= reset_handler_registry
+    end
+
+    def self.reset_handler_registry
```
casperisfine: Couple style nitpicks here.

I tend to prefer to reopen the singleton class (`class << self`) rather than use `def self.method`, simply because it's much easier for full text search. Here since it's purely a static module, you can also use `extend self`, some people frown upon it, but I IMO it's fine.

Second is naming, I think `reset_handlers!`, would be a better name.

lib/shipit.rb
```diff
@@ -50,6 +50,7 @@
 require 'shipit/github_http_cache_middleware'
 require 'shipit/cast_value'
 require 'shipit/line_buffer'
+require 'shipit/webhooks'
```
casperisfine: I actually don't like that this is being moved out of the autoloading path, I see no reason for it. You say:

> The Webooks registry is out of the Rails autoloader path. In development it was quite often the case the the Rails reloader would re-initialize the extra_handlers attribute of the WebhooksController to an empty array - effectively un-registering all of the custom handlers. 

But that's because it should be registered from a `to_prepare` callback. Granted that's an easy mistake to make.

lib/shipit/webhooks.rb
indiebrain: Addressed by e90f024 and baa9422  respectively.

lib/shipit.rb
indiebrain: Addressed by e03cefa

PR #975, opened 2020-01-07
Removes mistaken entries from schema.rb

benlangfeld:
These leaked in by mistake in a prior PR. Will be submitted properly in the future.


benlangfeld:
@casperisfine Apologies, but some of the previously merged PRs we submitted contained mistaken schema.rb entries from migrations that will be included in a future PR. This cleans up that mistake.

casperisfine:
No worries, I should have spotted it during review. 

Thanks for fixing it.

PR #967, opened 2019-12-17
Add Repository admin UI

indiebrain:
Allow users to manage `Repository` objects and their stacks.

PR #976, opened 2020-01-07
Implement ephemeral review stacks for pull requests

benlangfeld:
This replaces https://github.com/Shopify/shipit-engine/pull/967 and is a work in progress; most of the business logic for this lives in our downstream application and is yet to be extracted.


casperisfine:
Thanks for publicizing your draft, so far the implementation seem quite simple.

One thing though, I'm not so sure this configuration should be done in the repository interface, by default most shipit configuring is done through `shipit.yml`, and I feel like this one should be too.

That being said, `shipit.yml` is the Stack configuration, and this indeed belongs to the repo, so I'm not 100% sure about it, it might end up being a bit weird. What do you think ?

benlangfeld:
> That being said, `shipit.yml` is the Stack configuration, and this indeed belongs to the repo, so I'm not 100% sure about it, it might end up being a bit weird. What do you think?

Indeed, that's what we found. Much of the time we need to react to events from GH before we have the config synced, and there is of course the question of which branch's config applies to the repository as a whole.

There's also the question of who should be able to modify this configuration; in our use-case, a release engineering team is supposed to control this, so keeping it in the database allows for hiding it behind a future granular permissions system in Shipit.

Lastly, this configuration might need to be changed more rapidly than a git workflow (with branch rebasing, etc) would allow for, for operational reasons (shutting off review stack provisioning quickly for a repository whose deployments are broken/consuming many resources/etc).

Overall, we felt this was the best approach.

benlangfeld:
Sorry, I hit the ready for review button by mistake! Closing this for now.

PR #977, opened 2020-01-08
Expose GitHub repository info to tasks

DazWorrall:
A use case has arose for this.

### DIFF
--- a/lib/shipit/task_commands.rb
+++ b/lib/shipit/task_commands.rb
@@ -41,6 +41,8 @@ module Shipit
         'IGNORED_SAFETIES' => @task.ignored_safeties? ? '1' : '0',
         'GIT_COMMITTER_NAME' => @task.user&.name || Shipit.committer_name,
         'GIT_COMMITTER_EMAIL' => @task.user&.email || Shipit.committer_email,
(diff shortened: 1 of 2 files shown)


lib/shipit/task_commands.rb
```diff
@@ -41,6 +41,8 @@ def env
         'IGNORED_SAFETIES' => @task.ignored_safeties? ? '1' : '0',
         'GIT_COMMITTER_NAME' => @task.user&.name || Shipit.committer_name,
         'GIT_COMMITTER_EMAIL' => @task.user&.email || Shipit.committer_email,
+        'GITHUB_REPO_OWNER' => @stack.repository.owner,
+        'GITHUB_REPO_NAME' => @stack.repository.name,
```
casperisfine: Hum, I wonder if it should be two variables, or just one with the full name.

DazWorrall: It's easy enough for users to compose if they choose, rather than having to figure out some bash foo to split them.

DazWorrall: We could do both of course, and add a `GITHUB_REPO_FULL_NAME` or something.

PR #979, opened 2020-01-08
Add concept of extra variables to stack

viniciusgama:
While implementing the dynamic provision of stacks we realized we would also need ways to dynamically assign variables to the stack in order to have those available during the execution of the deployment script. Since the stacks are dynamically provisioned based on pull requests, using a `shipit.yml` would not be possible.

What we are proposing here is the inclusion of the concept `ExtraVariable` to add another option in which we could define `environment` variables.

These `extra variables` can be managed using the stack settings page and the idea is that they take precedence over any other environment variable.

This is related to a few PRs that were opened recently. #979 and #967 are among them

### DIFF
--- a/README.md
+++ b/README.md
@@ -625,6 +625,8 @@ These variables are accessible only during deploys and rollback:
 * `REVISION`: the git SHA of the revision that must be deployed in production
 * `SHA`: alias for REVISION
 
+On top of that, any other extra variable defined in the stack configuration page will also be made available during your deploy and rollback script execution.
+
 <h2 id="configuring-providers">Configuring providers</h2>
 
 ### Heroku
--- a/app/assets/javascripts/shipit/stacks.js.coffee
+++ b/app/assets/javascripts/shipit/stacks.js.coffee
@@ -22,6 +22,49 @@ $document.on 'click', '.action-set-release-status', (event) ->
     $deploy.attr('data-release-status', last_status.state)
   )
 
+$document.on 'click', '#add-new-variable', (event) ->
+  event.preventDefault()
+  formFields = $(event.target).closest('form').find('.field-list')[0]
+  template = document.querySelector("template[id='stack-variable-form-input']")
+
+  formFields.appendChild(document.importNode(template.content, true))
+
+$document.on 'change', '#stack_extra_variables__key', (event) ->
+  validateExtraVariables()
+
+$document.on 'click', '#remove-variable', (event) ->
+  event.preventDefault()
+  formField = $(event.target).closest('div')[0]
+  formField.parentNode.removeChild(formField)
+  validateExtraVariables()
+
+validateExtraVariables = () ->
+  $keysInputs = document.querySelectorAll('input#stack_extra_variables__key')
+
+  allKeys = []
+  $keysInputs.forEach (item) ->
+    allKeys.push(item.value)
+
+  $keysInputs.forEach (item) ->
+    $submit = $(item).closest('form').find("input[type='submit']")
+    if (allOcurrences(allKeys, item.value).length > 1)
+      $submit.addClass('btn--disabled')
+      $submit.prop('disabled', true)
+      $(item).addClass('field-invalid')
+    else
+      $submit.removeClass('btn--disabled')
+      $submit.prop('disabled', false)
+      $(item).removeClass('field-invalid')
+
+allOcurrences = (items, element) ->
+    indexes = []
+    idx = items.indexOf(element)
+    while (idx != -1)
+      indexes.push(idx)
+      idx = items.indexOf(element, idx + 1)
+
+    indexes
+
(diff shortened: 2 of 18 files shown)


casperisfine:
I'll be honest I don't like that feature. We decided quite early not to do it, and to keep this kind of config in the `shipit.yml`. So no it's kind of a duplicated feature, and it goes against the original design goals.

I'm curious what kind of variables you need to store in there.

viniciusgama:
> I'll be honest I don't like that feature. We decided quite early not to do it, and to keep this kind of config in the `shipit.yml`. So no it's kind of a duplicated feature, and it goes against the original design goals.
> 
> I'm curious what kind of variables you need to store in there.

Right now we need it to configure to which cluster the stack will be deployed continuously deployed to. And this is defined on PR base. In the future we also envision some extra config being passed on based on  extra information we find in the PR. We also thought about simply adding this as an attribute to the stack but decided to implement in a way that is not tied to our specific use case and so making it more generic would it could be accepted upstream.

casperisfine:
> Right now we need it to configure to which cluster the stack will be deployed continuously deployed to. And this is defined on PR base.

Ok, that's what I figured. The thing is that Shipit try to be as agnostic as possible, so generally what we recommend is for the deploy command to handle this sort of stateful considerations, e.g. talk to some other service to know what to do.

benlangfeld:
@casperisfine Here we're trying to have our Shipit installation orchestrate and abstract more, rather than requiring all projects that it deploys to repeat large implementations of logic. In this case, we're round-robin assigning stacks provisioned for PRs between multiple Kubernetes clusters to spread load - this state must be stored somewhere it can be queried; the shipit.yml file in the repository is not appropriate.

casperisfine:
> this state must be stored somewhere it can be queried; the shipit.yml file in the repository is not appropriate.

Yes I understood your need. We have similar ones for some of our stacks, and we're solving it outside of Shipit. 

I understand that it's tempting to make Shipit a bit more stateful and smart, but it goes against its very first design principle.

You can also solve it as a private shipit extension, or use consistent hashing (`hash(branch_name) % number_of_clusters`)

> The other use case that @viniciusgama mentioned is passing the list of PR labels through to deployments,

That again I think is going farther that what I envisioned, and I opens a huge can of worm:

  - Did the labels change since the stack was created ? Does that mean they need to be synced ? 
  - Then what about other PRs attributes ? 

When we initially talked about this feature, what I had in mind was a simple state machine:

  - provision
  - deploy
  - decommission

Each of this being just a list of executed commands.

IMHO if the deploy script need to read some Pull Request attributes to know how to deploy, it should query the GitHub API itself. So I'd be open to expose the PR id or url, but not arbitrary attributes.

benlangfeld:
> I understand that it's tempting to make Shipit a bit more stateful and smart, but it goes against its very first design principle.

What is that design principle?

> You can also solve it as a private shipit extension

That's probably what we'll do. Our intention has been to contribute as much of what we're doing as possible upstream, on the basis that we're almost certainly not the only ones who could make use of it. If there's some overriding principle like minimalism, it would be interesting to understand that vision so we can more accurately judge what might be appropriate for inclusion in shipit and what we must keep to ourselves.

> IMHO if the deploy script need to read some Pull Request attributes to know how to deploy, it should query the GitHub API itself. So I'd be open to expose the PR id or url, but not arbitrary attributes.

That means the deployment would require credentials to access the Github API, which is not practical.

Thanks for the quick feedback @casperisfine , it's most appreciated!

casperisfine:
> What is that design principle?

Right sorry, I sometimes kinda forgot that it's not really known outside of Shopify.

Shipit was built because our previous tool was requiring lots of configuration inside the tool itself, which was very painful. As a result one of the core principle is to keep as little configuration as possible in Shipit itself. The general idea is that you can entirely wipe the database and just recreate the list of stacks (without any specific config) and it should mostly works.

Of course there are small deviations to this, but it's mostly respected.

Another big tenant is that it should have no idea about how to deploy repos. It's the repo that should know all that. 

> That's probably what we'll do. Our intention has been to contribute as much of what we're doing as possible upstream, on the basis that we're almost certainly not the only ones who could make use of it.

Yeah I saw that and I appreciate it. TBH the idea of performing actions around a pull request kinda grew on me, and I think it's a good feature fo shipit. I'd just like to keep it stateless.

So hopefully we can find a solution to merge most of the features you need, but I'm really not keen on merging the stateful parts.

> That means the deployment would require credentials to access the Github API, which is not practical.

It already has them: https://github.com/Shopify/shipit-engine/blob/826e578c000d138471860b06bf62b53eb1359bf1/lib/shipit/commands.rb#L24 (before you ask they're short lived tokens).

viniciusgama:
thanks @casperisfine . We will try a different approach. I'll close this one now.

PR #980, opened 2020-01-09
Emits status update hook when deploy is triggered

benlangfeld:
Prior to this change, a hook was only being emitted when a task changed status, which was not considered to be the case when the task was created (status nil -> pending). This change simply causes the hook to be emitted on task creation, as this is effectively a status change. We're using this downstream to modify a deployment (manipulate its environment) in between it being triggered and executing.

### DIFF
--- a/app/models/shipit/task.rb
+++ b/app/models/shipit/task.rb
@@ -39,7 +39,8 @@ module Shipit
 
     after_save :record_status_change
     after_create :prevent_concurrency, unless: :allow_concurrency?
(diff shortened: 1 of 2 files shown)


casperisfine:
Sorry, it's been a long while I worked on this, could you explain (briefly is fine) the reason for this change?

benlangfeld:
@casperisfine Description updated.

casperisfine:
Thanks. Makes sense now.

PR #981, opened 2020-01-09
Exposes DEPLOY_URL to deployments

benlangfeld:


casperisfine:
I suppose you'll use this to store state ? 😉

benlangfeld:
@casperisfine Are this and maybe #980 good to go?

casperisfine:
Right sorry, there was no description so I though you wanted to modify it.

This is good to go, I'll check #980

PR #982, opened 2020-01-09
Create a GitHub deployment for the batch head

DazWorrall:
It's not currently possible for other apps to identify deployments associated with the head of the deploy batch - we create one for the PR head of any included pull requests, so context appears in the PR timeline, but this is not sufficient to to discover the full range of commits. This PR creates a deployment for the head of the batch being deployed, in addition to the deployments for each PR.

Some refactoring was required here:

* Remove the `CommitDeployment#commit` association (I will follow up with a schema cleanup once this has shipped successfully, though I may need a release cutting first)
* Add `sha` to `CommitDeployment`
* Handle deployment creation in a job rather than inline - this code path now queries GitHub for PR head refs, and that should not be done in the critical path of creating a `Deploy`

I have 🎩ed this locally.

### DIFF
--- a/db/migrate/20200109132519_add_sha_to_commit_deployments.rb
+++ b/db/migrate/20200109132519_add_sha_to_commit_deployments.rb
@@ -1,0 +1,5 @@
+class AddShaToCommitDeployments < ActiveRecord::Migration[6.0]
+  def change
+    add_column :commit_deployments, :sha, :string, limit: 40
+  end
+end
--- a/app/jobs/shipit/create_deployments_for_task_job.rb
+++ b/app/jobs/shipit/create_deployments_for_task_job.rb
@@ -1,0 +1,27 @@
+module Shipit
+  class CreateDeploymentsForTaskJob < BackgroundJob
+    include BackgroundJob::Unique
+
+    queue_as :default
+
+    def perform(task)
+      # Create one deployment for the head of the batch
+      task.commit_deployments.create!(sha: task.until_commit.sha)
+
+      # Create one for each pull request in the batch, to give feedback on the PR timeline
+      task.commits.select(&:pull_request?).each do |commit|
+        task.commit_deployments.create!(sha: pull_request_head_for_commit(commit))
+      end
+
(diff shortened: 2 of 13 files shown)


db/migrate/20200109132519_add_sha_to_commit_deployments.rb
```diff
@@ -0,0 +1,7 @@
+class AddShaToCommitDeployments < ActiveRecord::Migration[6.0]
+  def change
+    add_column :commit_deployments, :sha, :string, limit: 40
+    add_index :commit_deployments, :sha
+    add_index :commit_deployments, [:task_id, :sha], unique: true
```
casperisfine: Couple questions here:

  - Is this index (these indexes) actually needed? When do we query `commit_deployments` ?
  - Should we backfill the existing records, and then make `sha` `NOT NULL` ?

DazWorrall: > Is this index (these indexes) actually needed? When do we query commit_deployments ?

Good question. I’ve added these for parity with the existing association (which has equivalent indexes on `commit_id`), but will do some investigation into their use.

> Should we backfill the existing records, and then make sha NOT NULL ?

I plan this for the follow up - if I do that now, we will run the migration first and backfill the existing data, but any deployments created between `db:migrate` and the new code deploy would be created with a null sha. For safety we need to backfill the data after the code has been deployed.

DazWorrall: Ok I did some digging - the `commit_id` index is used only in one place that I can see, in `DestroyStackJob`. I changed that to search by task_id instead, and have removed that new index.

I'm not sure what the unique index was guarding - it doesn't look like we were catching `RecordNotUnique` anywhere in the call path, and I don't see any exceptions in Bugsnag. I've removed the new unique index, we can add it back if it causes problems in production.

casperisfine: 👍

PR #983, opened 2020-01-10
Fix CommitDeploymentStatus GitHub creation job

DazWorrall:
I'm debugging a race where the initial status is not created on GitHub after the deployment is created. Using `commit_deployment` as the argument here looks like a typo to me, surely it should be itself. The warning log is for further debugging if I've got this wrong. 

Should I deploy our instance from this branch and 🎩 before merging?


DazWorrall:
Actually I kinda see what this was trying to do here. Not sure how I've broken it yet 🤔

DazWorrall:
Disregard, I found the real issue in #984.

PR #989, opened 2020-01-13
Fetch 100 commit statuses at a time

DazWorrall:
Core deploys are getting stuck because Shipit thinks some statuses are missing, despite the statuses on GitHub being present and passing. After some debugging it seems that this is a [lack of] pagination problem - a core commit has something like ~50 status records now, and the default fetch size is 30. This is a hotfix to unblock deploys, pending some actual pagination support.

### DIFF
--- a/app/models/shipit/commit.rb
+++ b/app/models/shipit/commit.rb
@@ -128,7 +128,9 @@ module Shipit
     end
 
     def refresh_statuses!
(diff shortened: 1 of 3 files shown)


DazWorrall:
> ~50+ statuses? Wow

They aren't unique _contexts_, you create a new status record and the state of the latest one wins for any given context. So a typical context has at least 2 status records (pending -> success).

ajshepley:
Ah, I see. That makes more sense.

PR #990, opened 2020-01-13
Setup GitHub actions for CI

DazWorrall:
Motivated by the irritatingly long wait for Travis agents earlier, this replicates the existing CI setup on GitHub actions.

On the plus side, these jobs generally start much quicker than on Travis, and are well integrated into GitHub. On the down side, Actions themselves are still reasonably new and are changing rapidly, and annoyingly they [do not support Yaml anchors/aliases](https://github.community/t5/GitHub-Actions/Support-for-YAML-anchors/td-p/30336), so this config file is enormous and full of duplicated things.

I suggest we run this alongside Travis for a while and see how we fare.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,165 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    name: Lint Ruby
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Ruby 2.5
+      uses: actions/setup-ruby@v1
+      with:
+        ruby-version: 2.5.x
+    - uses: actions/cache@v1
+      with:
+        path: vendor/bundle
+        key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-gems-
+    - name: rubocop
+      run: |
+        sudo apt-get -yqq install libpq-dev libsqlite3-dev libmysqlclient-dev
+        gem install bundler:1.17.3
+        bundle config path vendor/bundle
+        bundle install --jobs 4 --retry 3
+        bundle exec rubocop
+
+  postgres:
+    name: Tests with Postgres
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,165 @@
...
+    - uses: actions/cache@v1
+      with:
+        path: vendor/bundle
+        key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-gems-
+    - name: rubocop
+      run: |
+        sudo apt-get -yqq install libpq-dev libsqlite3-dev libmysqlclient-dev
+        gem install bundler:1.17.3
```
wvanbergen: should be update to bundler 2?

.github/workflows/main.yml
```diff
@@ -0,0 +1,165 @@
...
+        ruby-version: 2.5.x
+    - uses: actions/cache@v1
+      with:
+        path: vendor/bundle
+        key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-gems-
+    - name: rubocop
+      run: |
+        sudo apt-get -yqq install libpq-dev libsqlite3-dev libmysqlclient-dev
```
wvanbergen: I wonder whether we can get away with not installing those, and only installing a partial bundler for Rubocop.

Not sure if the time to optimize this is worth it though. Feel free to leave it as is.

.github/workflows/main.yml
```diff
@@ -0,0 +1,165 @@
+name: CI
+
+on: [push, pull_request]
```
casperisfine: Do we need both? I don't know GH actions well, but all statuses seem duplicated on this PR.

DazWorrall: `pull_request` is required to build PRs from forks I think, I agree the duplicated statuses are annoying - afaict there's no option to squash these :/

.github/workflows/main.yml
DazWorrall: Probably, I will look at that separately though.

PR #991, opened 2020-01-15
Permit locking stacks with machine-readable reason

benlangfeld:
This allows (efficiently) querying for stacks locked for this reason while retaining a user-friendly presentation.

### DIFF
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -42,6 +42,7 @@ module Shipit
     validates_associated :repository
 
     scope :not_archived, -> { where(archived_since: nil) }
+    scope :locked_because, ->(reason_code) { where(lock_reason_code: reason_code) }
 
     def repository
       super || build_repository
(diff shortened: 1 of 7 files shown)


casperisfine:
Could you explain the use case ? We have a bunch of robots locking and unlocking stacks and we never felt the need for this.

I'm not totally opposed to it, but without an use case my first impression is that it's not particularly useful.

Also the locks API would need to be updated: https://github.com/Shopify/shipit-engine/blob/1c5ec450c98ad4a311f3ecfb43219117b4fcf972/app/controllers/shipit/api/locks_controller.rb#L6-L16

benlangfeld:
We’re locking stacks to indicate that they are in a pending state (waiting on external resource provisioning). We need to be able to get a list of those to provision them. The human-readable reason text may change, and the column is not indexed só query performance is poor.

casperisfine:
> We’re locking stacks to indicate that they are in a pending state (waiting on external resource provisioning)

That's part of your effort to implement automatic stack provisioning for pull requests right?

Wouldn't this need be better fulfilled by a dedicated state machine on the stack?

benlangfeld:
We could do a state machine, sure, I just assumed this change was less invasive and less controversial than doing that. Let me know what you prefer.

benlangfeld:
@casperisfine Could you confirm that you would accept a state machine on Stack so that we can step it through the various provisioning phases that are necessary for the ReviewStack concept?

casperisfine:
Well, it's hard to say without seeing the implementation. But as I imagine it, yes a state machine seems like a nicer solution that piggy backing onto locks.

benlangfeld:
This is superseded by https://github.com/Shopify/shipit-engine/pull/1102.

PR #994, opened 2020-01-20
Fix exception when commitless task is active

benlangfeld:
#974 made the commit range optional on tasks but didn't take this into account.


casperisfine:
Thanks!

PR #995, opened 2020-01-22
Relax gemfile ruby version detection regex.

ajshepley:
We try to detect a ruby version in Gemfiles and delete the line.

This was working for, say, `ruby '2.5.3'`, but not for lines with parenthesis like `ruby('2.5.3')` or `ruby ('2.5.3')`.

This results in deploys failing for those builds:
https://shipit.shopify.io/shopify/identity_client/production/deploys/914537

This PR changes the regex to match those cases as well.

### DIFF
--- a/app/models/shipit/deploy_spec/bundler_discovery.rb
+++ b/app/models/shipit/deploy_spec/bundler_discovery.rb
@@ -33,9 +33,9 @@ module Shipit
         # Heroku apps often specify a ruby version.
         if /darwin/i.match?(RUBY_PLATFORM)
           # OSX is nitpicky about the -i.
-          %q(/usr/bin/sed -i '' '/^ruby\s/d' Gemfile)
+          %q(/usr/bin/sed -i '' '/^ruby(\s|\()/d' Gemfile)
         else
-          %q(sed -i '/^ruby\s/d' Gemfile)
+          %q(sed -i '/^ruby(\s|\()/d' Gemfile)
         end
       end
 


JackTLi:
Do we need to change the above regex? And I think `ruby (2.5.3)` is invalid, since there can't be a space between a method name and the opening parenthesis

sorryeh:
I think it would be nice to include `ruby('2.5.3')` we are growing in size and more people with have linter autofixing these type of issues.. Without knowledge that valid ruby code will break the prod deploy pipeline.. It's a big waste of time:
- me --> messing around trying to understand what I did wrong, then try 1-2 things by pushing and changing git tags + releases on github) since I can't know I fixed it or not before pushing for a prod deploy :\
- everyone that then reads the developers-talk/shipping-pipeline posts
- then recognizing & explaining the issue

I'd rather see a change that adds the `(` to the regex :P @JackTLi

app/models/shipit/deploy_spec/bundler_discovery.rb
```diff
@@ -35,7 +35,7 @@ def remove_ruby_version_from_gemfile
           # OSX is nitpicky about the -i.
           %q(/usr/bin/sed -i '' '/^ruby\s/d' Gemfile)
         else
-          %q(sed -i '/^ruby\s/d' Gemfile)
+          %q(sed -i '/^ruby(\s|\s\(|\()/d' Gemfile)
```
sorryeh:
```suggestion
          %q(sed -i '/^ruby(\s|\()/d' Gemfile)
```

https://shopify.github.io/ruby-style-guide/
`I Do not put a space between a method name and the opening parenthesis.`

ajshepley: Ah. I wasn't quite sure. Makes this a bit simpler :)

JackTLi:
Yeah I'm for it, my comment was for the `Do not put a space between a method name and the opening parenthesis.` case

PR #996, opened 2020-01-23
Truncate commit messages if they are to big to fit the database column

casperisfine:
If the commit message is ridiculously long, the insertion might fail, and it breaks GItHub syncrhonization.

We could increase the storage, but the value would be low as we never really use more than the first few lines of the message.

### DIFF
--- a/app/models/shipit/commit.rb
+++ b/app/models/shipit/commit.rb
@@ -104,6 +104,14 @@ module Shipit
       )
     end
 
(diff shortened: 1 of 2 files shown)


casperisfine:
Damn it, this fails with postgres

DazWorrall:
Test failures look legit.

casperisfine:
I pushed another fix and now it's green.

PR #1000, opened 2020-01-28
Prefer locked stacks in MergeStatusController

DazWorrall:
Ref: https://github.com/Shopify/shipit/issues/934

If multiple stacks match the branch it makes more sense to prefer any locked stacks even if they doesn't use the merge queue, to give the user context that merging may not be desirable.

### DIFF
--- a/app/controllers/shipit/merge_status_controller.rb
+++ b/app/controllers/shipit/merge_status_controller.rb
@@ -60,12 +60,14 @@ module Shipit
       @stack ||= if params[:stack_id]
         Stack.from_param!(params[:stack_id])
       else
-        scope = Stack.order(merge_queue_enabled: :desc, id: :asc).includes(:repository).where(
-          repositories: {
-            owner: referrer_parser.repo_owner,
-            name: referrer_parser.repo_name,
-          },
-        )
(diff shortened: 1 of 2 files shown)


casperisfine:
Not sure I'm a fan of that. If someone creates a personal test stack  for an important repo, it will be displayed.

DazWorrall:
> Not sure I'm a fan of that. If someone creates a personal test stack for an important repo, it will be displayed.

If they also lock it, yes. It's already true if people create a test stack and enable the merge queue, though that's admittedly less likely.

The behaviour here is already poorly defined, I'm not convinced this makes things worse - as David experienced, if there is a locked stack and merging is undesirable, that's quite important context. If someone creates a test stack there's no reason to lock it (so long as CD is disabled).

casperisfine:
>  If someone creates a test stack there's no reason to lock it 

I did plenty of that when I was developing various notifications etc.

But if you feel like this change is positive, then :shipit:

DazWorrall:
I'll wait for some other opinions, happy to be outvoted here.

JackTLi:
I don’t love this behaviour but it makes sense in the context of the problem we’re solving

PR #1001, opened 2020-01-28
Only unlock a stack when the unlock button is used

benlangfeld:
Without this change, any modification to the configuration of a locked stack will implicitly unlock it.


benlangfeld:
The failures are fixed at https://github.com/Shopify/shipit-engine/pull/1003

casperisfine:
Are you sure about that one?

IIRC the whole form is submitted, including the existing `lock_reason`, no?

benlangfeld:
Yes I’m sure. The problem is there are multiple forms that hit this action. Try switching continuous delivery on for a locked stack and you’ll see the problem.

PR #1002, opened 2020-01-28
Translates stack navigation strings

benlangfeld:


benlangfeld:
The failures are fixed at https://github.com/Shopify/shipit-engine/pull/1003

PR #1009, opened 2020-01-30
Added status mappings for faulty and timedout shipit statuses.

ajshepley:
I think these statuses should both count as "finisher" statuses, e.g. ones that result in a deployment status being created in a 'done' state such as failure or error.

This will also let us rule out skipping these statuses as one potential source of the current missing deploy completion statuses.

### DIFF
--- a/app/models/shipit/deploy.rb
+++ b/app/models/shipit/deploy.rb
@@ -21,7 +21,9 @@ module Shipit
         'pending' => 'pending',
         'running' => 'in_progress',
         'failed' => 'failure',
+        'timedout' => 'failure',
         'success' => 'success',
+        'faulty' => 'error',
         'error' => 'error',
         'aborted' => 'error',
       }.freeze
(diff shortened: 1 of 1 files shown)


app/models/shipit/deploy.rb
```diff
@@ -21,7 +21,9 @@ class Deploy < Task
         'pending' => 'pending',
         'running' => 'in_progress',
         'failed' => 'failure',
+        'timedout' => 'failure',
         'success' => 'success',
+        'faulty' => 'error',
```
ajshepley: I'm not sure I understand how and when we differentiate error vs. failure.

IMO something involving user action or general deploy state being 'bad' is how I view error, and the deploy operation failing to finish is what I call 'failure'.

JackTLi: Yeah that breakdown makes sense. Specifically error would be things that are not retriable, and failure are things that are imo.

DazWorrall: I tend do think of them the other way around to you Jack 😅 Let's not stress over the semantics for now, it's more important to plug the gap in statuses.

ajshepley:
There was a timeout this morning, but it was before the logging changes were deployed.

Another deploy failure happened with a faulty status, which made me think it would be useful to set those statuses rather than clutter the not-found's with them. I haven't gone further into investigating that one yet.

I'm also going to log the API url here, I think it could be useful.

PR #1015, opened 2020-02-05
Allow host application to define system-wide links to stack header

indiebrain:
Give the mounting application a way to define additional links which
apply to stacks across the system rather than per-stack.


indiebrain:
Thanks for the review @casperisfine !

PR #1018, opened 2020-02-10
Add ability to perform variable substitution to stack links in shipit.yml

indiebrain:
...when at least one task has been performed on the stack - IE when
we have something we can use to grab the environment variables.

This maintains the previous functionality of the `shipit.yml` to
declare static links for stacks like:

```yml
links:
  logs: https://logs.application.com
```

but also affords consumers the ability to define template strings in
the `shipit.yaml` like:

```yml
links:
  logs: https://logs.$ENVIRONMENT.application.com
```

Which, if the stack's environment is set to `production` will render a
link to `https://logs.production.application.com` in the stack's header.

### DIFF
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -43,6 +43,17 @@ module Shipit
 
     scope :not_archived, -> { where(archived_since: nil) }
 
+    def env
+      {
+        'ENVIRONMENT' => environment,
+        'LAST_DEPLOYED_SHA' => last_deployed_commit.sha,
+        'GITHUB_REPO_OWNER' => repository.owner,
+        'GITHUB_REPO_NAME' => repository.name,
+        'DEPLOY_URL' => deploy_url,
+        'BRANCH' => branch,
+      }
+    end
+
     def repository
       super || build_repository
     end
@@ -77,7 +88,7 @@ module Shipit
     validates :lock_reason, length: {maximum: 4096}
 
     serialize :cached_deploy_spec, DeploySpec
-    delegate :find_task_definition, :supports_rollback?, :links, :release_status?, :release_status_delay,
+    delegate :find_task_definition, :supports_rollback?, :release_status?, :release_status_delay,
              :release_status_context, :supports_fetch_deployed_revision?, to: :cached_deploy_spec, allow_nil: true
 
     def self.refresh_deployed_revisions
@@ -526,6 +537,13 @@ module Shipit
 
     def sync_github
       GithubSyncJob.perform_later(stack_id: id)
+    end
+
+    def links
+      links_spec = cached_deploy_spec&.links || {}
+      context = EnvironmentVariables.with(env)
+
+      links_spec.transform_values { |url| context.interpolate(url) }
     end
 
     private
--- a/lib/shipit/task_commands.rb
+++ b/lib/shipit/task_commands.rb
@@ -28,22 +28,20 @@ module Shipit
     end
 
     def env
-      super.merge(
-        'ENVIRONMENT' => @stack.environment,
-        'BRANCH' => @stack.branch,
-        'SHIPIT_USER' => "#{@task.author.login} (#{normalized_author_name}) via Shipit",
-        'EMAIL' => @task.author.email,
-        'BUNDLE_PATH' => Rails.root.join('data', 'bundler').to_s,
-        'SHIPIT_LINK' => @task.permalink,
-        'LAST_DEPLOYED_SHA' => @stack.last_deployed_commit.sha,
-        'TASK_ID' => @task.id.to_s,
-        'DEPLOY_URL' => @stack.deploy_url,
-        'IGNORED_SAFETIES' => @task.ignored_safeties? ? '1' : '0',
-        'GIT_COMMITTER_NAME' => @task.user&.name || Shipit.committer_name,
-        'GIT_COMMITTER_EMAIL' => @task.user&.email || Shipit.committer_email,
-        'GITHUB_REPO_OWNER' => @stack.repository.owner,
-        'GITHUB_REPO_NAME' => @stack.repository.name,
-      ).merge(deploy_spec.machine_env).merge(@task.env)
+      super
+        .merge(@stack.env)
+        .merge(
+          'SHIPIT_USER' => "#{@task.author.login} (#{normalized_author_name}) via Shipit",
+          'EMAIL' => @task.author.email,
+          'BUNDLE_PATH' => Rails.root.join('data', 'bundler').to_s,
+          'SHIPIT_LINK' => @task.permalink,
+          'TASK_ID' => @task.id.to_s,
+          'IGNORED_SAFETIES' => @task.ignored_safeties? ? '1' : '0',
+          'GIT_COMMITTER_NAME' => @task.user&.name || Shipit.committer_name,
+          'GIT_COMMITTER_EMAIL' => @task.user&.email || Shipit.committer_email,
+        )
+        .merge(deploy_spec.machine_env)
+        .merge(@task.env)
     end
 
     def checkout(commit)
--- a/test/dummy/db/schema.rb
+++ b/test/dummy/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_01_09_132519) do
+ActiveRecord::Schema.define(version: 2020_02_11_145825) do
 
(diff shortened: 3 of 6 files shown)


app/models/shipit/stack.rb
```diff
@@ -528,8 +528,25 @@ def sync_github
       GithubSyncJob.perform_later(stack_id: id)
     end
 
+    alias links_orig links
```
casperisfine: Rather than aliasing the method, you can remove it from the delegation list, and call `cached_deploy_spec.links`.

Also note that this `delegate` has a `allow_nil: true`, so you have to handle `cached_deploy_spec` being nil.

app/models/shipit/stack.rb
```diff
@@ -528,8 +528,25 @@ def sync_github
...
+    def links
+      persisted_tasks = tasks.select(&:persisted?)
+      return links_orig if persisted_tasks.blank?
+
+      variables = Shipit::TaskCommands.new(persisted_tasks.last).env
+      links_orig.each_with_object({}) do |(name, url), replaced_links|
```
casperisfine: `Hash#transform_values` would be much cleaner here.

app/models/shipit/stack.rb
```diff
@@ -528,8 +528,25 @@ def sync_github
       GithubSyncJob.perform_later(stack_id: id)
     end
 
+    alias links_orig links
+    def links
+      persisted_tasks = tasks.select(&:persisted?)
+      return links_orig if persisted_tasks.blank?
+
+      variables = Shipit::TaskCommands.new(persisted_tasks.last).env
```
casperisfine: I have to think about that one. This seems quite a hack.

What variable do you need to interpolate here? 

I'm not against that interpolation, but if it could be done without loading the last deploy it would be preferable.

app/models/shipit/stack.rb
```diff
@@ -528,8 +528,25 @@ def sync_github
...
+      links_orig.each_with_object({}) do |(name, url), replaced_links|
+        replaced_links[name] = replace_variables(url, variables)
+      end
+    end
+
     private
 
+    def replace_variables(string, variables)
```
casperisfine: We already have code for that: https://github.com/Shopify/shipit-engine/blob/4192f63825a6e7cf8f437cb2fa0f626ad60f4cf2/lib/shipit/command.rb#L48-L55

Might just need to be extracted to it's own module.

app/models/shipit/stack.rb
indiebrain: Perhaps a practical example will be helpful. 

We run and deploy to a number of internal Kubernetes clusters. We have a custom task which chooses the _least active_ cluster and provisions a Kubernetes namespace for a stack when a new stack is added. We register internal hooks to deployments which capture information about the kubernetes cluster to which a stack is deployed and make that knowledge available via the deployment/task's `#env`. 

This information is useful to us because the cluster to which a stack is deployed determines the URL of its telemetry services - for example the pattern to get to a stack's logs looks something like `https://logs.${ENVIRONMENT}.application.${CLUSTER}.domain.com`. Does that help at least put a frame around the picture?

app/models/shipit/stack.rb
indiebrain: Wow, `Hash#transform_values` is much nicer. Thanks for the suggestion!

app/models/shipit/stack.rb
indiebrain: Addressed by https://github.com/Shopify/shipit-engine/pull/1018/commits/c2ec62ac53841a7e3d2f32e1530916a94428e964

app/models/shipit/stack.rb
indiebrain: Addressed by https://github.com/Shopify/shipit-engine/pull/1018/commits/ae342cc1e1b9e5447736efa569d88aa0b67e6ead

app/models/shipit/stack.rb
indiebrain: Addressed by https://github.com/Shopify/shipit-engine/pull/1018/commits/671d045de9eb4d64cd2e4dcc9e7bb8a9ac5cdf97

app/models/shipit/stack.rb
casperisfine: > We register internal hooks to deployments which capture information about the kubernetes cluster to which a stack is deployed and make that knowledge available via the deployment/task's `#env`.

Ok, so you use tasks to persist data?

It's quite a clever hack, but from a product design standpoint, using the last stack for interpolation would make for a very surprising behavior.

What if:

  - We use `Stack#env` for interpolation (doesn't exist yet, but easy to add).
  - You can then do this cluster selection in your host app in ruby, and store it either in the db or in `Rails.cache`.

indiebrain: > We use Stack#env for interpolation (doesn't exist yet, but easy to add).

Sounds like something I can chase down. Just want to make sure I'm on the same page. If this bit lives in `shipit-engine` what would your recommendation be to allow the host app - like mine - inject values into the stack's environment? Would it be appropriate to add another concept / association for these values to the Stack and use them to build the value of `Stack#env`? Perhaps setting one of these values in the host application looks something like `stack.envrionment_variable.create(name: "cluster", value: "cluster-1")`? Does this seem reasonable? Do you think the name `environment_variable` is appropriate here? If not, any suggestions?

casperisfine: IMO the best way is:

```ruby
module ExtraStackEnv
  def env
    env = super
    # add whatever you want here
    env
  end
end

Shipit::Stack.prepend(ExtraStackEnv)
```

indiebrain: Addressed by https://github.com/Shopify/shipit-engine/pull/1018/commits/bc0fd9d791729cfc0eb497c764f53df53112b38e and https://github.com/Shopify/shipit-engine/pull/1018/commits/a210081966d860400107e216f80446ab2f1f366a

indiebrain:
@casperisfine something like this agreeable?

casperisfine:
Yes. SOrry I started reviewing this hours ago and somehow got distracted. Reviwing it gain .

indiebrain:
👍 thanks!

lib/shipit/task_commands.rb
```diff
@@ -28,22 +28,20 @@ def steps
...
+          'BUNDLE_PATH' => Rails.root.join('data', 'bundler').to_s,
+          'SHIPIT_LINK' => @task.permalink,
+          'TASK_ID' => @task.id.to_s,
+          'IGNORED_SAFETIES' => @task.ignored_safeties? ? '1' : '0',
+          'GIT_COMMITTER_NAME' => @task.user&.name || Shipit.committer_name,
+          'GIT_COMMITTER_EMAIL' => @task.user&.email || Shipit.committer_email,
+        )
+        .merge(deploy_spec.machine_env)
+        .merge(@stack.env)
+        .merge(@task.env)
```
casperisfine: My last worry with the PR is here.

Environment variables precedence is important.

I don't think `Stack#env` should be able to override `Task#env`.

Intuitively I'd say the order should be:

  - `Commands#env` (`super` in this context, mostly comes from the global config file)
  - `Stack#env`
  - `TaskCommands#env`
  - `DeploySpec#machine_env`
  - `Task#env`

But I'm open to diverging opinions.

indiebrain: Entirely reasonable. Thanks for the feedback! Something like https://github.com/Shopify/shipit-engine/pull/1018/commits/09cc2991332935537453101c112f6855faf72a84 instead, then?

casperisfine: Yup 👍

indiebrain: Many, many thanks, friend. 🍻

test/dummy/db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_01_09_132519) do
+ActiveRecord::Schema.define(version: 2020_02_11_145825) do
```
indiebrain: Whoops. @casperisfine . Not sure why these changes weren't removed in https://github.com/powerhome/shipit-engine/commit/a210081966d860400107e216f80446ab2f1f366a

casperisfine: I should have seen it. No big deal.

indiebrain: My apologies.

casperisfine: No worries. I'll fix that on master.

indiebrain: You're a gentle-person and a scholar.

PR #1019, opened 2020-02-10
Render shipit.yml links as a menu-list in the nav

indiebrain:
For stacks with more than one or two links the navigation bar starts
to wrap links. We _might_ prevent this by rendering the links as a
sub-list instead.


casperisfine:
Could you showcase the difference with screenshots please?

indiebrain:
Aye.

Before
-------
before-2

After
------
after-2

casperisfine:
Thanks.

indiebrain:
Thank you!

PR #1022, opened 2020-02-12
docs: add new github related script variables to readme

damianpursey:
I reference this section of the docs often, would like this to be more visible for others as well.


casperisfine:
Thanks!

PR #1023, opened 2020-02-17
Optionally tracks pull request information on stack

viniciusgama:
For stacks that are dynamically created we might want to assign the pull request that triggered the creation.

This PR is adding the concept of a review request that can be assign to a stack if needed. I'm using the existing `PullRequest` model in order to take advantage of the code that creates and refreshes the data pulling extra info from Github.

### DIFF
--- a/app/jobs/shipit/refresh_pull_request_job.rb
+++ b/app/jobs/shipit/refresh_pull_request_job.rb
@@ -3,6 +3,8 @@ module Shipit
     queue_as :default
 
     def perform(pull_request)
+      raise Stack::NotYetSynced if pull_request.stack.commits.blank?
+
       pull_request.refresh!
       MergePullRequestsJob.perform_later(pull_request.stack)
     end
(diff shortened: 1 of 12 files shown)


viniciusgama:
@byroot would you take a look?

casperisfine:
Sorry, I've been a bit swamped the last few days and I forgot to answer this PR.

I don't see anything particularly wrong in the diff, even though, I'm a bit worried about the recrudescence of optional relations. I wonder if "review stacks" wouldn't be better implemented as a Stack subclass with STI.

But regardless, at this stage all this is very abstract, I can't really review a change like this without reviewing the feature as a whole. Don't be afraid to open a PR with the whole feature as a draft and we can split it later if needed.

If I can't see the final "flow" of the feature, I can't tell if this PR is a good implementation for it.

benlangfeld:
This is superseded by https://github.com/Shopify/shipit-engine/pull/1102.

benlangfeld:
Please close this one @viniciusgama

PR #1027, opened 2020-02-21
enable webhook consumers to detect anonymous user

exterm:
Keep https://github.com/Shopify/spy/issues/6843 from happening again.

Also requires a change to spy though to handle this attribute.

Paired with @lastgabs on this

### DIFF
--- a/test/models/hook_test.rb
+++ b/test/models/hook_test.rb
@@ -67,5 +67,12 @@ module Shipit
     ensure
         Shipit.internal_hook_receivers = original_receivers
     end
+
+    test ".coerce_payload coerces anonymous user correctly" do
+      locked_stack = Stack.first
+      locked_stack.lock("Some Reason", nil)
+      serialized = Hook.coerce_payload(stack: locked_stack)
+      assert_json("stack.lock_author.anonymous", true, document: serialized)
+    end
   end
 end
--- a/test/unit/user_serializer_test.rb
+++ b/test/unit/user_serializer_test.rb
@@ -1,0 +1,13 @@
+require 'test_helper'
+
+module Shipit
+  class UserSerializerTest < ActiveSupport::TestCase
+    test 'includes anonymous key' do
+      user = User.new
+      serializer = ActiveModel::Serializer.serializer_for(user)
+      assert_equal UserSerializer, serializer
+      serialized = serializer.new(user).to_json
+      assert_json("anonymous", false, document: serialized)
+    end
+  end
+end
--- a/app/serializers/shipit/user_serializer.rb
+++ b/app/serializers/shipit/user_serializer.rb
@@ -1,5 +1,9 @@
 module Shipit
   class UserSerializer < ActiveModel::Serializer
-    attributes :id, :name, :email, :login, :avatar_url, :created_at, :updated_at, :github_id
+    attributes :id, :name, :email, :login, :avatar_url, :created_at, :updated_at, :github_id, :anonymous
+
+    def anonymous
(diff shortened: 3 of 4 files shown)


casperisfine:
I'm fine with the feature, but don't like the implementation.

IMO all you have to do is to expose the existing `logged_in?` attribute.

No need for a new serializer, just add it to the existing user serializer.

exterm:
Oh, I did not see the `logged_in?` attribute. Let me check that.

exterm:
I actually find `logged_in?` misleading. It's not as if we had a user that wasn't logged in. We don't have a user.

I'll reuse `logged_in?` but expose as `anonymous`.

exterm:
Unless someone can suggest a better name.

casperisfine:
I don't particularly mind the naming, It's mostly about the extra serializer.

test/unit/user_serializer_test.rb
```diff
@@ -0,0 +1,15 @@
...
+module Shipit
+  class UserSerializerTest < ActiveSupport::TestCase
+    test 'includes anonymous key' do
+      serialized = UserSerializer.new(User.new).to_json
+      assert_equal false, JSON.parse(serialized)["anonymous"]
+    end
+
+    test 'sets anonymous to true for anonymous users' do
+      serialized = UserSerializer.new(AnonymousUser.new).to_json
+      assert_equal true, JSON.parse(serialized)["anonymous"]
```
exterm: @byroot the test proves that it works, doesn't it?

casperisfine: No, it proves that `UserSerializer` renders what you want, not that `UserSerializer` is used to serialize `AnonymousUser`.

AMS use a bunch of name based heuristic, so I assume that if I did add this empty `AnonymousUserSerializer` it was to comply to these.

If you want to prove it's useless (which is possible) then you need a controller test.

casperisfine: Controller or webhook etc. By that I meant a higher level test that does go through the serializer selection logic.

app/models/shipit/anonymous_user.rb
```diff
@@ -1,5 +1,9 @@
 module Shipit
   class AnonymousUser
+    def serializer_class
+      UserSerializer
+    end
```
exterm: What do you think @byroot ? I'd rather not have it in here but compared to making this inherit from `User` I prefer this solution.

Actually I'm slightly confused about how serialization worked previously for this so I'll look into it more

byroot: In general I'm not a huge fan of this AMS API, but it's fine. 

I'm like to get rid of AMS anyway.

test/models/hook_test.rb
```diff
@@ -67,5 +67,10 @@ class HookTest < ActiveSupport::TestCase
     ensure
         Shipit.internal_hook_receivers = original_receivers
     end
+
+    test ".coerce_payload coerces anonymous user correctly" do
+      serialized = Hook.coerce_payload(lock_author: AnonymousUser.new)
+      assert_equal true, JSON.parse(serialized)["lock_author"]["anonymous"]
```
byroot: Minor nitpick, but IIRC we have an `assert_json` test helper which allow this to be much cleaner.

exterm: fixed.

lastgabs:
You forgot to change the test name 😆

exterm:
I couldn't think of a better one 🤔 

Did you have a better idea? We can still make that change

PR #1030, opened 2020-02-26
Add job for reaping dead but running deploys.

ajshepley:
When infra disruptions occur (e.g. SQL) , tasks can get stuck in the Running state, but will return false when calling `alive?`. These deploys are then seemingly permanently stuck, requiring manual intervention.

This PR creates a job that will check for such stuck deploys and terminate them, while logging the information for further followup if necessary.

Scheduling of the job will be hooked up in a future PR, and can be done integration-side as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+* Added a new job `ReapDeadDeploymentsJob` to support cleanup of jobs that are stuck in running, but report as dead.
+  * This job runs every minute via the cron rake task.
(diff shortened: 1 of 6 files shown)


ajshepley:
@casperisfine I was planning to enqueue in our internal integration, as we were aiming for running it every ~15 minutes. I'm not sure if there's a way to have that kind of interval as a setting in cron.rake.

casperisfine:
> I'm not sure if there's a way to have that kind of interval as a setting in cron.rake.

I think it would be perfectly fine to put it in the minutely cron. That job should be rather fast and harmless.

It's fine if we run it less often in our custom integration, but the open source version should have it one way or the other.

PR #1032, opened 2020-03-03
Fix rake CVE-2020-8130

josacar:
I dunno if we should set something less restrictive or not or if `~> 12.0` will break in some scenarios.

Anyone with some input on this?


dwradcliffe:
Fixed in https://github.com/Shopify/shipit-engine/pull/1034
Thanks anyway @josacar!

PR #1034, opened 2020-03-04
Bump rake to 12.3.3

DazWorrall:
Fixes https://github.com/advisories/GHSA-jppv-gw3r-w3q8


casperisfine:
LGTM. We could even just remove any version lock on rake.

DazWorrall:
> We could even just remove any version lock on rake.

Done 👍

PR #1036, opened 2020-03-10
Always set ignored_safeties when emergency mode is enabled

actgardner:
We're looking for ways to work [Binary Authorization breakglass](https://cloud.google.com/binary-authorization/docs/deploying-containers#override_a_policy) into the existing deploy workflow. One proposal was to trigger breakglass on an emergency deploy (see https://github.com/Shopify/production-registry/pull/245)

A deploy blocked by Binary Authorization may have a green CI status. For this reason, we need to either:
- always set `ignored_safeties` when the Emergency Deploy mode is enabled, even if CI is passing
- create a new database field and environment variable which is similar to `ignored_safeties`, but which is always set when Emergency Deploy mode is enabled

This PR takes the first approach and changes the behaviour of the existing field.


DazWorrall:
Hmm, I worry this would create false positives in [github-certification](https://github.com/Shopify/github-certification/blob/d4e0a8811f0020a57aad0d3e9d3bd42a9c711889/app/models/rules/respect_safeties.rb#L6), @casperisfine wdyt?

casperisfine:
> I worry this would create false positives in github-certification

Yeah, I had the same worry.

I think overall this would be better handled with a boolean deploy variable. I also wonder if that makes sense as an open source feature.

actgardner:
> I think overall this would be better handled with a boolean deploy variable. I also wonder if that makes sense as an open source feature.

Like this? https://github.com/Shopify/shipit/pull/1037

casperisfine:
Yes, exactly what I had in mind.

PR #1037, opened 2020-03-12
release-gem: abort deploy if allowed_push_host property is not set

casperisfine:
When you generate a gem with bundler `spec.metadata['allowed_push_host']` is set to `"TODO: Set to 'http://mygemserver.com'"` to prevent publishing private gems onto the public registry.

The problem is that if people remove that property entirely, or craft the gemspec entirely by hand, it then default to `https://rubygems.org`.

So this PR change the `release-gem` script, so that it requires explicit whitelisting of `https://rubygems`.

cc @lastgabs @EiNSTeiN- @wildmaples

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+* Prevent gem released unless the `allowed_push_host` metadata is set in the gemspec (#1037).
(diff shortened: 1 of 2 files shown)


casperisfine:
NB: it's kind of proof of concept quality. Opinions on the error message are most welcome.

also cc @DazWorrall @wvanbergen

EiNSTeiN-:
I think this is a good approach, it will force developers to be explicit about where they push.

PR #1039, opened 2020-03-17
Added guard against 401 and log message when a user email cannot be retrieved

ajshepley:
We are already guarding against 403s and 404s, but apparently the credentials can be in a state where 401s are returned when trying to retrieve emails at all. This didn't show up in testing at the time.

This adds 401s to the rescue clause, as well as a log message so users with problematic app permission and email API settings can be more easily tracked down.

### DIFF
--- a/app/models/shipit/user.rb
+++ b/app/models/shipit/user.rb
@@ -140,8 +140,9 @@ module Shipit
                   .sort_by { |e| e.primary ? 0 : 1 }
                   .map(&:email)
                   .find { |e| email_valid_and_preferred?(e) }
-      rescue Octokit::NotFound, Octokit::Forbidden
+      rescue Octokit::NotFound, Octokit::Forbidden, Octokit::Unauthorized
         # If the user hasn't agreed to the necessary permission, we can't access their private emails.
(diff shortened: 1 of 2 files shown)


app/models/shipit/user.rb
```diff
@@ -140,8 +140,9 @@ def appropriate_email_for(github_user)
                   .sort_by { |e| e.primary ? 0 : 1 }
                   .map(&:email)
                   .find { |e| email_valid_and_preferred?(e) }
-      rescue Octokit::NotFound, Octokit::Forbidden
+      rescue Octokit::NotFound, Octokit::Forbidden, Octokit::Unauthorized
         # If the user hasn't agreed to the necessary permission, we can't access their private emails.
+        Rails.logger.warn("Failed to retrieve emails for user '#{name}'")
```
DazWorrall: `login` or `id` will be more stable I reckon.

ajshepley: I realized that it makes more sense to use the github_user that was passed in, so I've deferred to that and am now using the same name logic as is used in the record setter elsewhere.

PR #1042, opened 2020-03-18
More job timeout tweaks

DazWorrall:
I took a closer look through our exception reports and StatsD metrics over the last 24 hours, the aim with these changes is to further reduce the `ConcurrentJobError` exceptions. The timeout values are based on the max observed runtime for these jobs in the last 24 hours, and the lock timeout is intended to add a bit of a buffer for outliers.

### DIFF
--- a/app/jobs/shipit/create_on_github_job.rb
+++ b/app/jobs/shipit/create_on_github_job.rb
@@ -5,7 +5,8 @@ module Shipit
     queue_as :default
 
     # We observe that some objects regularly take longer than the default 10 seconds to create, e.g. deployments
-    self.timeout = 20
+    self.timeout = 40
+    self.lock_timeout = 20
 
     def perform(record)
       record.create_on_github!
--- a/app/jobs/shipit/deferred_touch_job.rb
+++ b/app/jobs/shipit/deferred_touch_job.rb
@@ -4,6 +4,9 @@ module Shipit
 
     queue_as :default
 
+    self.timeout = 30
+    self.lock_timeout = 15
+
     def perform
       DeferredTouch.touch_now!
     end


benlangfeld:
This turned https://github.com/Shopify/shipit-engine/issues/1049 into a problem. While it seems that the duplicate queueing has always happened, this timeout increase means that it is now almost certain that the rehydrated model object (instance of a `CommitDeployment`) instantiated in https://github.com/rails/rails/blob/40b7d93c5bf04974cfd0201a8e391c499c046f0b/activejob/lib/active_job/execution.rb#L24) is stale by the time the `perform` method is invoked, since the wait on the mutex is now longer than the period over which the job will retry with Sidekiq defaults.

Before this timeout increase, the window during which the first job could complete and set `github_id` while a waiting second job has a stale reference without that value was small. Now it's large, and we're creating duplicate data on Github.

I can imagine a few possible solutions:

1. Reload the object from the database at the start of `CreateOnGithub#perform`.
1. Reload all ActiveRecord objects from the database after getting a lock in `BackgroundJob::Unique#acquire_lock`.
1. Move lock acquisition to `around_execute` so that this happens before the arguments are deserialized.

What is your preference for a fix @DazWorrall ?

DazWorrall:
Nice debugging @benlangfeld. I think 3 > 2 > 1 in terms of preference of those fixes - ideally this is transparent to the job itself, and we do not need to hit the database more than is necessary.

benlangfeld:
@DazWorrall Option 3 proposed @ https://discuss.rubyonrails.org/t/activejob-before-perform-callback/74137

benlangfeld:
@DazWorral that post hasn’t gained any responses. Do you have a Rails core contact internally at Shopify that you could sell the idea to?

casperisfine:
We have a bunch yeah. I'll forward them your post. We'll see.

casperisfine:
But even if this is accepted, it will take a while before we can use it, so I'd recommend to not wait on such ticket.

benlangfeld:
@casperisfine We're using this workaround for now: https://github.com/powerhome/shipit-engine/commit/55fd47ceb195ae7cab46e16edc7222ba8ebfe018

casperisfine:
sounds good.

DazWorrall:
Sorry for the silence here @benlangfeld, I was away for a few days. FTR I'd accept a fix like that here while upstream thinks about better hooks for our use case, wdyt @casperisfine ?

casperisfine:
I talked with @rafaelfranca about it:

> Ah. He posted in the talk category instead of the core. I’d never see that. He is asking for the around_execution callback we have in core. I was avoiding to implement that since I found no compelling use case. I’ll study that case and see if makes sense to expose it

> One alternative implementation is override the perform_now method and run the lock around the super class method

> That is how we implement in core

```ruby
module ExecutionCallbacks
  extend ActiveSupport::Concern

  included do
    define_callbacks :execution, skip_after_callbacks_if_terminated: true
  end

  module ClassMethods
    def before_execution(*filters, &block)
      set_callback(:execution, :before, *filters, &block)
    end

    def around_execution(*filters, &block)
      set_callback(:execution, :around, *filters, &block)
    end

    def after_execution(*filters, &block)
      set_callback(:execution, :after, *filters, &block)
    end
  end

  def perform_now
    run_callbacks(:execution) do
      super
    end
  end
end
```

So yeah, we can either use that patch in Shipit and keep an eye on upstream Rails activity, or just go with the simpler `.reload` workaround.

I'd tend to go with the later as it's more robust to future changes, but can be talked into the former.

benlangfeld:
> Ah. He posted in the talk category instead of the core. I’d never see that.

Apologies, I thought that was the appropriate etiquette, core being an inappropriate place for me to post. My mistake :)

> I'd tend to go with the later as it's more robust to future changes, but can be talked into the former.

I have the same preference, for the same reason.

https://github.com/Shopify/shipit-engine/pull/1054

Issue #1024, opened 2020-02-19
/events URL returns no content in 0.31.0


pfhayes:
We have just upgraded to 0.31.0 from 0.26.0 to pick up the recent GitHub API change.

The installation runs fine for several hours but then all requests will start timing out

It appears to be because an unbounded number of Pubsubstub connections are being created. I see many repeated events of the form
```
Feb 19 03:36:14 ip-172-31-37-230 env[1235]: I, [2020-02-19T03:36:14.855652 #1709]  INFO -- : [Pubsubstub::Subscription] Connecting client #7815470111917132360 (stack.1)
```

followed by an ever-growing number of repeated events of the form
```
Feb 19 03:39:19 ip-172-31-37-230 env[1235]: D, [2020-02-19T03:39:19.247287 #1728] DEBUG -- : [Pubsubstub::Subscription] Sending event #1582083559247 to client #7268750048476322816
```

This appears to be because the requests to /events?channels%5B%5D=stack.1 hang and return no content. In 0.26.0 the requests return content of the form:
```
id: 1582083655404
event: heartbeat
retry: 10000
data: ping
```

Reverting to 0.26.0 appears to fix the issue

Please let me know if I can provide any additional information to help diagnose the issue

casperisfine:
That's probably caused by the pubsubstub update, but it's been a while so I don't quite remember what we changed exactly.

Ideally, If you can, I recommend to run pubsubstub in it's own process as described here: https://github.com/Shopify/shipit-engine/issues/496#issuecomment-351429330 (it's really how it's meant to be used, the integrated form is for ease of use but isn't ideal at all).

Alternatively, it also has a degraded polling mode you can enable with: `Pubsubstub.use_persistent_connections = false`. Unless you have a lots of users, it should be more than enough.

I'd be happy to dig into this problem but, I wouldn't know where to start without a way to reproduce the problem, there is just so many moving parts it could be anything.

pfhayes:
Thanks, I tried this and it does appear to be working better now

pfhayes:
I spoke too soon, I'm still seeing the behaviour where the /events URL misbehaves even when running pubsubstub as a separate process

`Pubsubstub.use_persistent_connections = false` does appear to patch the problem

casperisfine:
That's weird. Cloud you post more logs?

I'd like to fix this issue if I can reproduce it.

Also I'm interested to know how you run pubsubstub externally. With puma? What version ? How many threads? 

[...]

PR #1047, opened 2020-03-27
Various fixes for the Rails template

Sam-Killgallon:
I recently set up ship and came across a few things that required debugging in order to get it running

1. Shipit uses the `redis-rails` gem for caching which doesn't support cache_versioning, so the app refuses to start. Rails ships with a `redis_cache_store` that does support this, but I didn't have the time to fully test it to know if its a drop in replacement, so I just disabled versioning for now https://guides.rubyonrails.org/caching_with_rails.html#activesupport-cache-rediscachestore
Screenshot 2020-03-27 at 7 04 53 pm

2. I had some confusion getting Shipit to talk to Github because the provided secrets file used the same environment variable for the `app_id` and the `installation_id` even though they are different values

3. Pubsubstub seems to cause the whole app to hang after a little while (Discussed https://github.com/Shopify/shipit-engine/issues/1024). I appreciate in that thread the decision hadn't been made whether or not to make this the default, so can drop this commit if preferred.

4. A slightly confusing message was output during the templating, didn't cause any issues but thought it was worth clearing up. Rails seems to already supply the `ruby <version>` line in the Gemfile, and the message comes from `Thor` because it's already there https://github.com/erikhuda/thor/blob/da1e707627578eacbf7c6336ed1bb22fb3abfb86/lib/thor/actions/inject_into_file.rb#L63
Screenshot 2020-03-27 at 7 06 41 pm

Thank you for your hard work on this app and I'm looking forward to trying it out 👍


casperisfine:
> 1. Shipit uses the `redis-rails` gem for caching which doesn't support cache_versioning, so the app refuses to start.

Yeah, we no longer need it. I simply forgot to remove it after the Rails 6.0 upgrade.

template.rb
```diff
@@ -18,7 +18,9 @@
 worker: bundle exec sidekiq -C config/sidekiq.yml
 CODE
 
+environment 'Pubsubstub.use_persistent_connections = false'
 environment 'config.cache_store = :redis_store, Shipit.redis_url.to_s, { expires_in: 90.minutes }', env: :production
```
casperisfine: Let's go all the way and use the `:redis_cache_store` shipped with Rails.

Sam-Killgallon:
Thanks for the feedback. I'll apply the changes when I get a spare moment in the next couple of days

template.rb
```diff
@@ -11,14 +11,16 @@
...
+gsub_file 'Gemfile', "# Use Redis adapter to run Action Cable in production", ''
+gsub_file 'Gemfile', "# gem 'redis'", "gem 'redis'"
 
 create_file 'Procfile', <<-CODE
 web: bundle exec rails s -p $PORT
 worker: bundle exec sidekiq -C config/sidekiq.yml
 CODE
 
-environment 'config.cache_store = :redis_store, Shipit.redis_url.to_s, { expires_in: 90.minutes }', env: :production
+environment 'Pubsubstub.use_persistent_connections = false'
```
casperisfine, line 22: Good idea. That's probably a better default for most people.

casperisfine:
Thanks!

PR #1051, opened 2020-04-03
Add SameSite middleware to Shipit

mctaylorpants:
I'd like to propose adding the [`SameSite` middleware from shopify-app](https://github.com/Shopify/shopify_app/blob/194c6b4d379a96e3c8d39ead771b56e020a140c1/lib/shopify_app/middleware/same_site_cookie_middleware.rb) into Shipit. This PR:

* Brings over the middleware (and `browser_sniffer` dependency)
* Makes the middleware opt-in by passing a `SHIPIT_ENABLE_SAMESITE_NONE` environment variable.

At Clio, we use a browser extension to show the merge status of a PR right in GitHub. This extension needs to make API requests back to Shipit, and [Chrome's recent rollout of the SameSite attribute](https://chromestatus.com/feature/5088147346030592) have prevented the extension from working reliably in Chrome. Adding this cookie attribute would fix the issue.


casperisfine:
Thanks fort hat PR, we actually have a same problem internally.

Is browser_sniffer really necessary? Does setting `SameSite=None` breaks older browsers?

Seems like it would be much simpler to unconditionally set it.

cc @EiNSTeiN- @jenshenny (if you could shed some light on all this).

mctaylorpants:
Some older versions of Chrome are incompatible with `SameSite=None`, as well as some older versions of Safari. [There's a list of known incompatibilities on the Chromium site](https://www.chromium.org/updates/same-site/incompatible-clients). However, guarding against the incompatibility may not be as important here since developers tend to keep their browsers up to date 😁

casperisfine:
> Some older versions of Chrome are incompatible with `SameSite=None`, as well as some older versions of Safari.

Damn it... I hoped not to have to rely on `browser_sniffer`.

benlangfeld:
@casperisfine The versions without support are super old. Given the target audience, I wouldn't worry about it.

jenshenny:
>The versions without support are super old. Given the target audience, I wouldn't worry about it.

I agree, don't think it's necessary to add the check if `browser_sniffer` isn't a dependency of this project.

mctaylorpants:
Works for me - thanks for chiming in! I've removed `browser_sniffer` and the user-agent checks from this PR.

mctaylorpants:
I think I may have hit a transient failure, [this run](https://github.com/Shopify/shipit-engine/pull/1051/checks?check_run_id=565640259) is failing on a spec that passes on the other builds. Is there a way of retrying it from your side or should I push up something to re-trigger the build?

lib/shipit/same_site_cookie_middleware.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    def initialize(app)
+      @app = app
+    end
+
+    def call(env)
+      status, headers, body = @app.call(env)
+
+      if headers && headers['Set-Cookie'] &&
+         env['SHIPIT_ENABLE_SAMESITE_NONE'].present? &&
```
casperisfine, line 13: Two questions:
  - Is there any reason to gate this behind an env variable?
  - If yes, then wouldn't it be simpler to just not insert the middleware?

lib/shipit/engine.rb
```diff
@@ -38,6 +38,8 @@ class Engine < ::Rails::Engine
         end
       end
 
+      app.config.middleware.insert_after(::Rack::Runtime, Shipit::SameSiteCookieMiddleware)
```
casperisfine, line 41: I'm always worried with `insert_after/before`, as middleware stacks are kind of a moving target. Do we need to be that specific about it's order?

mctaylorpants: My understanding is that this middleware needs to be at least added before `ActionDispatch::Cookies`. This is the PR that introduced this change in shopify_app: https://github.com/Shopify/shopify_app/pull/898

lib/shipit/same_site_cookie_middleware.rb
mctaylorpants: Fair point! To me it makes sense to make the feature opt-in, since some users may want to prevent their cookies from being used externally. I agree that the check would be better at the middleware level, though.

Happy to defer to your judgement on whether the check is necessary at all - let me know.

casperisfine: Right, since we never figured a way to make the browser extension available, I suppose the vast majority of users don't need this middleware.

Maybe we should keep it a host app responsability?

mctaylorpants: Agreed - I'll keep this an opt-in feature via the environment variable, but I'll have it enable the middleware instead of check on each request.

mctaylorpants:
@casperisfine have you had a chance to review the recent changes? Let me know if there's anything else I can improve.

casperisfine:
@mctaylorpants sorry, I have GitHub push notifications disabled because they are way too noisy. When you need a review you have to ask it by comment.

Looking now.

lib/shipit/same_site_cookie_middleware.rb
```diff
@@ -0,0 +1,30 @@
...
+      @app = app
+    end
+
+    def call(env)
+      status, headers, body = @app.call(env)
+
+      if headers && headers['Set-Cookie'] &&
+         Rack::Request.new(env).ssl?
+
+        set_cookies = headers['Set-Cookie']
```
casperisfine, line 15: Something to check here. What if the response has multiple `Set-Cookie` headers? AFAIK it's possible right?

casperisfine: Looking at rack's implementation, seems like it can indeed be an array: https://github.com/rack/rack/blob/f9ad97fd69a6b3616d0a99e6bedcfb9de2f81f6c/lib/rack/utils.rb#L219-L258

mctaylorpants:
Thanks @casperisfine, good point about multiple cookies. I've added an explicit test to cover that behaviour.

I think we're good to go!

casperisfine:
The test suite is failing for unrelated reasons. I'll try to fix master and then I'll merge.

casperisfine:
Ok, I fixed CI in https://github.com/Shopify/shipit-engine/pull/1056. Merging now.

PR #1052, opened 2020-04-07
Don't attempt to checkout unreachable commits

benlangfeld:
Where the last commit for a stack is unreachable (has been rebased off the branch, for example), we should not consider it for use when checking out the repo to, for example, fetch the deployed revision. We should only consider the last reachable commit.

I couldn't find any test coverage for this in a quick look.

### DIFF
--- a/lib/shipit/stack_commands.rb
+++ b/lib/shipit/stack_commands.rb
@@ -26,7 +26,7 @@ module Shipit
     end
 
     def fetch_deployed_revision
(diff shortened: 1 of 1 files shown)


benlangfeld:
> Except maybe if you force push an ancestor?

Precisely. Disgusting? Yes. Should break shipit? No.

casperisfine:
Ahah, I'm not judging, I do force push master from time to time ;)

PR #1054, opened 2020-04-14
Reload records to be created on Github

### LINKED ISSUE #1049: CreateOnGithub queued twice when creating a CommitDeployment
When a single `CommitDeployment` is created at the start of a deployment, we see it being queued twice:

Screenshot 2020-04-02 at 18 39 09

This results in multiple deployments being reported to Github:

Screenshot 2020-04-02 at 18 24 50

Setting `on_duplicate :drop` for `CreateOnGithub` would be insufficient, because there's nothing to guarantee that the jobs are processed concurrently.

Does anyone know what the source of this duplicate queueing might be?

benlangfeld:
This prevents us from duplicating data on Github.

This is a temporary solution until the lock can be moved from the `around_perform` to `around_execute` callback in order to operate prior to database queries, which is blocked on [support in Rails core](https://discuss.rubyonrails.org/t/activejob-before-perform-callback/74137). See https://github.com/Shopify/shipit-engine/pull/1042#issuecomment-608118022

Fixes #1049

### DIFF
--- a/app/jobs/shipit/create_on_github_job.rb
+++ b/app/jobs/shipit/create_on_github_job.rb
@@ -9,7 +9,7 @@ module Shipit
     self.lock_timeout = 20
 
     def perform(record)
-      record.create_on_github!
+      record.reload.create_on_github!
     end
   end
 end


DazWorrall:
@benlangfeld if you rebase we can get this merged.

benlangfeld:
Done @DazWorrall

PR #1055, opened 2020-04-14
Prevent immediately marking newly started deployments as dead

indiebrain:
There appears to be a race condition between the time when the
PerformTaskJob marks a task as "running" and when the newly added
RemoveDeadTasks job attempts to reap dead tasks.

In some cases a deployment is almost immediately marked as dead after
it starts. We believe this is because there critical section in which
PerformTaskJob marks a task as running and ReapDeadTaskJob is run in
which newly a task can be transitioned to a "running" status, but not
yet marked alive in which a newly created task will be considered a
"dead task" by the ReapDeaTasks job.

This changeset aims a two pronged approach at preventing this
race. First it makes the ReapDeadTasksJob less aggressive in its
consideration of tasks as "dead" - it will now only consider tasks in
a "running" or "aborted" state which are missing an "alive" monitor
AND are 5 at least minutes old.

Second, when PerformTaskJob starts working a task the task will ping
the "alive" monitor as soon as the task is marked as running. The
implementation of reporting that tasks are "alive" is somewhat
imperfect, in that it it doesn't happen until we attempt to capture
command output. There are therefore several things which happen
between starting the task and the first command happening, namely one
or two checks that the git repo is already checked out and a wait on a
cache lock, whose timeout is also 15 seconds. This means a large
opportunity for the "alive" status to expire. We should probably
ensure that the task is considered alive for a period long enough to
span this (maybe 30 seconds rather than 15), as well as ping after
each check of `fetched?`.

References
----------

- [Task marked as running, but not yet alive](https://github.com/Shopify/shipit-engine/blob/d48ceb4ffc21afbdf2701ce8c3a8936268afc3d5/app/jobs/shipit/perform_task_job.rb#L18)
- [Task first marked as alive](https://github.com/Shopify/shipit-engine/blob/d48ceb4ffc21afbdf2701ce8c3a8936268afc3d5/app/jobs/shipit/perform_task_job.rb#L81)

### DIFF
--- a/app/jobs/shipit/perform_task_job.rb
+++ b/app/jobs/shipit/perform_task_job.rb
@@ -15,6 +15,7 @@ module Shipit
     end
 
     def run
+      @task.ping
       @task.run!
       checkout_repository
       perform_task
@@ -62,7 +63,12 @@ module Shipit
 
     def checkout_repository
       unless @commands.fetched?(@task.until_commit).tap(&:run).success?
+        # acquire_git_cache_lock can take upto 15 seconds
+        # to process. Try to make sure that the job isn't
+        # marked dead while we attempt to acquire the lock.
+        @task.ping
         @task.acquire_git_cache_lock do
+          @task.ping
           unless @commands.fetched?(@task.until_commit).tap(&:run).success?
             capture! @commands.fetch
           end
--- a/app/jobs/shipit/reap_dead_tasks_job.rb
+++ b/app/jobs/shipit/reap_dead_tasks_job.rb
@@ -1,19 +1,20 @@
 module Shipit
   class ReapDeadTasksJob < BackgroundJob
     include BackgroundJob::Unique
-
-    ZOMBIE_STATES = %w(running aborting).freeze
-
     queue_as :default
 
     def perform
-      zombie_tasks = Task.where(status: ZOMBIE_STATES).reject(&:alive?)
-
       Rails.logger.info("Reaping #{zombie_tasks.size} running tasks.")
       zombie_tasks.each do |task|
         Rails.logger.info("Reaping task #{task.id}: #{task.title}")
         task.report_dead!
       end
+    end
+
+    private
+
+    def zombie_tasks
+      @zombie_tasks ||= Task.zombies
     end
   end
 end
--- a/app/models/shipit/task.rb
+++ b/app/models/shipit/task.rb
@@ -4,7 +4,7 @@ module Shipit
 
     ConcurrentTaskRunning = Class.new(StandardError)
 
-    PRESENCE_CHECK_TIMEOUT = 15
+    PRESENCE_CHECK_TIMEOUT = 30
     ACTIVE_STATUSES = %w(pending running aborting).freeze
     COMPLETED_STATUSES = %w(success flapping faulty validating).freeze
     UNSUCCESSFUL_STATUSES = %w(error failed aborted flapping timedout faulty).freeze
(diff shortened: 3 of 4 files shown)


indiebrain:
@casperisfine , and team, FWIW, we've been running this internally for a little over 24 hours. The last time we saw this error condition was a minute or so before we pushed these changes. 

Apologies for the DM haven't been around here for a few weeks and I'm not sure who else is watching the project these days. As always, I'm happy to discuss further if anything above raises an eyebrow or doesn't make sense, and your time and attention is greatly appreciated.

casperisfine:
I'm watching no worries. I didn't react because when I saw the notification the PR was a draft. But in general feel free to ping.

I'll take a closer look first thing tomorrow, but that looks reasonable.

app/jobs/shipit/reap_dead_tasks_job.rb
```diff
@@ -7,13 +7,33 @@ class ReapDeadTasksJob < BackgroundJob
...
+    def self.recently_created_at
+      MINUTES_TO_CONSIDER_TASK_RECENT.minutes.ago
+    end
+
+    private
+
+    def zombie_tasks
+      @zombie_tasks ||= Task
+                        .where(status: ZOMBIE_STATES)
+                        .where(
```
casperisfine, line 28: I think this logic should live in a method called `Task.zombie`.

app/jobs/shipit/perform_task_job.rb
```diff
@@ -62,6 +63,10 @@ def perform_task
 
     def checkout_repository
       unless @commands.fetched?(@task.until_commit).tap(&:run).success?
+        # acquire_git_cache_lock can take upto 15 seconds
+        # to process. Try to make sure that the job isn't
+        # marked dead while we attempt to acquire the lock.
+        @task.ping
         @task.acquire_git_cache_lock do
```
casperisfine, line 70: We should also probably ping again after we got the lock.

indiebrain: I don't have strong feelings on renewing the "alive" key after acquiring the lock. I had considered it, but decided against it because it seemed like it might be overkill. My reasoning: 

1. I've upped the lifetime of the "alive" key in redis to 30 seconds. 
2. Acquiring the lock times out after 15 seconds. I imagined that it would take far less than the remaining 15 seconds on the "alive" key to execute `@commands.fetched?(@task.until_commit).tap(&:run).success?` 
3. The very next thing we do is reset the "alive" key countdown as a side effect of `capture! @commands.fetch`. 

Do you think it'd be worth it to add here anyway?

casperisfine: Yeah, I'd just rather be safe than sorry.

app/jobs/shipit/reap_dead_tasks_job.rb
indiebrain: Addressed by: https://github.com/Shopify/shipit-engine/pull/1055/commits/45ac417579572572b9c6bb1fe85921871bca2921

casperisfine:
Rubocop is unhappy, also please squash the commits.

After that I'll merge.

app/jobs/shipit/perform_task_job.rb
indiebrain: Addressed by: https://github.com/Shopify/shipit-engine/pull/1055/commits/da1a05e148667774fbb292b2a6ea11a1ed4167f0#diff-8f7af914ee8af7005ec9a5585f1d848eR71

indiebrain:
Many thanks, @casperisfine !

PR #1057, opened 2020-04-15
Delay continuous delivery for stacks without a cached DeploySpec

ryanmrodriguez:
If a stack is provisioned with continuous delivery enabled, it's possible for the continuous delivery routine to run on the stack prior to the DeploySpec being cached. This occurs particularly when the git repo is large and slow to clone. This results in the _'max_commits'_ default value being used, and potentially the deployment of a very old commit being triggered. 

Here we have added a new pre-condition check to the _'trigger_continuous_delivery'_ method, along with some minor refactoring. We also added a nil check into _'test "#trigger_continuous_delivery bails out if the stack is deployable but was deployed too recently"'_ just to enforce the return type of  _'trigger_continuous_delivery'_, which was not tested before.

### DIFF
--- a/app/models/shipit/deploy_spec.rb
+++ b/app/models/shipit/deploy_spec.rb
@@ -40,7 +40,7 @@ module Shipit
       end
     end
 
-    def config_empty?
+    def blank?
       config.empty?
     end
 
--- a/app/models/shipit/deploy_spec.rb
+++ b/app/models/shipit/deploy_spec.rb
@@ -40,6 +40,10 @@ module Shipit
       end
     end
 
+    def config_empty?
+      config.empty?
+    end
+
     def supports_fetch_deployed_revision?
       fetch_deployed_revision_steps.present?
     end
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -161,7 +161,7 @@ module Shipit
     end
 
     def trigger_continuous_delivery
-      return if empty_config?
+      return if cached_deploy_spec.config_empty?
 
       commit = next_commit_to_deploy
 
@@ -607,10 +607,6 @@ module Shipit
 
     def ci_enabled_cache_key
       "stacks:#{id}:ci_enabled"
-    end
-
-    def empty_config?
-      cached_deploy_spec.config.empty?
     end
 
     def should_resume_continuous_delivery?(commit)
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -43,17 +43,6 @@ module Shipit
 
     scope :not_archived, -> { where(archived_since: nil) }
 
-    def env
-      {
-        'ENVIRONMENT' => environment,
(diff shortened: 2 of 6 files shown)


app/models/shipit/stack.rb
```diff
@@ -607,5 +606,19 @@ def emit_merge_status_hooks
     def ci_enabled_cache_key
       "stacks:#{id}:ci_enabled"
     end
+
+    def should_delay_continuous_delivery?(commit)
+      commit.deploy_failed? ||
+        (checks? && !EphemeralCommitChecks.new(commit).run.success?) ||
+        commit.recently_pushed? ||
+        cached_deploy_spec.config.empty?
+    end
```
benlangfeld, line 615: This should be a performance improvement:

```suggestion
    def should_delay_continuous_delivery?(commit)
      cached_deploy_spec.config.empty? ||
        commit.deploy_failed? ||
        (checks? && !EphemeralCommitChecks.new(commit).run.success?) ||
        commit.recently_pushed?
    end
```

ryanmrodriguez:
@casperisfine Could you please review? Thank you!

app/models/shipit/stack.rb
```diff
@@ -607,5 +608,22 @@ def emit_merge_status_hooks
     def ci_enabled_cache_key
       "stacks:#{id}:ci_enabled"
     end
+
+    def empty_config?
+      cached_deploy_spec.config.empty?
```
casperisfine, line 613: This seems weird. Should `cached_deploy_spec` be `nil` ?

In which circumstances would `.config` be empty?

benlangfeld: It’s `{}` until the file has been cached. I guess it’s also that if no file is found, but then it seems to make sense to wait also.

casperisfine: Ok my bad. Then it would make sense to expose a more meaningful method on `DeploySpec`.

Accessing its `config` attribute like this isn't really intended.

benlangfeld: Looks like Ryan has corrected this.

app/models/shipit/deploy_spec.rb
```diff
@@ -40,6 +40,10 @@ def config(*keys, &default)
       end
     end
 
+    def config_empty?
```
casperisfine, line 43: Sorry to be nitpicking, but it's not really what I expected.

I'd like something that communicate the state in term of business rules. Something that explains that this deploy spec is imcomplete, or missing, or something like that.

`blank?` could work as well, but it might have some unexpected consequences.

ryanmrodriguez: I understand what you mean. I guess the way I saw it, I didn't want to imply that the whole DeploySpec was empty/blank/etc. I am okay with 'blank?'.  @benlangfeld Do you have any input?

casperisfine: I'd say try `blank?`. CI should tell us if there's fallouts.

casperisfine:
Thanks.

PR #1059, opened 2020-04-20
Allows deployment's branding to be used in config

benlangfeld:
If the downstream application is branded `foobar`, then load `foobar.yml` to find a deployment spec.


casperisfine:
Hum. I'm not so certain about that one. I'm not against the feature itself, but I think it would be preferable to use another configuration field than `app_name`. Because this is a much bigger backward change.

benlangfeld:
What do you mean? This is backward-compatible.

casperisfine:
Nevermind. I missed that it falls back to `shipit.yml`.

I wasn't properly woke up. Sorrry -_-.

PR #1061, opened 2020-04-24
Reload tasks when checking for concurrent execution

DazWorrall:
At Shopify we're observing duplicate deploy tasks on stacks with continuous delivery enabled. Doing some debugging it seems we have 2 running instances of `ContinuousDeliveryJob`, and our theory is - similar to https://github.com/Shopify/shipit-engine/pull/1042#issuecomment-608118022 - the later task has a stale association, and so the `prevent_concurrency` fails. If that is the cause then reloading the association should fix the duplicate tasks. It's very difficult to replicate the conditions and know for sure though, and given this is a relatively benign change I'd like to ship and see.

### DIFF
--- a/test/models/deploys_test.rb
+++ b/test/models/deploys_test.rb
@@ -509,7 +509,7 @@ module Shipit
       deploy1.save
 
       deploy2 = create_test_deploy(stack_id: stack_id, user_id: user_id, since_commit_id: commit_ids[1], until_commit_id: commit_ids[2])
-      deploy2.type = "Shipit::Fake"
+      deploy2.type = "Shipit::Task"
       deploy2.save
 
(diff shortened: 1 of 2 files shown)


test/models/deploys_test.rb
```diff
@@ -509,7 +509,7 @@ def generate_commits(amount:, stack_id:, user_id:, validate:)
       deploy1.save
 
       deploy2 = create_test_deploy(stack_id: stack_id, user_id: user_id, since_commit_id: commit_ids[1], until_commit_id: commit_ids[2])
-      deploy2.type = "Shipit::Fake"
+      deploy2.type = "Shipit::Task"
```
DazWorrall, line 512: I change this because the reload caused [`ActiveRecord::SubclassNotFound`](https://github.com/Shopify/shipit-engine/runs/615261539#step:6:240).

PR #1064, opened 2020-04-27
Store PR head ref sha when creating commits

DazWorrall:
When creating `CommitDeployment`s we have to lookup the pull request head sha using the GitHub API:

https://github.com/Shopify/shipit-engine/blob/ae2ef38c72c6a88d952756b66da4746120ae5d10/app/models/shipit/deploy.rb#L250-L254

This means that, because this code is ran in an `after_create` callback, we are making these calls inside a database transaction, and this transaction is kept open for a long time. 

This PR is a first step to fixing this: we already store some pull request information on the commits table, and as we can infer the PR head sha from the commits API, we can store it when the record is created. I have 🎩ed this locally:

```
irb(main):003:0> Shipit::Commit.last.pull_request_head_sha
  Shipit::Commit Load (0.6ms)  SELECT `commits`.* FROM `commits` ORDER BY `commits`.`id` DESC LIMIT 1
=> "5b6a99494ddba625083e8ac79e4a88fd7100b928"
```

[This](https://github.com/dazworrall/junk/pull/25) was the test PR.

### DIFF
--- a/app/models/shipit/commit.rb
+++ b/app/models/shipit/commit.rb
@@ -92,7 +92,7 @@ module Shipit
       committer = User.find_or_create_committer_from_github_commit(commit)
       committer ||= Anonymous.new
 
-      new(
+      record = new(
         sha: commit.sha,
         message: commit.commit.message,
         author:  author,
@@ -102,6 +102,12 @@ module Shipit
         additions: commit.stats&.additions,
         deletions: commit.stats&.deletions,
       )
+
+      if record.pull_request?
+        record.pull_request_head_sha = commit.parents.last.sha
+      end
+
+      record
     end
 
     def message=(message)
--- a/db/migrate/20200427135152_add_pull_request_head_sha_to_commit.rb
+++ b/db/migrate/20200427135152_add_pull_request_head_sha_to_commit.rb
@@ -1,0 +1,5 @@
+class AddPullRequestHeadShaToCommit < ActiveRecord::Migration[6.0]
+  def change
+    add_column :commits, :pull_request_head_sha, :string, limit: 40
+  end
+end
--- a/test/dummy/db/schema.rb
+++ b/test/dummy/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_02_26_211925) do
+ActiveRecord::Schema.define(version: 2020_04_27_135152) do
 
   create_table "api_clients", force: :cascade do |t|
     t.text "permissions", limit: 65535
@@ -78,6 +78,7 @@ ActiveRecord::Schema.define(version: 2020_02_26_211925) do
     t.integer "pull_request_id"
     t.boolean "locked", default: false, null: false
     t.integer "lock_author_id", limit: 4
+    t.string "pull_request_head_sha", limit: 40
     t.index ["author_id"], name: "index_commits_on_author_id"
     t.index ["committer_id"], name: "index_commits_on_committer_id"
     t.index ["created_at"], name: "index_commits_on_created_at"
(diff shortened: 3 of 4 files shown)


casperisfine:
> because this code is ran in an `after_create` callback

Sorry not to have realized this sooner, but could it be delayed to `after_commit on: :create` ?

app/models/shipit/commit.rb
```diff
@@ -102,6 +102,12 @@ def self.from_github(commit)
         additions: commit.stats&.additions,
         deletions: commit.stats&.deletions,
       )
+
+      if record.pull_request?
+        record.pull_request_head_sha = commit.parents.last.sha
```
casperisfine, line 107: This doesn't work with GitHub's squashed merges right? I don't quite remember if we handle them at all today.

DazWorrall: Good question, I will check.

DazWorrall:
> Sorry not to have realized this sooner, but could it be delayed to after_commit on: :create ?

That would massively shrink the open transaction window yes, and was my [first thought](https://github.com/Shopify/shipit-engine/compare/defer-cd-creation), but I realised that to properly fix the race I think we need to pessimistically lock the stack when creating deploys, and although we would make the transaction fast the row lock would still be held for a long time.

If you think that making the transaction fast and minimising the possibility of the race is preferable to this approach then we could consider that.

DazWorrall:
> This is what you need the LHM for, right?

Yep!

app/models/shipit/commit.rb
DazWorrall: No, it does not work for squashed merged - but we already do not detect these as pull requests anyway, so we aren't e.g. creating `CommitDeployment`s for these PRs etc. We could fix this (it's not a big deal at Shopify - squashed merges are not preferred - but it may be elsewhere), but I would prefer to do that out of scope of the PR.

casperisfine:
> If you think that making the transaction fast and minimising the possibility of the race is preferable to this approach

I'd need to put more thoughts into it to be sure. I have to admit I'm busy on other things and been only superficially looking at this problem, so I'm kinda relying on your own understanding of it.

What I was trying to say is that if we need to lock to fix the the race condition for good, then we need to make sure the transaction is as fast as possible so that we don't create a bottleneck.

So I was not suggesting to make the transaction fast as a way to reduce the likeness of the race, but as a condition to be able to lock.

DazWorrall:
> So I was not suggesting to make the transaction fast as a way to reduce the likeness of the race, but as a condition to be able to lock.

Agreed with this. I don't think a fast transaction is sufficient on its own to fix the race: I think we _also_ need pessimistic locking of the stack (this would follow in a later PR) if there is a possibility that `ContinuousDeliveryJob`s can overlap, and while using `after_create_commit` would move the HTTP requests outside of the transaction, they would happen while the stack is locked, and I think that's too long a time for such a regularly queried table. To fix that we would either need to move this query async - something we previously undid because of a _different_ race condition - or store the PR sha ourselves.

There are other ways we could look at this - e.g. creating a `PullRequest` record rather than storing PR information on the commit. Right now that table is only used for the merge queue, so it would be a larger refactor.

`ContinuousDeliveryJob` has `BackgroundJob::Unique` on it, so we should _not_ have overlapping jobs in flight, but we are definitely seeing this in production. We could spend more time trying to fix this, but I think it's better if we dont have to rely on that mechanism entirely.

casperisfine:
> e.g. creating a `PullRequest` record rather than storing PR information on the commit. Right now that table is only used for the merge queue, so it would be a larger refactor.

Yeah, I had the same thought. It would be cleaner, but that's lots of work.

> ContinuousDeliveryJob has BackgroundJob::Unique on it, so we should not have overlapping jobs

It's been a while I suspect `BackgroundJob::Unique` isn't perfect. 

> We could spend more time trying to fix this, but I think it's better if we dont have to rely on that mechanism entirely.

Agreed, the improper locking is the root cause, we should fix it. Job unicity is only there to save on work we know is useless, not to actually prevent race conditions.

PR #1066, opened 2020-04-29
Move us closer to the Shopify style guide

DazWorrall:
I needed a change of pace so decided to do some style gardening :)

Our rubocop config has drifted far from the [style guide](https://github.com/Shopify/ruby-style-guide), this PR doesn't bring us all the way up to date but does knock out the quick wins. Our rubocop config is much smaller after this, and the `_todo` file is completely removed. There are still a few exclusions in the rubocop config I will work on in follow ups.

There only only a couple of substantive changes:

* Switch a use of `JSON.load` to `JSON.parse` ([here](https://github.com/Shopify/shipit-engine/commit/d5fa95d629ad999f7d88dfc123f62baf58273f5e#diff-7a20f6bc03c18cce1ab02564d23cf82dR8))
* Check for frozen strings in `OutputChunk` https://github.com/Shopify/shipit-engine/commit/6fce3c06eeb99ef70e0a005e608638dedf35b058

That last one is required because write out string literals at the end of tasks, and these are now frozen:

https://github.com/Shopify/shipit-engine/blob/7c5affca34928cdc266017acb6eea86bcacc4927/app/models/shipit/task.rb#L274-L277

Is there a cleaner fix?


DazWorrall:
Oh and just for fun: adopting the style guide without making any changes gives us `1163 offenses detected` 😁

casperisfine:
> Our rubocop config has drifted far from the [style guide](https://github.com/Shopify/ruby-style-guide)

Well, it actually predates it significantly.

app/models/shipit/pull_request.rb
```diff
@@ -129,8 +130,8 @@ def self.request_merge!(stack, number, user)
           stack: stack,
           number: number,
         )
-      rescue ActiveRecord::RecordNotUnique
-        retry
+                     rescue ActiveRecord::RecordNotUnique
+                       retry
```
casperisfine, line 134: It totally borked here.

app/models/shipit/pull_request.rb
```diff
@@ -111,11 +112,11 @@ def self.schedule_merges
     def self.extract_number(stack, number_or_url)
       case number_or_url
       when /\A#?(\d+)\z/
-        $1.to_i
+        Regexp.last_match(1).to_i
```
casperisfine, line 115: That cop is total 🐄 💩 . It's just as global as before, just way more verbose.

I know it's in the style guide, but let's disable it.

app/models/shipit/commit_deployment_status.rb
```diff
@@ -12,14 +13,14 @@ def create_on_github!
...
         create_status_on_github(author.github_api)
-      rescue Octokit::ClientError
-        raise if Shipit.github.api == author.github_api
-        # If the deploy author didn't gave us the permission to create the deployment we falback the the main shipit
-        # user.
-        #
-        # Octokit currently raise NotFound, but I'm convinced it should be Forbidden if the user can see the repository.
-        # So to be future proof I catch boths.
-        create_status_on_github(Shipit.github.api)
+                 rescue Octokit::ClientError
```
casperisfine, line 16: It borked the indentation here as well.

app/models/shipit/commit_deployment.rb
```diff
@@ -20,14 +21,14 @@ def create_deployment_on_github!
...
         create_deployment_on_github(author.github_api)
-      rescue Octokit::ClientError
-        raise if Shipit.github.api == author.github_api
-        # If the deploy author didn't gave us the permission to create the deployment we falback the the main shipit
-        # user.
-        #
-        # Octokit currently raise NotFound, but I'm convinced it should be Forbidden if the user can see the repository.
-        # So to be future proof I catch boths.
-        create_deployment_on_github(Shipit.github.api)
+                 rescue Octokit::ClientError
```
casperisfine, line 24: And here.

app/controllers/shipit/merge_status_controller.rb
```diff
@@ -113,9 +114,9 @@ class ReferrerParser
 
       def initialize(referrer)
         if URL_PATTERN =~ referrer.to_s
-          @repo_owner = $1.downcase
-          @repo_name = $2.downcase
-          @pull_request_number = $3.to_i
+          @repo_owner = Regexp.last_match(1).downcase
+          @repo_name = Regexp.last_match(2).downcase
+          @pull_request_number = Regexp.last_match(3).to_i
```
wvanbergen, line 119: If you don't want the global, `if (match_info = URL_PATTERN.match(referrer.to_s)` could do the trick

app/models/shipit/commit_deployment.rb
```diff
@@ -20,14 +21,14 @@ def create_deployment_on_github!
...
+                   raise if Shipit.github.api == author.github_api
+                   # If the deploy author didn't gave us the permission to create the deployment we falback the the main shipit
+                   # user.
+                   #
+                   # Octokit currently raise NotFound, but I'm convinced it should be Forbidden if the user can see the repository.
+                   # So to be future proof I catch boths.
+                   create_deployment_on_github(Shipit.github.api)
```
wvanbergen, line 31: This indentation is off. The `rescue` alignment rule is broken: https://github.com/rubocop-hq/rubocop/pull/7531

lib/shipit/same_site_cookie_middleware.rb
```diff
@@ -13,12 +14,12 @@ def call(env)
...
-                      .map do |cookie|
-            cookie << '; Secure' if cookie !~ /;\s*secure/i
-            cookie << '; SameSite=None' unless cookie.match?(/;\s*samesite=/i)
-            cookie
+          .split(COOKIE_SEPARATOR)
+          .compact
+          .map do |cookie|
+          cookie << '; Secure' if cookie !~ /;\s*secure/i
+          cookie << '; SameSite=None' unless cookie.match?(/;\s*samesite=/i)
+          cookie
```
wvanbergen, line 22: Maybe 

``` ruby
set_cookies = headers['Set-Cookie'].split(COOKIE_SEPARATOR).compact
set_cookies.map! do ... end
```

wvanbergen:
> Is there a cleaner fix?

You could also mark the literal that gets fed to this method as non-frozen: `+"literal"`

DazWorrall:
> You could also mark the literal that gets fed to this method as non-frozen: +"literal"

This was a good suggestion, but I decided against it because it turns out the test suite is also peppered with `write` calls.

DazWorrall:
As for the `PerlBackrefs` cop I split the difference: added a `match_info = foo.match` pattern where that was an option, and where not reverted to `$n` usage and whitelisted it.

PR #1067, opened 2020-04-30
Truncate commit msg

JackTLi:


DazWorrall:
I think this is safe to close now.

PR #1068, opened 2020-04-30
Normalise Stack#cached_deploy_spec in test env 

DazWorrall:
I've broken this out from something else I'm working on:

```
(byebug) @stack.changed?
false
(byebug) @stack.cached_deploy_spec.blank?
false
(byebug) @stack.changed?
true
```

Right now every time we touch `cached_deploy_spec` in the tests the model instance gets marked as dirty - this is because  the serialised spec does not match exactly the literals we have in the fixture. This fixes that by:

* Supporting pretty generation on the serialiser
* Enabling it in the test env
* Normalising the fixture

### DIFF
--- a/test/models/stacks_test.rb
+++ b/test/models/stacks_test.rb
@@ -390,7 +390,7 @@ module Shipit
 
     test "updating the stack emit a hook" do
       expect_hook(:stack, @stack, action: :updated, stack: @stack) do
-        @stack.update(repo_name: 'foo')
+        @stack.update(environment: 'foo')
       end
     end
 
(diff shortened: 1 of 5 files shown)


test/models/stacks_test.rb
```diff
@@ -390,7 +390,7 @@ def self.deliver(event, stack, payload)
 
     test "updating the stack emit a hook" do
       expect_hook(:stack, @stack, action: :updated, stack: @stack) do
-        @stack.update(repo_name: 'foo')
+        @stack.update(environment: 'foo')
```
DazWorrall, line 393: This test has been incorrect since the `repo_*` attributes were moved to the repository model - the stack instance was marked as dirty because of this issue, `repo_name` doesn't exist anymore!

PR #1069, opened 2020-05-01
Lock stack when creating a deploy

DazWorrall:
~Requires #1068~

This I think is the final piece to properly close the multiple-auto-deploys problem - pessimistic locking to ensure that only one caller at a time can try to create a deploy. Since #1064 and #1065 the window for the race to happen is now very small, but something like this is required to properly close it.

Of course Stacks are loaded from the DB very regularly, so we need to be careful with locks. However as we only create deploys very infrequently, I don't expect this to cause a problem.

### DIFF
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -142,8 +142,11 @@ module Shipit
 
     def trigger_deploy(*args, **kwargs)
       run_now = kwargs.delete(:run_now)
-      deploy = build_deploy(*args, **kwargs)
-      deploy.save!
+      deploy = with_lock do
+        deploy = build_deploy(*args, **kwargs)
+        deploy.save!
+        deploy
+      end
       run_now ? deploy.run_now! : deploy.enqueue
       continuous_delivery_resumed!
       deploy


app/jobs/shipit/continuous_delivery_job.rb
```diff
@@ -7,6 +7,9 @@ class ContinuousDeliveryJob < BackgroundJob
     on_duplicate :drop
 
     def perform(stack)
+      # We reload to ensure that, if the record is locked to create a deploy, the following short ciruits work correctly
+      stack = stack.reload
```
wvanbergen, line 11: I am not sure exactly, but assuming that `stack` is a object from a GlobalID, it should be fresh out of the database at this point, right? Why is the reload needed? Local caching?

DazWorrall: It's [this](https://github.com/Shopify/shipit-engine/pull/1042#issuecomment-608118022) issue (sorry, should have linked). tl;dr the hydrated model instance here could be stale, because of the work of `BackgroundJob::Unique`.

DazWorrall: Actually, now I think about it, because we don't specify a timeout on the job I don't _think_ that applies.

https://github.com/Shopify/shipit-engine/blob/da691856a1033c2b7cd107cd640c50d7ffdc7ae0/app/jobs/shipit/background_job/unique.rb#L16-L26

Without a timeout `Redis::Lock::LockTimeout` is raised immediately I think 🤔

DazWorrall: [Yes it is](https://github.com/nateware/redis-objects/blob/1b1d289fdc1dd4585f52fd8f3e6fcf6f1026f40a/lib/redis/lock.rb#L76-L87), so I'll take the reload out.

PR #1071, opened 2020-05-01
Truncate commit messages on creation.

ajshepley:
I figured we should use the same logic for both the assignment truncation and the instantiation truncation.

Didn't know if we want `...` vs. slice truncation, but the slice is what we were doing before so I went with reusing the old logic. It also checks the field size first.

### DIFF
--- a/app/models/shipit/commit.rb
+++ b/app/models/shipit/commit.rb
@@ -114,7 +114,7 @@ module Shipit
     def message=(message)
       limit = self.class.columns_hash['message'].limit
       if limit && message && message.size > limit
-        message = message.slice(0, limit)
+        message = message.truncate_bytes(limit)
       end
       super(message)
     end
--- a/test/models/commits_test.rb
+++ b/test/models/commits_test.rb
@@ -69,6 +69,39 @@ module Shipit
       refute_predicate commit, :revert?
     end
 
+    test '.create_from_github truncates long messages' do
+      message = 'ABCDEFGHIJ' * 7000
+
+      assert_difference -> { Commit.count }, +1 do
+        @stack.commits.create_from_github!(
+          resource(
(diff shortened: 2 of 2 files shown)


app/models/shipit/commit.rb
```diff
@@ -92,10 +92,11 @@ def self.from_github(commit)
       author ||= Anonymous.new
       committer = User.find_or_create_committer_from_github_commit(commit)
       committer ||= Anonymous.new
+      truncated_message = truncate_message(commit.commit.message)
 
       record = new(
         sha: commit.sha,
-        message: commit.commit.message,
+        message: truncated_message,
```
casperisfine, line 99: `ActiveRecord::Base.new` calls `message=`, so I think you can do the truncate in `def message=(message)` only.

casperisfine:
I don't understand the need for that change.

app/models/shipit/commit.rb
ajshepley: We observed that this was apparently not happening in production, such that a large commit message was causing this to blow up.
https://app.bugsnag.com/shopify/shipit/errors/5eaafd2df8753600184d823b?filters[event.since][0]=30d&filters[error.status][0]=open

That said, I will test a bit more to confirm locally. Perhaps we are mistaken about what occurred?

casperisfine: Yeah, that's not supposed to happen. If you don't mind I'll try to dig that issue.

casperisfine: So I've but some breakpoint and I'm 100% certain the message was truncated as expected.

I'm unsure what happened in production, but I can tell with certainty that this isn't the fix you are looking for.

If I had to guess, I would say that the bug is because MySQL size is a bytesize, whereas `String#truncate` works on a character count basis.

PR #1072, opened 2020-05-13
Work Continuous Delivery jobs as deploys

benlangfeld:
These jobs may run checks similarly to `PerformCommitChecksJob`, which may require any access to special resources which are granted to deployment jobs; for example, our case involves making a Docker daemon available to deployment jobs so deployment routines can be executed in containers, and we don't provision this or the resources necessary for it on the nodes which work other queues.


casperisfine:
Thanks.

PR #1073, opened 2020-05-20
Do not use response if no data was returned from block

JackTLi:
Internal shopify issue: https://github.com/Shopify/shipit/issues/1137

Inside of `GithubSyncJob`, I've noticed that it's possible for non-existant repos to end up with commits from other repos showing up on the Stack.

Upon investigation, it seems that it's possible for the `last_response` here to grab the response from a previous job on the sidekiq thread, since it seems the `last_response` variable is not being set in the case that a repo doesn't exist.

### Evidence

Exception handling on Shipit side:
https://github.com/Shopify/shipit-engine/blob/master/app/models/shipit/stack.rb#L380

Octokit raises the exception before the response makes it back to setting the `last_response` variable:
https://github.com/octokit/octokit.rb/blob/master/lib/octokit/response/raise_error.rb#L15

### The fix

Filter out the response from the iterator if no data was returned from the block.

### DIFF
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -377,6 +377,8 @@ module Shipit
       handle_github_redirections do
         Shipit.github.api.commits(github_repo_name, sha: branch)
       end
+    rescue Octokit::Conflict
+      [] # Repository is empty...
     end
 
     def handle_github_redirections
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -377,8 +377,6 @@ module Shipit
       handle_github_redirections do
         Shipit.github.api.commits(github_repo_name, sha: branch)
       end
-    rescue Octokit::Conflict
-      [] # Repository is empty...
     end
 
     def handle_github_redirections
--- a/lib/shipit/octokit_iterator.rb
+++ b/lib/shipit/octokit_iterator.rb
@@ -4,11 +4,10 @@ module Shipit
     include Enumerable
 
     def initialize(relation = nil)
-      if relation
-        @response = relation.get(per_page: 100)
+      @response = if relation
+        relation.get(per_page: 100)
       else
-        yield Shipit.github.api
-        @response = Shipit.github.api.last_response
+        yield
       end
     end
 
--- a/lib/shipit/octokit_iterator.rb
+++ b/lib/shipit/octokit_iterator.rb
@@ -7,16 +7,13 @@ module Shipit
       if relation
         @response = relation.get(per_page: 100)
       else
-        yield Shipit.github.api
-        @response = Shipit.github.api.last_response
+        data = yield Shipit.github.api
+        @response = Shipit.github.api.last_response if data.present?
       end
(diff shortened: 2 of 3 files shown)


casperisfine:
Unfortunately that doesn't work. Maybe Octokit methods returns an Array, and you can't get the actual response from it:

```
>> Shipit.github.api.commits('rails/rails', sha: 'master').class
HTTP Cache: [GET /repos/rails/rails/commits?sha=master] miss, store
=> Array
```

That's what `last_response` is for. 

See also how `request` is implemented in octokit:

```ruby
    def request(method, path, data, options = {})
      if data.is_a?(Hash)
        options[:query]   = data.delete(:query) || {}
        options[:headers] = data.delete(:headers) || {}
        if accept = data.delete(:accept)
          options[:headers][:accept] = accept
        end
      end

      @last_response = response = agent.call(method, Addressable::URI.parse(path.to_s).normalize.to_s, data, options)
      response.data
    end
```


So I can see two ways out of this:

  - Stop using these methods, and go one layer deeper using the agent directly. But that's brittle.
  - Figure out why the `last_response` are leaking. Because [we're supposed to have one client per thread, so it's not supposed to happen](https://github.com/Shopify/shipit-engine/blob/829918721709288100f42cfb4bc4ea0699c7e790/lib/shipit/github_app.rb#L62-L68).

JackTLi:
@casperisfine hm good catch on that, I missed the methods only returning the response body.

The problem here should be that even though we are thread safe, a sidekiq worker on a thread can leak state between jobs since it’s using the same client instance. In the 409 case, the octokit library doesn’t do a write to last_response, so we end up using the last_response from the previous job on the thread. 

Maybe we could just use a new client on every job, or I’ll go through the shipit code again and see if we could function with the response return without the whole response

casperisfine:
> In the 409 case, the octokit library doesn’t do a write to last_response, so we end up using the last_response from the previous job on the thread.

That sounds fishy. Why do we even access the `last_response` if we got a 409? We shouldn't. Does something swallow that exception?

casperisfine:
Also I agree that this interface is crap, but I'm not aware of any alternative :/

JackTLi:
Yup, we swallow here - https://github.com/Shopify/shipit-engine/blob/master/app/models/shipit/stack.rb#L380

It seems that this method is implemented assuming we use the return value here, instead of indirectly looking at `last_response`. I changed the yield to try to reflect this, but unfortunately not a clean fit 😢

casperisfine:
So, as a quick fix we can check what `yield` returns, and if it's an empty array, we don't take the last response. It's fairly ad hoc but that should work.

In parallel I think Octokit should at the very least reset `last_response` of errors.

JackTLi:
@casperisfine that's a good idea, I added a check

JackTLi:
FYI also merged a fix upstream https://github.com/octokit/octokit.rb/pull/1260

casperisfine:
> FYI also merged a fix upstream

👏 💯

PR #1077, opened 2020-06-10
Add dedicated view for tasks

DazWorrall:
We have new and different tooling providing views into the state of deploys and rollbacks, but tasks are currently out of scope for those new things. Shipit may be used more as a generic task executor while we are experimenting and transitioning our delpoy tooling, so this adds a dedicated task-only scoped view which could be used to help with that. Instead of commits it only lists available tasks, and the job history is also only tasks.

Screen Recording 2020-06-10 at 14 55 40

Someone with some _actual_ frontend skills should make the new task area nicer I think!

### DIFF
--- a/app/controllers/shipit/stacks_controller.rb
+++ b/app/controllers/shipit/stacks_controller.rb
@@ -78,6 +78,11 @@ module Shipit
     def settings
     end
 
+    def tasks
+      @stack = Stack.from_param!(params[:id])
+      @tasks = @stack.tasks.where(type: nil).order(id: :desc).preload(:since_commit, :until_commit, :user).limit(10)
+    end
+
     def statistics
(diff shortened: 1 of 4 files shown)


gigr:
@DazWorrall I can see how this buys us a way to get rid of many other views in Shipit. I don't think the UX needs much improvement either—you did a good job. I would get rid of the Tasks drop-down and always link to the list/history page (it's two clicks to kick off a task either way).

DazWorrall:
>  I would get rid of the Tasks drop-down and always link to the list/history page

I considered this but decided against it - I figured it was best to keep the current task workflows identical for our existing users (at Shopify and beyond). If people feel strongly I could be convinced though.

DazWorrall:
This is useful in its current state so I'd like to ship it, if we want to make it prettier we can do that in a follow up.

PR #1080, opened 2020-06-11
Add rollback api

JackTLi:
Add api to be able to trigger rollbacks.

Very similar to deployments api, with a few slight modifications. Tests copied over from there as well.

### DIFF
--- a/app/controllers/shipit/api/rollbacks_controller.rb
+++ b/app/controllers/shipit/api/rollbacks_controller.rb
@@ -1,0 +1,24 @@
+
+module Shipit
+  module Api
+    class RollbacksController < BaseController
+      require_permission :deploy, :stack
+
+      params do
+        requires :sha, String, length: { in: 6..40 }
+        accepts :force, Boolean, default: false
+        accepts :env, Hash, default: {}
+      end
+      def create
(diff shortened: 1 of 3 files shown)


app/controllers/shipit/api/rollbacks_controller.rb
```diff
@@ -0,0 +1,24 @@
+
+module Shipit
+  module Api
+    class RollbacksController < BaseController
+      require_permission :deploy, :stack
+
+      params do
+        requires :sha, String, length: { in: 6..40 }
+        accepts :force, Boolean, default: false
```
gigr, line 9: How do you intend to use this on the shipit-next side? I think we could always set to true and always rollback, regardless of whether or not the stack is locked.

app/controllers/shipit/api/rollbacks_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+        accepts :env, Hash, default: {}
+      end
+      def create
+        commit = stack.commits.by_sha(params.sha) || param_error!(:sha, 'Unknown revision')
```
ajshepley, line 13: It'd be a good idea to surface any errors from this API on ShipitNext - or perhaps, to the user's slack - when the button is pressed.

PR #1086, opened 2020-06-15
Abort active deploy on rollback with force

JackTLi:
If active deploy on the stack, abort it with the `force` flag and perform the rollback. 
This is the equivalent of pressing "Abort and rollback" from the deploy logs

### DIFF
--- a/app/controllers/shipit/api/rollbacks_controller.rb
+++ b/app/controllers/shipit/api/rollbacks_controller.rb
@@ -15,6 +15,12 @@ module Shipit
         deploy = stack.deploys.find_by(until_commit: commit) || param_error!(:sha, 'Cant find associated deploy')
         deploy_env = stack.filter_deploy_envs(params.env)
 
+        if !params.force && stack.active_task?
+          param_error!(:force, "Can't rollback, deploy in progress")
+        elsif stack.active_task?
+          stack.active_task.abort!(aborted_by: current_user)
+        end
+
         rollback = deploy.trigger_rollback(current_user, env: deploy_env, force: params.force)
 
         render_resource(rollback, status: :accepted)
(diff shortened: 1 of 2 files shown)


app/controllers/shipit/api/rollbacks_controller.rb
```diff
@@ -15,6 +15,12 @@ def create
         deploy = stack.deploys.find_by(until_commit: commit) || param_error!(:sha, 'Cant find associated deploy')
         deploy_env = stack.filter_deploy_envs(params.env)
 
+        if !params.force && stack.active_task?
+          param_error!(:force, "Can't rollback, deploy in progress")
+        elsif stack.active_task?
+          stack.active_task.abort!(aborted_by: current_user)
+        end
```
DazWorrall, line 22: This could be racey.

`abort!` is not a blocking operation - it sets the task state to `aborting` in the database and requests an abort:

https://github.com/Shopify/shipit-engine/blob/d72f535702c7f974def29f416924994b63bb4724/app/models/shipit/task.rb#L295-L300

Asynchronously PerformTaskJob is checking if an abort has been requested, and starts trying to kill the task:

https://github.com/Shopify/shipit-engine/blob/829918721709288100f42cfb4bc4ea0699c7e790/app/jobs/shipit/perform_task_job.rb#L50-L58

Note that not only is this being done async, it is also not immediate - a few `TERM`s are sent before a final `KILL`.

Your code carries on and starts a new task rolling back, but for a time the existing job will still run, and both could be running together. It's not impossible to see how the original task could collide with or undo the rollback.

This is why the `rollback_once_aborted` exists on the deploy, to ensure the rollback only starts once the task is flight has definitely stopped. You can use that here if it's sufficient, but I can see how you might want to extend it with a `rollback_to_revision` or similar.

PR #1087, opened 2020-06-15
Perform rollback after abort

JackTLi:
@DazWorrall Mentioned that the current strategy won't work since it takes time for the abort to take place. 

This change will use mechanism for triggering revert after abort to trigger a rollback to
a specific deploy, by saving the targeted deploy.

### DIFF
--- a/app/models/shipit/task.rb
+++ b/app/models/shipit/task.rb
@@ -23,6 +23,8 @@ module Shipit
     belongs_to :until_commit, class_name: 'Commit', required: false
     belongs_to :since_commit, class_name: 'Commit', required: false
 
+    belongs_to :rollback_once_aborted_to, class_name: 'Task', optional: true
+
     deferred_touch stack: :updated_at
 
     has_many :chunks, -> { order(:id) }, class_name: 'OutputChunk', dependent: :delete_all, inverse_of: :task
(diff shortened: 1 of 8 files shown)


app/models/shipit/task.rb
```diff
@@ -23,6 +23,8 @@ class Task < ActiveRecord::Base
     belongs_to :until_commit, class_name: 'Commit', required: false
     belongs_to :since_commit, class_name: 'Commit', required: false
 
+    belongs_to :rollback_once_aborted_to, class_name: 'Task', optional: true
```
ajshepley, line 26: `rollback_once_aborted_to` reads a bit weird but I can't think of anything better.

DazWorrall:
@casperisfine do you have thoughts on this? For context this is a follow up to https://github.com/Shopify/shipit-engine/pull/1086.

casperisfine:
Seems fine. However there's two way to go about such API.

This PR makes it so that when you ask for a rollback you first abort the existing running tasks.

Or you can go the other way (like the UI controllers do) and have an option to rollback when you abort the active task.

Both can be justified based on what you want to do with the API, if this is the most convenient API for you I think ti make sense, and doesn't really prevent to implement the alternative interface later on if needed.

app/models/shipit/task.rb
JackTLi: yeah it does, but at least its kinda descriptive 😛

PR #1088, opened 2020-06-18
Banners: tweak copy to be less stressful

kaelig:
At the moment, banners have titles ending with "!", which could induce unnecessary stress. Let's be chill :)

### DIFF
--- a/app/views/shipit/stacks/_banners.html.erb
+++ b/app/views/shipit/stacks/_banners.html.erb
@@ -2,10 +2,10 @@
   <div class="banner">
     <div class="banner__inner wrapper">
       <div class="banner__content">
-        <h2 class="banner__title">Heads Up!</h2>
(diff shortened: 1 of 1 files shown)


app/views/shipit/stacks/_banners.html.erb
```diff
@@ -53,9 +53,9 @@
   <div class="banner">
     <div class="banner__inner wrapper">
       <div class="banner__content">
-        <h2 class="banner__title">Continuous Delivery Delayed!</h2>
+        <h2 class="banner__title">Continuous delivery delayed</h2>
```
kaelig, line 56: We use sentence case in headings in the product, this should probably be the case here as well (see Polaris content guidelines).

kaelig:
Closing this as a new shipit is coming.

benlangfeld:
@kaelig Do you have any details on that?

PR #1090, opened 2020-06-29
Adapt merge status to new GitHub UI

robinbrandt:
Github changed the width of their page which is why the "Add to merge queue" button looks out of place now. I removed the max-width constraint as this gave me the right results when playing around in the chrome web inspector.

Before:
Screenshot 2020-06-29 08 50 25

After:

Screenshot 2020-06-29 08 51 26

Would be great if somebody could 🎩 this.

### DIFF
--- a/app/assets/stylesheets/merge_status.scss
+++ b/app/assets/stylesheets/merge_status.scss
@@ -1,3 +1,0 @@
-.merge-status-container {
-  max-width: 700px;
-}


robinbrandt:
@casperisfine is it one or two reviews in this repo?

casperisfine:
No particular rules. Feel free to merge.

PR #1094, opened 2020-08-04
Stack api filter

rdji123:
Make api calls to filter stacks with given repo name and owner if given as param

### DIFF
--- a/app/controllers/shipit/api/stacks_controller.rb
+++ b/app/controllers/shipit/api/stacks_controller.rb
@@ -10,7 +10,7 @@ module Shipit
         accepts :repo_name, String
       end
       def index
-        @stacks = Stack.all
+        @stacks = stacks
         if params[:repo_owner] && params[:repo_name]
           full_repo_name = [repo_owner, repo_name].join('/')
           @stacks = if (repository = Repository.from_github_repo_name(full_repo_name))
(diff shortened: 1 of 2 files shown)


app/controllers/shipit/api/stacks_controller.rb
```diff
@@ -5,8 +5,21 @@ class StacksController < BaseController
...
+        accepts :repo_owner, String
+        accepts :repo_name, String
+      end
       def index
-        render_resources(stacks)
+        @stacks = stacks
+        if params[:repo_owner] && params[:repo_name]
+          full_repo_name = [repo_owner, repo_name].join('/')
+          @stacks = if (repository = Repository.from_github_repo_name(full_repo_name))
+            repository.stacks
```
gigr, line 17: I think this should be `stacks.where(repository: Repository.from_github_repo_name(full_repo_name))`.

JackTLi: There's a relation here so it should be issuing doing that under the hood: https://github.com/Shopify/shipit-engine/blob/master/app/models/shipit/repository.rb#L16

unless im missing something?

gigr: It's true, but `stacks` is already defined in the `BaseController`. Since there is already filtering logic there, we should rely on it instead of re-defining a way to lookup here. @rdji123 and I paired on this and fixed the underlying issue (plus added a test).

https://github.com/Shopify/shipit-engine/blob/1e7c4be36d5d986f56fc2ba0fd8fd4e12d3f336e/app/controllers/shipit/api/base_controller.rb#L60-L62

Issue #1093, opened 2020-07-29
[Proposal] Restructuring PullRequest 


ryanmrodriguez:
As an ongoing effort on our part to deliver Review Stacks into Shipit, we have a couple of decisions to make pertaining the structure of our changes. We would like to know your thoughts on the subject matter. 

*What are we trying to do?*
[Deliver a Review Stack feature into Shipit.](https://github.com/Shopify/shipit-engine/issues/960) 

*How are we implementing this?*
We extend the Stack model, and rely on the creation and synchronization of Pull Request objects. This is kicked off with custom handlers.

*What issues are we facing?*
In order to tighten up the lifecycle of Review Stacks, we believe they should be owned by Pull Requests. However, within the current codebase, we feel that a PullRequest doesn't accurately reflect what we are meant to be modeling (a traditional Github Pull Request). Currently the PullRequest class is really only used for a merge queue. So that leads to a few decisions that need to be made before we proceed. 

1) We believe that PullRequest should be split into "MergeRequest" and "PullRequest" to more accurately reflect what each is meant to be. MergeRequest would fill the role of the current PullRequest, and the new PullRequest class would take on modeling a traditional GitHub pull request in the same lightweight fashion as things like Commit, and would be used in our Review Stacks feature. How would you feel about this change? Does it make sense to you? Any other insights?

2) We have read about Merge Queue v2, and we are wondering how this would affect the merge queue v1 feature in Shipit. Is this something you guys can see being completely removed from Shipit in favor of using merge queue v2 (assuming it would be open-sourced)?

casperisfine:
> MergeRequest would fill the role of the current PullRequest, and the new PullRequest class would take on modeling a traditional GitHub pull request in the same lightweight fashion as things like Commit, and would be used in our Review Stacks feature. How would you feel about this change? Does it make sense to you? Any other insights?

Yeah I'm fine with that. That's actually a design decision I've come to regret a while ago. So I'm very much ok for at least a rename of `PullRequest` into `MergeRequest`. Further refactoring would be on a case by case basis. And I'd be fine with actual PullRequest tracking in the Commit fashion, as long as it's optional.

> We have read about Merge Queue v2, and we are wondering how this would affect the merge queue v1 feature in Shipit

What's been publicly presented as "Merge Queue v2" is an entirely distinct project from `shipit-engine`. It might be open sourced at some point, but if it is it won't be part of `shipit-engine`. So I'd say don't be concerned with that.

[...]

PR #1095, opened 2020-08-05
Rename Shipit::PullRequest to Shipit::MergeRequest

indiebrain:
Implements the proposal of Shopify/shipit-engine issue #1093 by
renaming the concept of the `Shipit::PullRequest` to
`Shipit::MergeRequest`. Consideration was given to maintaining the
name "PullRequest" where shipit-engine is interacting directly with
the Github API - for example, in the `Shipit.github.api client`. In
this way the `Shipit::MergeRequest` is related to Github Pull
Requests, but not necessarily _of_ a Github Pull Request.

The goal is to make way for a more direct modeling of a Github Pull
Request in shipit-engine - `Shipit::PullRequest` - which like
`Shipit::Commit` models the Github concept. The plan is to use this
closer modeling of the Github Pull Request to finish preparations for
submission of our "Review Stack" feature to `Shopify/shipit-engine`.

Notes
-------

This does include non-backward compatible HTTP API changes to 
shipit-engine - the `pull_requests` collection resources have been 
renamed to `merge_requests`. 

References
----------

- https://github.com/Shopify/shipit-engine/issues/1093


app/models/shipit/merge_request.rb
```diff
@@ -112,17 +112,17 @@ def self.schedule_merges
     def self.extract_number(stack, number_or_url)
       case number_or_url
       when /\A#?(\d+)\z/
-        $1.to_i
+        Regexp.last_match(1).to_i
```
casperisfine, line 115: I just disabled that silly cop on master. Please revert that change.

indiebrain: [There was a another definition of that cop later on in the `.rubocop.yml`](https://github.com/Shopify/shipit-engine/pull/1095/commits/5dd9f8a1d03f1d2ec532c83ed3383dd1fa451789#diff-11a0d7906801d9dea0eccb85667ad811) which overrode the changes in https://github.com/Shopify/shipit-engine/commit/e0acb4e68805d04b119001a38c765c4c3358461b. I've removed the later cop definition so the `Style/PerlBackrefs` cop is totally disabled for the project as part of this body of work. Hope that doesn't muddy the waters too much.

casperisfine: Thanks.

app/jobs/shipit/merge_merge_requests_job.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+module Shipit
+  class MergeMergeRequestsJob < BackgroundJob
```
casperisfine, line 3: I just noticed this. `MergeMerge` is a bit weird. I'm not sure what would be better though. Maybe `ProcessMergeRequestsJob`?

app/models/shipit/commit.rb
```diff
@@ -219,7 +219,7 @@ def detach_children!
       children.detach!
     end
 
-    def pull_request?
+    def merge_request?
```
casperisfine, line 222: I think this shouldn't be changed. The meaning of this method is to tell wether the commit looks like a Pull Request merge commit.

app/models/shipit/commit.rb
```diff
@@ -289,20 +289,20 @@ def deploy_failed?
 
     def identify_pull_request
```
casperisfine, line 290: However this should be renamed in `identify_merge_request `

app/jobs/shipit/merge_merge_requests_job.rb
indiebrain: On a scale of 1-5, I'm a 5 (Strongly Agree). 😄

app/models/shipit/commit.rb
indiebrain: Do you think the name `Commit#pull_request_merge?` would ever-so-slightly elevate that intent to the reader over `Commit#pull_request?`? Would `Commit#pull_request_merge?` be an acceptable name. Do you think it improves clarity at all for the reader? Or would you prefer we stick with `Commit#pull_request?` predicate?

casperisfine: I think it's slightly too verbose. I'd rather stick to `pull_request?` for now, we can always rename it later on.

app/jobs/shipit/merge_merge_requests_job.rb
indiebrain: Resolved in: https://github.com/Shopify/shipit-engine/pull/1095/commits/ce8120a5f1e10b478c6b0facf5893a2027974d60

app/models/shipit/commit.rb
indiebrain: Resolved in: https://github.com/Shopify/shipit-engine/pull/1095/commits/ce8120a5f1e10b478c6b0facf5893a2027974d60

app/models/shipit/commit.rb
indiebrain: Resolved in: https://github.com/Shopify/shipit-engine/pull/1095/commits/ce8120a5f1e10b478c6b0facf5893a2027974d60

benlangfeld:
@casperisfine Looks like this should be ready to go :) Thank you for your very agile reviews, we appreciate it.

casperisfine:
@benlangfeld more than welcome o7

PR #1096, opened 2020-08-06
Don't redirect on enabling emergency mode

JackTLi:


JackTLi:
Actually none of the other pages redirect since the redirect is only set up for the main page, the emergency mode is just a bit of an outlier since it does a self-redirect 🙂

ajshepley:
I get redirected if I go to, say, settings and then `Commits & Deploys`.

PR #1097, opened 2020-08-06
Awaiting provision stack should not be deployable

viniciusgama:
We need to prevent deployments from happening for review stacks that are still in the queue waiting for resources to be provisioned.


viniciusgama:
created by mistake

Issue #1076, opened 2020-06-05
[Proposal] Collect deployment 'Failure Reasons'


ryanmrodriguez:
We would like to have the ability to classify and submit 'reasons' for failed deployments. Our goal ultimately is to have greater visibility on these failures for our own metrics. 

Before we build this out, we would like your feedback on the initial concept. The basic idea would be to add some extra inputs on the deployment page when that deployment has failed. 

Questions:
- Would this be something that you would consider accepting into Shipit?
- What would be the preferred way to handle additional datapoints?

I have attached some screenshots of the initial concept. Thanks!

<img width="1537" alt="base" src="https://user-images.githubusercontent.com/62253265/83906276-e58d6200-a728-11ea-95ac-c39d40ef9667.png">
<img width="1539" alt="dropdown" src="https://user-images.githubusercontent.com/62253265/83906280-e8885280-a728-11ea-8729-66ebf140db66.png">
<img width="1538" alt="missing_dep" src="https://user-images.githubusercontent.com/62253265/83906286-ec1bd980-a728-11ea-92f7-99f04f9d1856.png">
<img width="1543" alt="other_blank" src="https://user-images.githubusercontent.com/62253265/83906299-f0e08d80-a728-11ea-98cc-6179c3073ade.png">
<img width="1539" alt="other_typed" src="https://user-images.githubusercontent.com/62253265/83906308-f3db7e00-a728-11ea-95bc-4ed82ffe4004.png">

casperisfine:
> Would this be something that you would consider accepting into Shipit?

Not sure. I feel like this classification would be fairly dependent on each installation and the type of software you often deploy. Or we'd need to make the categories entirely dynamic, which means even more complexity.

But I guess I could be convinced otherwise by a use case that would suits a large proportion of Shipit users.

However I'd be interested in making this cleanly implementable. There is already a "plugin" system that allows to load arbitrary JS & CSS in the task output page, and it's fairly easy to add custom controllers to the host app. So it wouldn't be too hard to implement this already, I think the only thing missing would be to add a setting to have a list of plugin enabled globally.

> What would be the preferred way to handle additional datapoints?

In your host app you can do what you want. Either add a column to `tasks` (simple for risk of name clashes in the future) or add another model with a `1:1` relationship to `tasks`.

Issue #960, opened 2019-12-09
[Feature proposal] Whitelist Repositories allowed to dynamically create stacks


indiebrain:
Greetings,

At PowerHRG we desire to implement the ability to automatically deploy a branch as a new "review instance" of our application to our Kubernetes cluster when a Pull Request for that branch is opened - and subsequently clean up the instance when the PR is closed or merged - functionality similar to Heroku's Review Apps. However, we maintain a number of projects and do NOT want to expose this functionality for every application/repository/pipeline.

As we understand it, Shipit offers no mechanism at the repository level to control which repositories can dynamically provision stacks. We'd like to add such a feature to Shipit and are seeking design guidance / gauging appetite for possible upstream contribution. It seems the Stack is the highest level concept which involves an application/repository/pipeline, but this remains insufficient since the Stack also marries a particular branch and environment to the application/repository/pipeline concept. What we _think_ we'd like is a higher level concept which represents the application/repository/pipeline to which stacks could belong. This application/repository/pipeline could then contain its own configuration for the project-level features we desire.

As I've stated earlier, we seek guidance:

1. Are we alone in this problem?
1. What aren't we considering?
1. How are others achieving this goal already given the existing tools?
1. Does this seem like a reasonable feature for Shipit?
1. Does the design direction make sense given the goals of Shipit?
1. Aside from the [workflow guidelines](https://github.com/Shopify/shipit-engine#instructions) listed in the README, what's the least contentious path for upstream adoption of contributions?

casperisfine:
> Are we alone in this problem?

Probably not.

> What aren't we considering?

Stacks aren't designed to be ephemeral, and they're also the "top level object", so you wouldn't have anywhere to put the auto-provisioning code.

> How are others achieving this goal already given the existing tools?

We don't have that exact workflow, but we achieve similar things using tasks on a stack, plus some glue code in a Webhook handler using Shipit's API. e.g. trigger a task with some parameters on PR opening, and another one on PR close.

indiebrain:
@casperisfine , thanks for the swift reply.

> Stacks aren't designed to be ephemeral

[...]

PR #1102, opened 2020-09-03
Add review stacks

indiebrain:
Add Review Stacks
-----------------

This change set represents a months long effort to build, utilize, evolve, and now - hopefully - contribute-for-general-use the concept of "Review Stacks" to shipit-engine.

As discussed in issue #960, it is our desire to use shipit-engine to provide something akin to Heroku Pipeline Review Applications - a dynamically managed application instance. Review Stacks are intended to run the code of a Github Pull Request as a disposable application instance. 

Terminology
-----------

- Review Environment, a dynamically (de)provisioned set of resources used to host an instance of an application
- Review Stack, a dynamically managed stack which relates a Github Pull Request to a Review Environment

Details
-------

Detailed feature documentation and user guide are provided in `docs/review_stacks.md`.

The Review Stack feature's behavior is intended to be flexible and provides options to the host application to:

1. opt Github repositories in/out of this feature
1. provide a mechanism to opt individual Pull Requests in/out of the "Review Stack" feature based on Pull Request labels
1. configure how Review Environments are provisioned - at either globally for the host application or on a per-repository basis

Anecdotes
---------

We've been using this feature since December 2019 to serve workload across our entire development staff. Here are some metrics to give a sense of how exercised this feature is within our organization. Our host application:

- deploys review stacks for 10 distinct repositories
- has created over 2_500 Review Stacks and dynamically managed the (de)provisioning of their Review Environments
- has performed over 13_400 deployments of Review Stack commits to their Review Environments

References
----------

- https://devcenter.heroku.com/articles/pipelines#review-apps
- https://github.com/Shopify/shipit-engine/issues/960
- https://github.com/Shopify/shipit-engine/issues/1093
- https://github.com/Shopify/shipit-engine/issues/1076
- https://github.com/Shopify/shipit-engine/pull/967


casperisfine:
Just to acknowledge this PR. However it's quite massive so that will take me a while to review.

app/models/shipit/merge_request_assignment.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module Shipit
+  class MergeRequestAssignment < Record
```
casperisfine, line 4: Is this dead code? I can't see it used anywhere?

app/models/shipit/provisioning_handler/unregistered_provisioning_handler.rb
```diff
@@ -0,0 +1,26 @@
...
+
+      private
+
+      def lock_and_prevent_transition
+        stack.lock(
+          "Failed to find a provisioning handler named '#{stack.provisioning_handler_name}' in the " \
+          "ProvisioningHandler registry. Have you registered it via Provisioning::Handler.register?",
+          Shipit::AnonymousUser.new
+        )
+        throw(:halt)
```
casperisfine, line 22: Two things here:

  - The only `catch` I see are in `test/`. 
  - `catch` is really not idiomatic at all, I'm really unsure what you are trying to do here.

app/models/shipit/pull_request.rb
```diff
@@ -0,0 +1,66 @@
...
+
+module Shipit
+  class PullRequest < Record
+    include DeferredTouch
+
+    belongs_to :stack
+    belongs_to :user
+    belongs_to :head, class_name: 'Shipit::Commit', optional: true
+
+    has_many :pull_request_assignments
```
casperisfine, line 11: I don't see this used anywhere either. The data is collected but as far as I can tell nothing uses it.

casperisfine: I mean it's exposed in the API, but that's about it.

app/models/shipit/repository.rb
```diff
@@ -14,6 +44,16 @@ class Repository < ApplicationRecord
     validates :name, format: { with: /\A[a-z0-9_\-\.]+\z/ }, length: { maximum: NAME_MAX_SIZE }
 
     has_many :stacks, dependent: :destroy
+    has_many :review_stacks, dependent: :destroy
+
+    enum(
+      provisioning_behavior: {
+        allow_all: 0,
```
casperisfine, line 51: We don't use numeric enums, they're a big PITA long term. See: https://github.com/Shopify/shipit-engine/blob/329272fb79ea1496a004b62a470abdf076d12c57/app/models/shipit/status.rb#L7-L8

casperisfine:
I PRed a few minor changes and cleanup against your branch: https://github.com/powerhome/shipit-engine/pull/76

This PR mostly look good, but there are a few outstanding question / concerns:

  - [ ] The integer based enum should be changed
  - [ ] I don't think we should record Pull Request assignee, as nothing in Shipit uses. I assume you need it for some private feature?
  - [ ] I'm not too fan of the PR -> label many2many. From what I gathered the labels are never queries inside the database, so I wonder if it wouldn't be better to store them inside the `pull_requests` table itself as a serialized field. This would save use the join table.
  - [ ] The `throw(:halt)` thing is fishy.

app/models/shipit/merge_request_assignment.rb
indiebrain: Aye, good catch. It looks like its removal was missed in the merge that renamed the model-formerly-known-as-PullRequest to MergeRequest. Putting together a commit to remove it now.

casperisfine: I removed it in powerhome#76

app/models/shipit/provisioning_handler/unregistered_provisioning_handler.rb
indiebrain: I believe it's used internally by the [state_machines](https://github.com/state-machines/state_machines-activerecord) gem as a transition control mechanism. This is intended to prevent the `ReviewStack`'s `provision_status` state machine from transitioning to a "provisioned" state if the stack couldn't provision.

app/models/shipit/pull_request.rb
indiebrain: Yup, it's intended to expose the collection of pull request assignees through Hook emission to external systems. In our case, we have a system that receives shipit-engine PullRequest hooks and sends notifications to the PullRequest owner and assignees about the "readiness" of a Review Stack for reviews. In this case we use the `PullRequest#assignees` collection to determine part of the recipient list for notifications. 

Does that help add some color?

benlangfeld: This is used by consumers of Shipit hooks in order to send notifications about the deployment of a review stack to the assignees of a PR. It would be very inconvenient to have to store this separately and emit custom hooks from the wrapper application to achieve this.

app/models/shipit/repository.rb
indiebrain: > they're a big PITA long term

...and there's a part of me that knew this. Thanks for the feedback. We wrote this almost a year ago, I'll dig into the context, but I doubt there was a pressing architectural need for a numeric enum. I suspect a change for this will be inbound shortly.

indiebrain:
I am monumentally grateful for the review, @casperisfine - I recognize it's a mammoth amount of code to get through. I'll be dedicating time today to addressing the concerns. Thanks again!

app/models/shipit/provisioning_handler/unregistered_provisioning_handler.rb
casperisfine: I see. I didn't know about that, and it's not really documented in their README, but OK. Can you just add an inline comment explaining it abort the state transition?

app/models/shipit/pull_request.rb
casperisfine: Ok, that's what I suspected. I'm a bit on the fence about this, mostly because it requires 2 extra tables. If it was an inline attribute it wouldn't mind a much.

Is having to query the GitHub API to get the extra details really a big deal? I think we do that quite a bit in our hook consumers. Or alternatively do you think it could be done as a private extension?

benlangfeld: > Is having to query the GitHub API to get the extra details really a big deal?

Our hook consumers (via HTTP) don't have access to the Github API.

app/models/shipit/repository.rb
indiebrain: Addressed by: 33c80107

app/models/shipit/provisioning_handler/unregistered_provisioning_handler.rb
indiebrain: Addressed by: 0af0c0b7

casperisfine:
Thanks for addressing all these concerns. My last sizeable concern is the schema for `labels` and `assignees`. I'm always a bit worried of `has_many though`, not that it should never be used, but IMO it has a sizeable cost, e.g. dangling Label management etc.

In the current instance, it seems to me like `labels` and `assignees` aren't much more than just meta data. Labels is used a bit more to filter which PR need a review, but it's never part of SQL queries either. 

All this to say: how would you feel about saving these inline in the `PullRequest` table as serialized blobs?

indiebrain:
> All this to say: how would you feel about saving these inline in the PullRequest table as serialized blobs?

I can get behind serializing the `Labels` into the Pull Request - I think our original implementation did this. 

The assignees are a bit more tricky / nuanced. IF the assignees were a simple array of  Github logins - simliar to what labels are - there would be no discussion about serializing the list of assgnees it to the Pull Request, however an assignee has much more interesting-to-the-API-consumer information associated with it - name, an email, a Github login, etc. Consequently, Shipit::User already models the concept we're after here. As the system of record of deployment details, assuming that consumers of shipit-engine's API have access to the Github API - or even _should_ assume the responsibility of syncing data from Github - to round out the picture of who is responsible for the deployment seems to diminish the API's ergonomics / usefulness. Architecturally, it _would_ be possible to serialize the association linkage between PullRequest Assignees / User and their users directly on the PullRequest, but this seems flawed. In essence, this pushes the responsibility of managing the "relation" into the shipit-engine Rails application rather than allowing the RDBMS engine do that for what it is expressly designed. Other association modeling options seem to require de-normalizing the Shipit::User, which seems like another path better to avoid.

Does that distinction / discussion make a convincing enough argument to serialize Labels, but  keep the assignees assication/relational modeling as is?

casperisfine:
> Does that distinction / discussion make a convincing enough argument to serialize Labels, but keep the assignees assication/relational modeling as is?

I think so yeah. The fact that  `User` already exist, makes the join table not as bad as adding two tables for labels.

I can leave with that join table I think.

Ok, let me know when `labels` is moved inside PullRequest, and I'll do another round of review. I think the biggest concerns were addressed, after that it will likely be only small refactoring / cleanup etc.

benlangfeld:
@casperisfine We should have that done by the end of the day Eastern. Thank you for the flexibility.

casperisfine:
No rush, it's getting late here so I'll probably only look at it tomorrow morning.

indiebrain:
> Ok, let me know when labels is moved inside PullRequest, and I'll do another round of review. 

Addressed by: https://github.com/Shopify/shipit-engine/pull/1102/commits/73a3f3b5f33297f0abc193c7d4608c4085e6ac07

benlangfeld:
Morning @casperisfine . Is it possible for you to trigger another CI run here in order to establish that this is ready to go? Also, do you have any new review comments that we need to address in order for it to be merged?

casperisfine:
> Is it possible for you to trigger another CI run

Done.

> Also, do you have any new review comments

I have a couple high priority stuff to deal with right now, but I should be able to do a second review later today. Either way I won't forget this, it's super high on my todo list.

app/models/shipit/webhooks.rb
```diff
@@ -5,6 +5,16 @@ class << self
...
           'push' => [Handlers::PushHandler],
+          'pull_request' => [
+            Handlers::PullRequest::OpenedHandler,
+            Handlers::PullRequest::ClosedHandler,
+            Handlers::PullRequest::ReopenedHandler,
+            Handlers::PullRequest::EditedHandler,
+            Handlers::PullRequest::AssignedHandler,
+            Handlers::PullRequest::LabeledHandler,
+            Handlers::PullRequest::UnlabeledHandler,
+            Handlers::PullRequest::LabelCapturingHandler,
```
casperisfine, line 16: Not a deal breaker, but I feel like there's lots of duplication in all these handlers, and that a single one handling all the events would be much more concise and easier to grasp.

But again, not a deal breaker, it can be refactored post merge in a follow up if there's no other blocker.

db/migrate/20200823172649_add_head_reference_to_pull_requests.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+class AddHeadReferenceToPullRequests < ActiveRecord::Migration[6.0]
+  def change
+    add_reference :pull_requests, :head
```
casperisfine, line 4: There's a bunch of tables that are migrated several times, for instance this `add_reference` could totally be part of the previous `RecreateShipitPullRequests`. I understand that this is a dump of an iterative development, but if we're to merge this one shot it would be worth merging merging the migrations on a per table basis (one for each table).

test/models/shipit/webhooks/handlers/pull_request/closed_handler_test.rb
```diff
@@ -0,0 +1,192 @@
...
+          test "archives stacks for repos that are tracked" do
+            create_stack
+            assert_difference -> { Shipit::Stack.not_archived.count }, -1 do
+              Shipit::Webhooks::Handlers::PullRequest::ClosedHandler.new(payload_parsed(:pull_request_closed)).process
+            end
+          end
+
+          test "ignored duplicate deliveries" do
+            Shipit::Webhooks::Handlers::PullRequest::ClosedHandler.new(payload_parsed(:pull_request_opened)).process
+            Shipit::Webhooks::Handlers::PullRequest::ClosedHandler.new(payload_parsed(:pull_request_closed)).process
```
casperisfine, line 33: More of a nitpick, but I see many test using fully qualified class names when they could use much shorter ones.

casperisfine:
Ok, so the migration thing is my last real concerns, the other two are optional, feel free to disregard them.

To clarify my concern about migrations: our DB is quite big, and migrating it can sometime be tricky, so I'd rather limit the amount of `ALTER`, the ones that can be merged together should be merged together. It would also be a bit cleaner.

benlangfeld:
Migration squashing coming right up @casperisfine

casperisfine:
<3

benlangfeld:
@casperisfine Reduced to 4 migrations, one per table. Do you prefer it like that, or a single migration for the whole feature?

casperisfine:
> Do you prefer it like that, or a single migration for the whole feature?

One per table is great, it makes it much easier to retry them in case of timeout or whatever else.

Anything else you'd like to add or is it good to go?

casperisfine (approved):

benlangfeld:
@casperisfine 🚢

PR #1103, opened 2020-09-10
Add skipped as a valid check run conclusion

DazWorrall:
Similar to https://github.com/Shopify/shipit-engine/pull/1043, we're seeing exceptions in the wild because a new conclusion has been added - [skipped](https://developer.github.com/changes/2020-05-08-skipped-conclusion/):

> If a check run or suite is never started, you'll see the `skipped` conclusion. This is a common check run conclusion for GitHub Actions that have a job using an `if` condition.

As skipped means _intentionally_ not ran, I think we should treat is as a success state for the purposes of shipping.

### DIFF
--- a/app/models/shipit/check_run.rb
+++ b/app/models/shipit/check_run.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 module Shipit
   class CheckRun < ApplicationRecord
-    CONCLUSIONS = %w(success failure neutral cancelled timed_out action_required stale).freeze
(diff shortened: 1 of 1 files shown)


casperisfine (approved):

DazWorrall:
> As skipped means _intentionally_ not ran, I think we should treat is as a success state for the purposes of shipping.

I should say this is a bit misleading of course - from the Shipit UI it will look like this ran and passed, when it didn't - but 'unknown' isn't right either, and we cant do the equivalent of 'ignore' here afaict. Maybe something we want to refactor in future, if this becomes a big thing.

PR #1105, opened 2020-10-05
Use bytesize for commit message string truncation.

ajshepley:
I noticed that we were using size instead of bytesize when truncating strings, which may fail if we run into a commit message containing multibyte chars.

I figure it's worth guarding against this.

### DIFF
--- a/test/models/commits_test.rb
+++ b/test/models/commits_test.rb
@@ -170,6 +170,15 @@ module Shipit
       assert_equal limit, @commit.message.bytesize
     end
 
+    test "#message= truncates multibyte messages" do
+      skip unless Shipit::Commit.columns_hash['message'].limit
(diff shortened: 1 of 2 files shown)


test/models/commits_test.rb
```diff
@@ -170,6 +170,15 @@ class CommitsTest < ActiveSupport::TestCase
       assert_equal limit, @commit.message.bytesize
     end
 
+    test "#message= truncates multibyte messages" do
+      skip unless Shipit::Commit.columns_hash['message'].limit
+      limit = Shipit::Commit.columns_hash['message'].limit
+
+      @commit.update!(message: '国' * limit)
```
rdji123, line 177: A very national message ;)

rdji123 (approved):

casperisfine (approved):

PR #1107, opened 2020-10-12
Allow variance in task execution strategy

indiebrain:
Why
===

The host application _may_ want more control over how work is scheduled and executed given the topology / capabilities of its environment. 

An example, _we_ deploy our host application into a Kubernetes cluster. We'd like the ability to schedule Shipit::Deploy tasks to run in their own pod, rather than relying on a processing the Shipit::Deploy task inside the context of the Sidekiq worker container - which _may_ be processing other jobs, like Sidekiq health checks, etc.

How
===

This change set extracts the current task execution logic into a "Default Execution Strategy" - a vanilla instance of shipit-engine should NOT be affected by this change. We also expose an interface whereby the host application _may_ register a custom "execution strategy" for a given task type - overriding the default. The `PerformTaskJob` looks-up the strategy to use for a given task, then executes the task using the appropriate execution strategy.

### DIFF
--- a/docs/task_execution_strategy.md
+++ b/docs/task_execution_strategy.md
@@ -1,0 +1,40 @@
+# Task execution strategies
+
+# Default
+
+For most applications, the default execution strategy is sufficient. This strategy processes a task within the context of a shipit-engine Sidekiq worker. shipit-engine provides the `Shipit::TaskExecutionStrategy::Default` strategy to do this - on a vanilla shipit-engine instance _this_ is the default execution strategy.
+
+# Custom task execution strategies
+
+## Registering a custom TaskExecutionStrategy
+
+1. Create a custom TaskExecutionStrategy by extending `Shipit::TaskExecutionStrategy::Base` and implementing the `#execute` method. The `Shipit::Task` to be processed is available as the `task` local - courtesy of the inherited `Shipit::TaskExecutionStrategy::Base#initialize` method:
+    ```ruby
+    class KubernetesPodExecutionStrategy < Shipit::TaskExecutionStrategy::Base
+      def execute
+        # Example: apply kubernetes templates to create a pod in which the job should
+        # be executed passing along the task to be executed
+        #
+        # Consider running a Kubernetes Pod of the host-application's image
+        # whose command - in turn - runs a rake task invoking the
+        # Shipit::TaskExecutionStrategy::Default stratgy on the task.
+        #
+        # Pseudo code below:
+        kube.apply(task_execution_pod_template)
+      end
+
+      def task_execution_pod_template
+        <<~YAML
+        ...
+          command: [ "bin/rails" ]
+          args: [ "shipit:task:run[#{task.id}]" ]
+        ...
+        YAML
+      end
+    end
+    ```
+2. Register the custom strategy. For example, the following uses the above `KubernetesPodExecutionStrategy`
+    ```ruby
+    # probably in the host application's shipit initializer
+    Shipit.task_execution_strategy = KubernetesPodExecutionStrategy
+    ```
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -38,7 +38,7 @@ Naming/PredicateName:
 
 Lint/RescueException:
   Exclude:
-    - app/jobs/shipit/perform_task_job.rb
+    - app/models/shipit/task_execution_strategy/default.rb
 
 Lint/SuppressedException:
   Enabled: false
--- a/app/jobs/shipit/perform_task_job.rb
+++ b/app/jobs/shipit/perform_task_job.rb
@@ -4,104 +4,10 @@ module Shipit
     queue_as :deploys
 
     def perform(task)
-      @task = task
-      @commands = Commands.for(@task)
-      unless @task.pending?
-        logger.error("Task ##{@task.id} already in `#{@task.status}` state. Aborting.")
-        return
-      end
-      run
-    ensure
-      @commands.clear_working_directory
-    end
-
-    def run
-      @task.ping
-      @task.run!
-      checkout_repository
-      perform_task
-      @task.write("\nCompleted successfully\n")
-      @task.report_complete!
-    rescue Command::TimedOut => error
-      @task.write("\n#{error.message}\n")
-      @task.report_timeout!(error)
-    rescue Command::Error => error
-      @task.write("\n#{error.message}\n")
-      @task.report_failure!(error)
-    rescue StandardError => error
-      @task.report_error!(error)
-    rescue Exception => error
-      @task.report_error!(error)
-      raise
-    end
-
-    def abort!(signal: 'TERM')
-      pid = @task.pid
-      if pid
-        @task.write("$ kill #{pid}\n")
-        Process.kill(signal, pid)
-      else
-        @task.write("Can't abort, no recorded pid, WTF?\n")
-      end
-    rescue SystemCallError => error
-      @task.write("kill: (#{pid}) - #{error.message}\n")
-    end
-
-    def check_for_abort
-      @task.should_abort? do |times_killed|
-        if times_killed > 3
-          abort!(signal: 'KILL')
-        else
-          abort!
-        end
-      end
-    end
-
-    def perform_task
-      capture_all!(@commands.install_dependencies)
-      capture_all!(@commands.perform)
-    end
-
-    def checkout_repository
-      unless @commands.fetched?(@task.until_commit).tap(&:run).success?
-        # acquire_git_cache_lock can take upto 15 seconds
-        # to process. Try to make sure that the job isn't
-        # marked dead while we attempt to acquire the lock.
-        @task.ping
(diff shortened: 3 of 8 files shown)


benlangfeld:
@indiebrain Could you take a look at the cop violation here?

indiebrain:
Aye, @benlangfeld  I have considered it. Rubocop is mad about rescuing `Exception` and is suggesting we handle `StandardError` instead. Since this bit already handles `StandardError`, and the [source of this logic explicitly handles `Exception`](https://github.com/Shopify/shipit-engine/blob/master/app/jobs/shipit/perform_task_job.rb#L33-L35) it seems to me that the use of `Exception` is intentional so I left it. 

I _could_ disable the cop to prevent the linter from tripping, but I could also try turning this into a "catch all" to avoid the cop, but the logic seems intentional to me. Thoughts?

benlangfeld:
> Aye, @benlangfeld I have considered it. Rubocop is mad about rescuing `Exception` and is suggesting we handle `StandardError` instead. Since this bit already handles `StandardError`, and the [source of this logic explicitly handles `Exception`](https://github.com/Shopify/shipit-engine/blob/master/app/jobs/shipit/perform_task_job.rb#L33-L35) it seems to me that the use of `Exception` is intentional so I left it.
> 
> I _could_ disable the cop to prevent the linter from tripping, but I could also try turning this into a "catch all" to avoid the cop, but the logic seems intentional to me. Thoughts?

I think we need to update [this](https://github.com/Shopify/shipit-engine/blob/master/.rubocop.yml#L39-L41) then.

indiebrain:
Thanks @benlangfeld . I missed that bit.

benlangfeld (approved):

indiebrain:
Any guidance / suggestions from the project maintainers on this? /cc @casperisfine

casperisfine:
Sorry, I've been quite busy the last few month and just got back from vacation.

This PR is something I had in mind for a while but never got around to, so yes I'm very interested. I'll try to find the time to review this ASAP.

indiebrain:
Danke, friend. I figured the "pause" in conversation was attributable to the end-of-year / holiday break. Just wanted to ping to make sure this didn't fall through the cracks. Thanks for the swift reply.

casperisfine (commented):
So I'm kinda fine with this simple extraction, however let me ask a couple open ended questions:

  - Does the strategy selection based on Task class really makes sense? From the use cases I envision, a selection based on the stack would be more likely to be used. e.g. Stack X have to be executed in docker image `foo`, stack Y in docker image `bar`
  - Should the strategy be fully responsible for managing the entire state machine? I see this as quite "private". I'd be worried it would break easily. Also things like wrting logs etc should probably be in the base class as custom strategies are very likely to use them.

indiebrain:
> Does the strategy selection based on Task class really makes sense? From the use cases I envision, a selection based on the stack would be more likely to be used. e.g. Stack X have to be executed in docker image foo, stack Y in docker image bar

That doesn't _feel_ wrong to me. This might require a bit of redesign of the proposed solution. 

> Should the strategy be fully responsible for managing the entire state machine? I see this as quite "private". I'd be worried it would break easily. Also things like wrting logs etc should probably be in the base class as custom strategies are very likely to use them.

Again at first glance doesn't feel wrong. Perhaps there's some role burred in here - a task executor perhaps - whose job is to mediate between the Task for lifecycle events and run/monitor the execution of the strategy? I'd have to ruminate on it a bit. I'd want to make sure whatever's coordinating between the task and its execution could anticipate the ways execution could co sideways and appropriately recover / record such events.

benlangfeld:
> > Does the strategy selection based on Task class really makes sense? From the use cases I envision, a selection based on the stack would be more likely to be used. e.g. Stack X have to be executed in docker image foo, stack Y in docker image bar
> 
> That doesn't _feel_ wrong to me. This might require a bit of redesign of the proposed solution.

The reason we did it this way is that the strategy itself is global to the installation, and can use values from the DeploySpec to decide how to operate. It wasn't necessary for us to select the strategy per stack for the use case specified (different docker image) since this is conditional logic inside the strategy itself.

casperisfine:
> The reason we did it this way is that the strategy itself is global to the installation

Yes, but that's seems a bit too limited to me. I'd have been fine with a single global implementation, but here there's code to select distinct implementation for `Task/Deploy/Rollback`, I really don't see the advantage of this compared to a single global one.

But if we're to extract this, I see value in the implementation being selectable on a per stack basis. This way you could have some stack use say a `ruby:2.5` image, and others a `ruby:2.7` image. That's an old environment isolation problem we never solved, so it seems like a good opportunity to do that.

benlangfeld:
> > The reason we did it this way is that the strategy itself is global to the installation
> 
> Yes, but that's seems a bit too limited to me. I'd have been fine with a single global implementation, but here there's code to select distinct implementation for `Task/Deploy/Rollback`, I really don't see the advantage of this compared to a single global one.
> 
> But if we're to extract this, I see value in the implementation being selectable on a per stack basis. This way you could have some stack use say a `ruby:2.5` image, and others a `ruby:2.7` image. That's an old environment isolation problem we never solved, so it seems like a good opportunity to do that.

You can do that with a single global strategy too, putting those image specifications in the `DeploySpec`. If we were to simplify this to be a single global strategy would you accept this, leaving it open for extension to per-stack selection in the future?

casperisfine:
> You can do that with a single global strategy too

Sure, but then why the need for selection based on `Task/Deploy/Rollback`?

benlangfeld:
> > You can do that with a single global strategy too
> 
> Sure, but then why the need for selection based on `Task/Deploy/Rollback`?

I think we can remove this distinction to simplify this feature without negative impact on our use case. If that makes this easier to incorporate, we'd be happy to do it.

casperisfine:
Ok, then I'll be ok with a single swappable strategy. It should be easy enough to implement a MultiStrategy connector for those who need it.

indiebrain:
@benlangfeld @casperisfine , does https://github.com/Shopify/shipit-engine/pull/1107/commits/91b0610ecf35dd01ab4f31d49075cbea753452c4 look a bit more like what we're going for?

docs/task_execution_strategy.md
```diff
@@ -0,0 +1,40 @@
...
+      def execute
+        # Example: apply kubernetes templates to create a pod in which the job should
+        # be executed passing along the task to be executed
+        #
+        # Consider running a Kubernetes Pod of the host-application's image
+        # whose command - in turn - runs a rake task invoking the
+        # Shipit::TaskExecutionStrategy::Default stratgy on the task.
+        #
+        # Pseudo code below:
+        kube.apply(task_execution_pod_templte)
```
casperisfine, line 23: typo: `template`

app/models/shipit/task_execution_strategy.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module Shipit
+  module TaskExecutionStrategy
+    class << self
+      def current
+        @strategy || default
+      end
+
+      def register(strategy)
```
casperisfine, line 10: I don't really see the point of having a `register` method if there's only once active strategy. `Shipit.task_execution_strategy = ...` would do fine.

docs/task_execution_strategy.md
indiebrain: Addressed by: https://github.com/Shopify/shipit-engine/pull/1107/commits/ad744b33eee2b0de1c42842ec83ad71a5a149b03

app/models/shipit/task_execution_strategy.rb
indiebrain: Addressed by https://github.com/Shopify/shipit-engine/pull/1107/commits/ad50b895fa98515030aa503a022dbb82110f54a8

docs/task_execution_strategy.md
```diff
@@ -0,0 +1,40 @@
...
+          args: [ "shipit:task:run[#{task.id}]" ]
+        ...
+        YAML
+      end
+    end
+    ```
+2. Register the custom strategy as the `#current` strategy. For example, the following uses the above `KubernetesPodExecutionStrategy`
+    ```ruby
+    # probably in the host application's shipit initializer
+    Shipit::TaskExecutionStrategy.register(KubernetesPodExecutionStrategy)
```
casperisfine, line 39: This needs to be updated now.

casperisfine (approved):
One last change and then LGTM. Please also squash you commits.

docs/task_execution_strategy.md
indiebrain: Dang it. Thanks for the save!

indiebrain: Addresed by: https://github.com/Shopify/shipit-engine/pull/1107/commits/4dd7c27d01da71bd87140e140463007c9e7fc235

casperisfine:
Thanks!

indiebrain:
Many thanks, friend! I appreciate your time and attention.

PR #1108, opened 2020-10-13
Update rubocop-shopify to 1.0.5

ajshepley:
I was getting deprecated cop errors, so I figured it was time to update rubocop-shopify (via `bundle lock --update=rubocop-shopify`).

I noticed some errors for not calling super and not using parenthesis, so I ran `rubocop -a` for the latter and manually added calls (or ignored the cop; I deferred to that rather than creating empty Initializers) for the former.


lib/shipit/github_app.rb
```diff
@@ -52,7 +52,7 @@ def initialize(config)
       oauth = (@config[:oauth] || {}).with_indifferent_access
       @oauth_id = oauth[:id]
       @oauth_secret = oauth[:secret]
-      @oauth_teams = Array.wrap(oauth[:teams] || oauth[:teams])
+      @oauth_teams = Array.wrap(oauth[:teams])
```
ajshepley, line 55: I'm not sure if this was accomplishing something or if it was just a typo. Maybe there's some obvious Ruby thing I'm missing here.

casperisfine (approved):

lib/shipit/github_app.rb
casperisfine: Typo or brainfart most likely.

gigr (approved):

PR #1109, opened 2020-10-14
Retry tasks (deploys) that timeout

nishanthooda:
## What
- adds logic to retry tasks (more specifically deploys and rollbacks) that timeout

## Why
- we have quite a few deploy timeouts that are solved simply by retrying. This PR automates the retry process
- for some data see: https://app.mode.com/editor/shopify/reports/28f5c21e2860 (note that currently core deploy timeouts are recorded as failures as oppose to timeouts - a krane fix for this is WIP)

## How
**new fields on the `Task` model**:
- `max_retries` tracks the number of retries we will perform - when field is nil, we do not perform any retries
- `retry_attempt` tracks which retry attempt number this task is - this allows us to keep retrying until we reach the max number of retries set on the task

**when to retry**:
- whenever the state machine in the task model transitions to the `timedout` state, it will trigger the `retry_if_necessary` method
- `retry_if_necessary` will only retry if the `max_retires` is not nil and the current `retry_attempt` is less than the `max_retries`
- retries can be enabled on a stack by setting the max retries on a deploy or timeout using the `retries_on_timeout` field in the `shipit.yml`

**the retry process**:
- on a retry we create a new task, which is duplicated from the timed out task with the retry attempt incremented
- we enqueue this new task

## Tophat Process
1. Run `shipit-engine` locally using the test dummy app
2. Add this repo as a stack: https://github.com/Shopify/shipit-retry-tester
3. Make a commit to the `shipit-retry-tester` repo (can just add a character to the README)
4. Click `Refresh statuses & commits` on the `shipit-retry-tester` stack, you will now see the new commit you made
5. Deploy this commit
6. This deploy will use the [`test_timeout`](https://github.com/Shopify/shipit-retry-tester/blob/main/test_timeout) deploy script, which will timeout 50% of the time.
7. If it does not timeout, create new commit and try deploying it again
8. If it does timeout, verify on the stack page that it automatically created and launched a new task (i.e. deploy)

## Considerations
- [Thoughts on `enqueue` v.s. `run_now` for the retry task ?](https://github.com/Shopify/shipit-engine/pull/1109/files#r504380853)

cc @ajshepley

### DIFF
--- a/app/models/shipit/task.rb
+++ b/app/models/shipit/task.rb
@@ -88,7 +88,7 @@ module Shipit
       end
 
       after_transition any => :timedout do |task|
-        retry_if_necessary
+        task.retry_if_necessary
       end
 
       event :run do
@@ -388,7 +388,7 @@ module Shipit
 
     # Make private before commiting
     def retry_if_necessary
-      if self.retry_attempt < self.max_retries
+      if retry_attempt < max_retries
         retry_task = duplicate_task
         retry_task.retry_attempt = duplicate_task.retry_attempt + 1
         retry_task.save!
--- a/app/models/shipit/task.rb
+++ b/app/models/shipit/task.rb
@@ -423,7 +423,7 @@ module Shipit
     end
 
     def duplicate_task
-      copy_task = self.dup
+      copy_task = dup
       copy_task.status = 'pending'
       copy_task.started_at = nil
       copy_task.ended_at = nil
--- a/app/models/shipit/task.rb
+++ b/app/models/shipit/task.rb
@@ -87,6 +87,10 @@ module Shipit
         task.async_update_estimated_deploy_duration
       end
 
+      after_transition any => :timedout do |task|
+        retry_if_necessary
+      end
+
       event :run do
         transition pending: :running
       end
@@ -382,6 +386,17 @@ module Shipit
         .reject(&:alive?)
     end
 
+    # Make private before commiting
+    def retry_if_necessary
+      if self.retry_attempt < self.max_retries
+        retry_task = duplicate_task
+        retry_task.retry_attempt = duplicate_task.retry_attempt + 1
+        retry_task.save!
+
+        retry_task.enqueue
+      end
+    end
+
     private
 
     def prevent_concurrency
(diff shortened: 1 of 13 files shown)


app/models/shipit/task.rb
```diff
@@ -382,6 +386,18 @@ def self.zombies
...
 
+    def retry_if_necessary
+      return if max_retries.nil?
+
+      if retry_attempt < max_retries
+        retry_task = duplicate_task
+        retry_task.retry_attempt = duplicate_task.retry_attempt + 1
+        retry_task.save!
+
+        retry_task.enqueue
```
nishanthooda, line 397: There is also a good argument here to run right away instead of enqueueing, so we can get the final state of a task (even if all retries time out) and take appropriate action without having to wait on the job to be enqueued. 

We ultimately elected enqueue because:
- same reason as using enqueue for tasks right now: allow tasks to be performed asynchronously
- the app does not process a large number of jobs, so there likely won't be a big delay in getting the final result of the task as chances are the job will be enqueued in the very near future

Would love to hear others thoughts on this

gigr (approved):

app/models/shipit/task.rb
casperisfine: Operation wise it's better to re-enqueue. Very long running jobs make shipit slow to deploy.

casperisfine (approved):
Implementation look sane to me. Design wise I still believe retrying commands rather than tasks would have made more sense, but hey.

rdji123 (approved):

app/serializers/shipit/task_serializer.rb
```diff
@@ -6,7 +6,7 @@ class TaskSerializer < ActiveModel::Serializer
     has_one :author
     has_one :revision, serializer: ShortCommitSerializer
 
-    attributes(:id, :url, :html_url, :output_url, :type, :status, :action, :title, :description, :started_at, :ended_at, :updated_at, :created_at, :env, :ignored_safeties)
+    attributes(:id, :url, :html_url, :output_url, :type, :status, :action, :title, :description, :started_at, :ended_at, :updated_at, :created_at, :env, :ignored_safeties, :max_retries, :retry_attempt)
```
ajshepley, line 9: For posterity: These changes came from https://github.com/Shopify/shipit-engine/pull/1111/.

@nishanthooda Another option is to keep the second PR around until the first one is merged, and then retarget to master. This _may_ require re-approvals from reviewers. But it also prevents "surprise" files in the first PR. YMMV, no approach is perfect.

ajshepley (approved):

PR #1110, opened 2020-10-20
Pull git recursivelly for tmp directories

jsaubry:
This [native gem](https://github.com/Shopify/woff2-encoder) contains git submodules. Shipit ignore those submodules when building the gem.


casperisfine:
You linked to a private repo. But yeah, this `--recursive` should be there.

PR #1111, opened 2020-10-21
Add retry fields to task serializer 

nishanthooda:
## What
- update the `task_serializer` to send retry fields: `retry_attempt` and `max_retries`

## Why
- so spy can receive a webhook and parse that this deploy/rollback is one that is automatically being retried
- see PR with spy changes here: https://github.com/Shopify/spy/pull/8414

### DIFF
--- a/app/models/shipit/task.rb
+++ b/app/models/shipit/task.rb
@@ -387,7 +387,7 @@ module Shipit
     end
 
     def retry_if_necessary
-      return if max_retries.nil?
+      return unless retries_configured?
 
       if retry_attempt < max_retries
         retry_task = duplicate_task
@@ -396,6 +396,10 @@ module Shipit
 
         retry_task.enqueue
       end
+    end
+
+    def retries_configured?
+      !max_retries.nil? && max_retries > 0
     end
 
     private
--- a/app/serializers/shipit/task_serializer.rb
+++ b/app/serializers/shipit/task_serializer.rb
@@ -6,7 +6,7 @@ module Shipit
     has_one :author
     has_one :revision, serializer: ShortCommitSerializer
 
-    attributes(:id, :url, :html_url, :output_url, :type, :status, :action, :title, :description, :started_at, :ended_at, :updated_at, :created_at, :env, :ignored_safeties)
+    attributes(:id, :url, :html_url, :output_url, :type, :status, :action, :title, :description, :started_at, :ended_at, :updated_at, :created_at, :env, :ignored_safeties, :max_retries, :retry_attempt)
 
     def revision
       object.until_commit
--- a/test/models/tasks_test.rb
+++ b/test/models/tasks_test.rb
@@ -68,5 +68,16 @@ module Shipit
         task.retry_if_necessary
       end
     end
+
+    test "#retries_configured? returns true when max_retries is not nil and is greater than zero" do
+      task_with_three_retries = shipit_tasks(:shipit)
+      assert_predicate task_with_three_retries, :retries_configured?
+
+      task_with_nil_retries = shipit_tasks(:shipit2)
+      refute_predicate task_with_nil_retries, :retries_configured?
+
+      task_with_zero_retries = shipit_tasks(:shipit_restart)
+      refute_predicate task_with_zero_retries, :retries_configured?
+    end
   end
 end


app/models/shipit/task.rb
```diff
@@ -340,7 +340,8 @@ def emit_hooks
     end
 
     def hook_event
-      self.class.name.demodulize.underscore.to_sym
+      event = self.class.name.demodulize.underscore.to_sym
+      append_retry_if_necessary(event)
```
ajshepley, line 344: I think [`task_serializer`](https://github.com/Shopify/shipit-engine/blob/dd5fb153d14be67c3ad12adc7a9c86b9c454bbc3/app/serializers/shipit/task_serializer.rb#L9) or `deploy_serializer` is used for these webhooks, so could we instead add `retry_attempt` to the serializer's attribute fields and avoid the need for a different event type and this wrapper function?

Then the Spy end should only need to check that `retry_attempt > 0`.

app/models/shipit/task.rb
```diff
@@ -398,6 +399,10 @@ def retry_if_necessary
       end
     end
 
+    def retries_configured?
+      !max_retries.nil? && max_retries > 0
```
ajshepley, line 403: If we assume that max_retries will always be a number or nil, then this conditional can be expressed as `max_retries&.nonzero?`.

app/models/shipit/task.rb
```diff
@@ -429,5 +434,11 @@ def duplicate_task
       copy_task.ended_at = nil
       copy_task
     end
+
+    def append_retry_if_necessary(event)
+      return event unless retries_configured?
+      return "#{event}_retry".to_sym if (status == "running" && retry_attempt > 0) || status == "timedout"
```
ajshepley, line 440: Is this codepath only called when a Task is in the running state? That is, we may send Deploy/Task webhooks on creation, completion, etc. and so their status won't be "Running" in those cases.

Or is the intention to only differentiate a Deploy from a Retried deploy when it's a Running deploy event for a new retried deploy? (Oof, word salad.)

I think an approach like [my other comment](https://github.com/Shopify/shipit-engine/pull/1111/files#r509317438) would allow the receiver to make that differentiation if they so choose and remove the complexity here.

It's possible that some other webhook receivers may care to know if a failed or newly started deploy was also an automatic retry (e.g. I can envision some UI elements in ShipitNext to differntiate autodeploys like CD and retries from manual deploys).

ajshepley, line 440: A bit pedantic/YAGNI/futureproofing: it might be a bit more correct if the status check here was `RETRIABLE_STATUSES.include?(status) || (status == "running" ...)` but that's not as important until we have more than one retriable status.

app/models/shipit/task.rb
nishanthooda: Hm, I want to opt for the way I have it currently because its easier to read. And although this should never be the case, the greater than check also guards against negative values for `max_retries`.
Also `nonzero?` returns does not return boolean values so we would have to append a `!!` to convert the values into booleans. Not a big deal but just a little more confusing and the way we have it right now isn't too messy imo

app/models/shipit/task.rb
nishanthooda: > the intention to only differentiate a Deploy from a Retried deploy when it's a Running deploy event for a new retried deploy

This^ is the intention. But you make a good point about this logic lying in the receiver as oppose to over here. So, I've taken it out and moved it to the spy PR 👍

nishanthooda: Logic has been removed as per https://github.com/Shopify/shipit-engine/pull/1111#discussion_r510295799

app/models/shipit/task.rb
nishanthooda: Good point, updated to remove the `append_retry_if_necessary` method and added the `max_retries` and `retry_attempt` values to the `task_serializer`

gigr (approved):

app/models/shipit/task.rb
ajshepley: > Also nonzero? returns does not return boolean values

Ah, I hadn't read the docs.

rdji123 (approved):

Issue #1104, opened 2020-09-16
Migrate from travis-ci.org


ajshepley:
Travis CI recently sent out this messaging to users.

>Hi there!

>As you’ll no doubt have noticed, Travis CI has, for some time, been split between travis-ci.com and travis-ci.org...
>
>In an effort to ensure that all of our users - whether you build open-source, public or private repositories - receive regular feature updates, security patches and UX/UI enhancements, we are announcing that travis-ci.org will be officially closed down completely on December 31st, 2020, allowing us to focus all our efforts on bringing new features and fixes to travis-ci.com and all of our awesome users like yourself on the travis-ci.com domain. 

...

>We encourage you to migrate your existing repositories that are currently on travis-ci.org over to travis-ci.com as soon as possible, enabling you to identify any additional changes required to your code or configuration well in advance of the December 31st deadline. We’ve documented everything you need to know in order to migrate your repositories, and of course, you can reach out on the community forum for any additional help.
>
>If you have any questions or concerns, please feel free to contact our support team who will be happy to talk through any migration queries you may have!

In short, travis-ci.org will be shut down by December 31st 2020, so we'll need to adjust shipit-engine.

We can migrate to travis-ci.com using their [linked migration guide](https://docs.travis-ci.com/user/migrate/open-source-repository-migration).

Alternatively, if we're confident in how GitHub actions have been working, we may be able to eschew Travis CI altogether.

A quick grep shows that our only text reference to travis-ci.org is for the build status in our README, and nothing seems suspicious in our [.travis.yml](https://github.com/Shopify/shipit-engine/blob/master/.travis.yml)

casperisfine:
GitHub Actions have been working fine. we can simple 🔥 the `.travis.yml`

PR #1112, opened 2020-10-26
Remove Travis CI integration.

ajshepley:
Travis-ci.org is being shut down by December 31st, 2020 in favor of [travis-ci.com](https://docs.travis-ci.com/user/migrate/open-source-repository-migration).

We could migrate, but we've been using GitHub Actions concurrently for CI for a while now, and it's been working out fine.

As such, I think we can remove the Travis CI integration altogether.

We may additionally need to go into the TravisCI page and delete the builds/pipelines, as I don't know if they'll go away or GC automatically.

Resolves https://github.com/Shopify/shipit-engine/issues/1104


casperisfine (approved):

.travis.yml
```diff
@@ -1,34 +0,0 @@
-language: ruby
-sudo: false
-script: 'script/cibuild'
```
gigr, line 3: Looks like we can also remove this file.

ajshepley: I couldn't find any consumers, and the file and directory don't seem to be referenced by the GitHub Actions [workflow file](https://github.com/Shopify/shipit-engine/blob/master/.github/workflows/main.yml) but I'm not 100% sure. I'll remove that in a subsequent PR, it'll be easier to revert/rollback that way.

PR #1119, opened 2020-11-11
Support merge commits in assert-npm-version-tag

gmcgibbon:
I think it reasonable to support merged version bumps in node packages. This commit changes the tag assertion to pass if the tag is on either HEAD or HEAD without merge commits.

### DIFF
--- a/lib/snippets/assert-npm-version-tag
+++ b/lib/snippets/assert-npm-version-tag
@@ -11,9 +11,10 @@ if [ $? != '0' ]; then
 fi
 
 TAG_REV=`git rev-list -n1 v$VERSION`
-HEAD_REV=`git rev-parse HEAD`
+HEAD_REV=`git rev-list -n1 HEAD`
+HEAD_REV_WITHOUT_MERGES=`git rev-list -n1 --no-merges HEAD`
 
-if [ $TAG_REV != $HEAD_REV ]; then
+if [ $TAG_REV != $HEAD_REV ] || [ $TAG_REV != $HEAD_REV_WITHOUT_MERGES ]; then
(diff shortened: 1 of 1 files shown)


lib/snippets/assert-npm-version-tag
```diff
@@ -11,9 +11,10 @@ if [ $? != '0' ]; then
 fi
 
 TAG_REV=`git rev-list -n1 v$VERSION`
-HEAD_REV=`git rev-parse HEAD`
+HEAD_REV=`git rev-list -n1 HEAD`
+HEAD_REV_WITHOUT_MERGES=`git rev-list -n1 --no-merges HEAD`
 
-if [ $TAG_REV != $HEAD_REV ]; then
+if [ $TAG_REV != $HEAD_REV ] || [ $TAG_REV != $HEAD_REV_WITHOUT_MERGES ]; then
```
ismail-syed, line 17: I'm trying to recall what happened yesterday, could be wrong. But, I don't think this will actually solve the issue you had yesterday. `$TAG_REV != $HEAD_REV` will still yield true and error out, no?

ismail-syed (commented):
IIRC, much of this logic and need for this assertion was because our ruby deploy also has one, https://github.com/Shopify/shipit-engine/blob/master/lib/snippets/assert-gem-version-tag.

gmcgibbon:
> IIRC, much of this logic and need for this assertion was because our ruby deploy also has one, https://github.com/Shopify/shipit-engine/blob/master/lib/snippets/assert-gem-version-tag. 

Oh, weird. I've definitely pushed a gem version with a merge commit on HEAD before. I can't find an example, but looking at deployment logs I don't see that tag assert actually running on any of my recent gem releases. More investigation needed.

PR #1120, opened 2020-11-13
Do not auto retry if stack is locked

nishanthooda:
## What

- Will not automatically retry a task is stack is locked

## Why

- If a stack is locked mid deploy and deploy times out, we do not want to automatically retry it


gigr (approved):

ajshepley (approved):

ajshepley:
We could also make this configurable on the shipit.yml or the shipit stack itself at some point in the future.

PR #1127, opened 2020-12-06
Add ENVIRONMENT to StackCommands

systemist:
This allows `$ENVIRONMENT` to be used in `fetch`.


casperisfine:
Thank you. I pushed 05d94d8 instead.

PR #1128, opened 2020-12-11
Include plugins on other stack pages

viniciusgama:
## Why
Shipit currently allows the parent application to define extension and/or modification to a page by the use of plugins. Unfortunately the defined plugins are only included on the task output page, a.k.a the deployment logs page. Considering that the plugin is defined at the stack level, it would be helpful if it is also enable on other pages that involve the stack. This would allow a new level o customization on the stack pages. An example would be extra warnings on the commits and tasks page.

## How
We are basically moving the inclusion of the plugins scripts one level above so it can be defined for all stacks pages.

I'm not quite sure this is really the best place to have this. I'm open for feedback.

### DIFF
--- a/app/views/shipit/stacks/_links.html.erb
+++ b/app/views/shipit/stacks/_links.html.erb
@@ -1,1 +1,3 @@
 <%# Placeholder to be used by the host application %>
+
+<%= include_plugins(@stack) %>
--- a/app/views/shipit/tasks/_task_output.html.erb
+++ b/app/views/shipit/tasks/_task_output.html.erb
@@ -1,7 +1,6 @@
 <% content_for :main_classes do %>no-footer<% end %>
 
 <%= javascript_include_tag 'task' %>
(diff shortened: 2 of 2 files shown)


casperisfine:
The problem is that so far the plugins expect to be loaded on the task output page. They might break if loaded elsewhere.

viniciusgama:
I agree with you here. In our case it is just 2 plugins and we haven't had any issue as the plugin knows where to render itself.

In that case, do you have any suggestion on a better/safer approach for accomplishing the same?

casperisfine:
I think you should be able to customize this template in your host app, and basically do the same.

viniciusgama:
> I think you should be able to customize this template in your host app, and basically do the same.

it would mean redefine the entire view right? If so, Iit would also mean keeping a constant eye on the changes on that view to keep it updated. That sounds a bit risky for me. 

What about an extra config to enable the plugins at the stack level? Would you be willing to accept something like that?

PR #1132, opened 2021-01-12
Retry task failures in addition to timeouts.

gigr:
This is a refactor of #1109 to include deploy failures. It is technically a breaking change since the `shipit.yml` syntax must change to:

```
deploy:
  retries: 2
rollback:
  retries: 2
 ```

As a follow-up (if necessary) we can allow configuring the types of "failures" that need retries, but right now this is hardcoded.

### DIFF
--- a/app/models/shipit/deploy_spec/file_system.rb
+++ b/app/models/shipit/deploy_spec/file_system.rb
@@ -70,11 +70,11 @@ module Shipit
             'variables' => deploy_variables.map(&:to_h),
             'max_commits' => maximum_commits_per_deploy,
             'interval' => pause_between_deploys,
-            'retries_on_timeout' => retries_on_deploy_timeout,
+            'retries' => retries_on_deploy,
           },
           'rollback' => {
             'override' => rollback_steps,
-            'retries_on_timeout' => retries_on_rollback_timeout,
+            'retries' => retries_on_rollback,
           },
           'fetch' => fetch_deployed_revision_steps,
           'tasks' => cacheable_tasks,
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: app/models/shipit/deploy_spec/file_system.rb [hunk 1]
```diff
@@ -70,11 +70,11 @@ def cacheable_config
             'variables' => deploy_variables.map(&:to_h),
             'max_commits' => maximum_commits_per_deploy,
             'interval' => pause_between_deploys,
-            'retries_on_timeout' => retries_on_deploy_timeout,
```
ajshepley, line 73: Since it's a breaking yml config change, we'll want to reach out to or create PRs for the projects that have the current definition.

https://github.com/search?q=org%3AShopify+retries_on_timeout&type=code

Alternatively, we could keep supporting the old flag alongside the new.

app/models/shipit/deploy_spec/file_system.rb
```diff
@@ -70,11 +70,11 @@ def cacheable_config
...
+            'retries' => retries_on_deploy,
```
ajshepley, line 73: I liked the idea of being able to specify the retries for timeouts and failures and errors separately, but we can leave that idea for later and if we see that it's needed.

ajshepley (approved):

app/models/shipit/deploy_spec/file_system.rb
gigr: The list is small enough that I think I'll PR changes to each to use the new flag.

suprnova32 (approved):
:shipit:

ajshepley (approved):

PR #1133, opened 2021-02-08
Use a redis cache instead of the output_chunks table

suprnova32:
Use a redis cache instead of the `output_chunks` table to store the output of a task as it's being performed.

This PR replaces the use of the `chunks` relation within `Shipit::Task` with a Redis cache where the intermediate output of a task will be stored. By using `append` and `getrange` most of the existing implementation around getting the tail of the output can be kept, keeping the scope of the changes small, and to the point.

Thanks @casperisfine for the suggestion.

I will submit another PR to actually get rid of the `Shipit::OutputChunks` model and its table. 

Closes https://github.com/Shopify/shipit/issues/1366


app/models/shipit/task.rb
```diff
@@ -201,16 +199,16 @@ def run_now!
 
     def write(text)
       log_output(text)
-      chunks.create!(text: text)
+      Shipit.redis.append(output_key, text)
```
casperisfine, line 202: I'm tempted to set an expiry on that key, just in case something goes wrong. But I'm the same time I'm worried some users might not have output rollup setup properly.

Not too sure what to do.

benlangfeld:
Would you mind clarifying the motivation for this change? The ticket you reference is internal to Shopify. Us external users would like to understand why this is happening.

casperisfine:
> The ticket you reference is internal to Shopify.

Apologies for that.

The gist of the issue is that the current implementation tend to put heavy pressure on our MySQL instance. We think Redis would handle this much better.

benlangfeld:
> > The ticket you reference is internal to Shopify.
> 
> Apologies for that.
> 
> The gist of the issue is that the current implementation tend to put heavy pressure on our MySQL instance. We think Redis would handle this much better.

Thanks for clarifying @casperisfine . I suspected as much, but wanted to be sure.

app/models/shipit/task.rb
```diff
@@ -402,6 +404,10 @@ def retries_configured?
       !max_retries.nil? && max_retries > 0
     end
 
+    def output_key
```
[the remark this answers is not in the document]
suprnova32: You're right. I only use it in the tests, so I can make it private.

app/helpers/shipit/chunks_helper.rb
```diff
@@ -3,7 +3,7 @@ module Shipit
   module ChunksHelper
     def next_chunks_url(task)
       return if task.finished?
-      tail_stack_task_path(task.stack, task, last_id: task.chunks.last)
+      tail_stack_task_path(task.stack, task, last_id: task.chunk_output&.bytesize || 0)
```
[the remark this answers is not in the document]
suprnova32: Good idea. Done.

app/models/shipit/task.rb
casperisfine: So to answer myself:

  - Currently tasks are eligible for rollup once they are completed and older than 1 hour.
  - The rollup is supposed to be triggered by and hourly cron.

Somehow I don't trust Redis much to keep this data at rest, so I'm thinking we should enqueue a rollup job once the task complete, to make sure we don't keep too much data only in Redis.

suprnova32: My initial proposal did include triggering a rollup right after the task finished, but once we decided on the this specific approach, it seemed to me that the hourly cron job plus the eligibility threshold should be enough. Redis should be resilient enough to keep the output for a couple of hours at most.

But it makes sense to be extra sure, and roll the task up after it finishes 👍

casperisfine (approved):

ajshepley (approved):
Might be worth testing / 🎩 locally before shipping.

gigr (approved):

casperisfine:
Hum, a bit late, but this is gonna be tricky to rollout as is. As when this change will be deployed, a bunch of tasks will be in flight, or waiting to be rolled up.

The in flight ones will end up truncated, and the ones waiting to be rolled up will lose their logs.

We should probably make an extra change to the rollup task to consider both sources of logs, at least for one version.

suprnova32:
@casperisfine I had something in mind for `Shopify/shipit` but you are right, we should take care of that here as well. I'll create a new PR that changes the rollup job. It should look for data in both sources and combine them. 

Then once those changes have been deployed, the next PR that removes the `OutputChunks` code, should also take care of reverting the changes to the rollup job.

Thanks for pointing that out.

PR #1135, opened 2021-02-10
Take into account existing output_chunks when rolling up

suprnova32:
With the changes introduced in #1133 existing output chunks would not be taken into account when rolling up the output of a task. This meant that task that were in flight as the new code was deployed would lose part of their output.

These changes to the rollup code make sure that existing data in the `output_chunks` table is taken into account, and inserted before the redis data.

### DIFF
--- a/app/models/shipit/task.rb
+++ b/app/models/shipit/task.rb
@@ -238,9 +238,14 @@ module Shipit
 
     def rollup_chunks
       ActiveRecord::Base.transaction do
-        self.output = chunk_output
+        chunks = Shipit::OutputChunk.where(task: self).pluck(:text)
+        chunks << chunk_output
+        self.output = chunks.join("\n")
+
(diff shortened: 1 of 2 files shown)


app/models/shipit/task.rb
```diff
@@ -238,9 +238,14 @@ def schedule_rollup_chunks
 
     def rollup_chunks
       ActiveRecord::Base.transaction do
-        self.output = chunk_output
+        old_chunks = Shipit::OutputChunk.where(task: self).pluck(:text).join
+        self.output =
+          old_chunks.empty? ? chunk_output : [old_chunks, chunk_output].join("\n")
```
casperisfine, line 243:
```suggestion
        chunks = Shipit::OutputChunk.where(task: self).pluck(:text).join
        chunks << chunk_output
        self.output = chunks
```

casperisfine (approved):

PR #1136, opened 2021-02-10
Fix String coercion error

suprnova32:
Convert the `last_byte` param to integer to avoid a string coercion error

https://app.bugsnag.com/shopify/shipit/errors/60245d4e06dadd0017d851d9?event_id=60245d5f0069a9de55710000&i=sk&m=fq

### DIFF
--- a/app/serializers/shipit/tail_task_serializer.rb
+++ b/app/serializers/shipit/tail_task_serializer.rb
@@ -33,7 +33,7 @@ module Shipit
     delegate :stack, to: :object
 
     def offset
-      (output&.bytesize || 0) + (context[:last_byte] || 0)
+      (output&.bytesize || 0) + (context[:last_byte]&.to_i || 0)
     end
 
     def rollback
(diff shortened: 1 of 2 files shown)


gigr (approved):

ajshepley (approved):

rdji123 (approved):

app/serializers/shipit/tail_task_serializer.rb
```diff
@@ -33,7 +33,7 @@ def include_rollback_url?
     delegate :stack, to: :object
 
     def offset
-      (output&.bytesize || 0) + (context[:last_byte] || 0)
+      (output&.bytesize || 0) + (context[:last_byte]&.to_i || 0)
```
casperisfine, line 36: that ` || 0` is no longer useful then

suprnova32: There is still a chance that `context[:last_byte]` is `nil`. In that case we do need the `|| 0`, otherwise we end up with a "No implicit conversion of `nil` to integer" error.

casperisfine: Either way I think the proper fix would have bene to do the casting in the controller. I might push a tiny refactor on master.

casperisfine: c25a37c7678cf0a9250e4cd9b15211ed779d5c48

PR #1137, opened 2021-02-11
Bump version to 0.33.0

suprnova32:
Bump the Gem version to 0.33.0 to officially release the Redis cache changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Unreleased
+# 0.33.0
 
 * Use a Redis cache to store task output, instead of `output_chunks` (deprecate their use).
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    shipit-engine (0.32.0)
+    shipit-engine (0.33.0)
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-# Unreleased
+# 0.33.0
```
benlangfeld, line 1: There are many other changes in this version.

suprnova32: You are correct. None of the other commits made an entry to `CHANGELOG` and I forgot to check how many commits have been made since 0.32.0. I'll have a look and add them here.

casperisfine: Yeah, usually I update the CHANGELOG before making a release. 

I can do it tomorrow, unless you want to take a stab at it.

suprnova32: @casperisfine thanks, that would be great! you are intimately more familiar with the codebase, and were there to review all the changes as well. It will take you a lot less time, than it would take me.

casperisfine:
Closing in favor of https://github.com/Shopify/shipit-engine/commit/1dc6620730c813d5c833976cffee18a319c13e41

PR #1138, opened 2021-02-13
Add new mode param to allow dark and light mode on merge status endpoint

kfinn-clio:
Hi  from Clio!! 👋🏻   
GitHub recently released the ability to [toggle between different themes](https://github.com/settings/appearance).
This PR adds a data attribute to the html body in order to use the dark or light mode CSS. The value of the data attribute is passed in through the new `mode` parameter. If the user passes in an invalid value for example `darks` or `light` the light mode CSS is loaded.

### Why
I looked at the GitHub HTML and CSS and they define a data attribute `data-color-mode=dark` or `data-color-mode=light` to switch the CSS to dark or light mode. Adding this data attribute to the body of this HTML document allows the ability to toggle between light and dark mode.

## Before
Screen Shot 2021-02-12 at 9 30 38 AM

## After 
Screen Shot 2021-02-12 at 1 03 25 PM

### DIFF
--- a/app/views/layouts/merge_status.html.erb
+++ b/app/views/layouts/merge_status.html.erb
@@ -4,7 +4,7 @@
     <%= stylesheet_link_tag *(params[:stylesheets] || []).select { |url| url.start_with?('https://assets-cdn.github.com/', 'https://github.githubassets.com/') } %>
     <%= stylesheet_link_tag 'merge_status' %>
   </head>
-  <body>
+  <body data-color-mode="<%= params[:mode] %>">
     <div class="merge-status-container" data-layout-content><%= yield %></div>
     <%= javascript_include_tag 'merge_status' %>
   </body>


app/views/layouts/merge_status.html.erb
```diff
@@ -4,7 +4,7 @@
     <%= stylesheet_link_tag *(params[:stylesheets] || []).select { |url| url.start_with?('https://assets-cdn.github.com/', 'https://github.githubassets.com/') } %>
     <%= stylesheet_link_tag 'merge_status' %>
   </head>
-  <body>
+  <body data-color-mode="<%= params[:mode] %>">
```
casperisfine, line 7: How does it behave if this is empty?

casperisfine (approved):
Just 1 question but LGTM.

app/views/layouts/merge_status.html.erb
kfinn-clio: When `params[:mode]` is empty or missing from the params list the body will default to light mode so no changes to the current user experience.

casperisfine:
Thanks!

PR #1139, opened 2021-02-15
Replace outdated ActiveModel::Serializers by Panko::Serializer

casperisfine:
Most of the code was super easy to convert.

So far I've found one blocker:

  - There's no way to conditionally include attributes. Best we can do is to have a `null` value. 

I'll see if I can get that feature added upstream.


casperisfine:
> * There's no way to conditionally include attributes. Best we can do is to have a `null` value.

Upstream accepted the patch very quickly. There is still a minor limitation as it only work for regular attributes, relations can't be skipped. But I think it's probably fine. It just means a couple of API fields will be explicitly `null` rather than being skipped. I can live with that.

Overall the integration was super easy, I'm just a bit worried because the test coverage of the serializers is relatively limited, so It's not unlikely I missed some difference between the two serailizers.

Issue #1140, opened 2021-02-18
Does not support non-v prefixed tags


larouxn:
Currently the deploy system assumes usage of `v` prefixed tags such as `v0.9.0` and creates `v` prefixed tags on deploy which leaves repos with non-`v` prefixed tags with extra, unused tags sitting around. (see screenshots below)

The code that checks for tags and proceeds with creating tags on deploy is here.
https://github.com/Shopify/shipit-engine/blob/a4899ed60a23c2d05d74ede3a683f82907d4232f/lib/snippets/release-gem#L18-L33

<details>
  <summary>Example screenshots</summary>
  <img width="666" alt="Screen Shot 2021-02-18 at 17 17 39" src="https://user-images.githubusercontent.com/1557529/108327498-d4c94180-720e-11eb-9106-9b710031f9e0.png">
  <img width="943" alt="Screen Shot 2021-02-18 at 17 18 11" src="https://user-images.githubusercontent.com/1557529/108327508-d6930500-720e-11eb-9119-1950d6e7ef3b.png">
  <img width="1233" alt="Screen Shot 2021-02-18 at 17 07 44" src="https://user-images.githubusercontent.com/1557529/108327509-d72b9b80-720e-11eb-9902-6f82dfb1dba6.png">
</details>

Note, if only supporting `v` prefixed tags is an intentional decision, that's unideal but understandable for standardization and simplicity's sake. Please let me know. :)

casperisfine:
> if only supporting v prefixed tags is an intentional decision

It is, that's a convention that comes from the `rake release` task provided by bundler. Shipit assumes your gem follow this convention.

I'd recommend to change your gem to follow this convention.

larouxn:
> It is, that's a convention that comes from the rake release task provided by bundler. Shipit assumes your gem follow this convention.

Understood. Wasn't aware of this convention. Will use `v` prefixed tags going forward.

PR #1141, opened 2021-02-18
Support non-v prefixed tags

larouxn:
Check for both v prefixed and non-v prefixed tags in `tag_exists?` so we avoid creating unnecessary v prefixed tags during deploys, leaving repos with unused tags like so.
Screen Shot 2021-02-18 at 17 07 44

### Minor backwards compatability 🎩 
Screen Shot 2021-02-18 at 17 57 59

Resolves https://github.com/Shopify/shipit-engine/issues/1140

### DIFF
--- a/lib/snippets/release-gem
+++ b/lib/snippets/release-gem
@@ -24,7 +24,8 @@ module Git
   end
 
   def tag_exists?(version)
-    system('git', 'rev-parse', '--verify', "v#{version}^{commit}", out: File::NULL, err: File::NULL)
+    system('git', 'rev-parse', '--verify', "v#{version}^{commit}", out: File::NULL, err: File::NULL) ||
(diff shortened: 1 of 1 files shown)


larouxn:
`Tests with MySQL (2.6.x)` CI containers seems to be experiencing an unrelated infrastructure level failure.

```
LoadError: Error loading the 'mysql2' Active Record adapter. Missing a gem it depends on? libmysqlclient.so.20: cannot open shared object file: No such file or directory - /home/runner/work/shipit-engine/shipit-engine/vendor/bundle/ruby/2.6.0/gems/mysql2-0.5.3/lib/mysql2/mysql2.so
```

larouxn:
Closing re: https://github.com/Shopify/shipit-engine/issues/1140#issuecomment-781205362.

PR #1142, opened 2021-02-19
Revert gemoji gem updates.

ajshepley:
Changes in 3.0 resulted in this bug:

```ruby
ActionView::Template::Errorshipit/stacks#show
The asset "emoji/unicode/1f516.png" is not present in the asset pipeline
```

 as the emojis are no longer included in the gem.

This led to some stacks not being openable if they had a commit message containing such an emoji.


suprnova32 (approved):

gigr (approved):

ajshepley:
>LoadError: Error loading the 'mysql2' Active Record adapter. Missing a gem it depends on? libmysqlclient.so.20: cannot open shared object file: No such file or directory - /home/runner/work/shipit-engine/shipit-engine/vendor/bundle/ruby/2.6.0/gems/mysql2-0.5.3/lib/mysql2/mysql2.so

Flakey or... broken?

casperisfine (approved):

casperisfine:
I'll do the proper update later. Basically the gem no longer onboard images, and instead gives an UTF-8 character to use as replacement. My bad for not paying attention enough.

casperisfine:
> Flakey or... broken?

Looks broken, sounds like the base image changed and no longer include mysqllcient, or has it in another location.

I'd say merge for now, we can fix CI in another PR.

PR #1144, opened 2021-03-04
Remove rollback retries tests and functions

rdji123:
Retries on rollback should not be configurable, rollbacks trigger a lock on the stack and are blocked from retries 
https://github.com/Shopify/shipit-engine/blob/0f1e09f2adeac6da01efcd78cd1334481cc54a43/app/models/shipit/task.rb#L407


casperisfine:
@rdji123 please mind that shipit-engine is a public repository with external contributors, please don't link to private issues and provide a publicly understandable rationale.

gigr (approved):

PR #1149, opened 2021-03-25
No longer assume `master` is the default branch

gmcgibbon:
Since `main` is the new preferred default branch name, it seems appropriate to have this preference in shipit as well.

### DIFF
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -90,6 +90,7 @@ module Shipit
     }
     validates :environment, format: { with: /\A[a-z0-9\-_\:]+\z/ }, length: { maximum: ENVIRONMENT_MAX_SIZE }
     validates :deploy_url, format: { with: URI.regexp(%w(http https ssh)) }, allow_blank: true
+    validates :branch, presence: true
 
     validates :lock_reason, length: { maximum: 4096 }
 
@@ -593,7 +594,13 @@ module Shipit
 
     def update_defaults
       self.environment = 'production' if environment.blank?
-      self.branch = 'master' if branch.blank?
+      self.branch = default_branch_name if branch.blank?
+    end
+
+    def default_branch_name
+      Shipit.github.api.repo(github_repo_name).default_branch
+    rescue Octokit::NotFound, Octokit::InvalidRepository
+      nil
     end
 
     def set_locked_since
--- a/app/views/shipit/stacks/new.html.erb
+++ b/app/views/shipit/stacks/new.html.erb
@@ -17,7 +17,7 @@
         </div>
         <div class="field-wrapper">
           <%= f.label :branch %>
-          <%= f.text_field :branch, placeholder: 'master' %>
+          <%= f.text_field :branch, placeholder: '<default branch>' %>
         </div>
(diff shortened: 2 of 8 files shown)


casperisfine:
As suggested by @etiennebarrie it would be better to allow to leave it empty, and to query GitHub to get the name of the default branch.

casperisfine:
> to query GitHub to get the name of the default branch.

If you don't feel like doing that last part, then you can also just make it so there's no default.

I'd rather have to write `master/main` explicitly every time than to get bitten by some default that don't match mine.

gmcgibbon:
> As suggested by @etiennebarrie it would be better to allow to leave it empty, and to query GitHub to get the name of the default branch.

We still need to allow specification of a branch name in case users have staging or something on another branch. I can refactor to use the API to populate the placeholder though. I'll try that.

gmcgibbon:
> If you don't feel like doing that last part, then you can also just make it so there's no default.
> 
> I'd rather have to write master/main explicitly every time than to get bitten by some default that don't match mine.

Yeah it depends on the tooling that already exists to make this easy. I'll make it blank and remove the column default as a fallback. 👍

casperisfine:
> We still need to allow specification of a branch name in case users have staging or something on another branch.

I meant if you leave it blank then Shipit set it to the default branch set in GitHub, but still have a field so you can pout whatever you want there.

app/models/shipit/stack.rb
```diff
@@ -593,7 +593,13 @@ def clear_cache
...
     def update_defaults
       self.environment = 'production' if environment.blank?
-      self.branch = 'master' if branch.blank?
+      self.branch = default_branch_name if branch.blank?
+    end
+
+    def default_branch_name
+      Shipit.github.api.repo(github_repo_name).default_branch
+    rescue Octokit::NotFound, Octokit::InvalidRepository
+      'main'
```
casperisfine, line 602: I think we're better to display a validation error or something like that.

Maybe just return `nil` so that the presence validator fails.

lib/shipit/command.rb
```diff
@@ -13,7 +13,7 @@ class Command
     Denied = Class.new(Error)
     TimedOut = Class.new(Error)
 
-    BASE_ENV = Bundler.clean_env.merge((ENV.keys - Bundler.clean_env.keys).map { |k| [k, nil] }.to_h)
+    BASE_ENV = Bundler.unbundled_env.merge((ENV.keys - Bundler.unbundled_env.keys).map { |k| [k, nil] }.to_h)
```
casperisfine, line 16: Please don't bundle other unrelated changes. I'm aware of this warnings, but if you want to fix it it's better to do it in it's own PR, as it's not necessarily as harmless as it might seem.

app/models/shipit/stack.rb
gmcgibbon: There's no presence validation, but I can add one.

lib/shipit/command.rb
gmcgibbon: It is in a separate commit, so it is easy to move around. Will PR separately.

PR #1151, opened 2021-03-30
Support multiple GH organisations

MrMarvin:
## Summary

A Github application can only authenticate to the organization its installed in. This is a problem if one wishes to deploy code from multiple Github organizations using a single Shipit instance. This pull requests adds the capability to configure multiple Github applications in the `github` section of `config/secrets.yml`. It also preserves the current schema for backwards compatibility support.

### New `config/secrets.yml` Example

```yml
github:
  somegithuborg:
    app_id:
    installation_id:
    webhook_secret: # nil
    private_key:
    oauth:
      id:
      secret:
      teams: # Optional
  someothergithuborg:
    app_id:
    installation_id:
    webhook_secret: # nil
    private_key:
    oauth:
      id:
      secret:
      teams: # Optional
```

### DIFF
--- a/app/controllers/shipit/stacks_controller.rb
+++ b/app/controllers/shipit/stacks_controller.rb
@@ -154,7 +154,7 @@ module Shipit
     end
 
     def create_params
-      params.require(:stack).permit(:environment, :branch, :deploy_url, :ignore_ci)
+      params.require(:stack).permit(:repo_owner, :repo_name, :environment, :branch, :deploy_url, :ignore_ci)
     end
 
     def update_params
@@ -180,7 +180,7 @@ module Shipit
     end
 
     def repository_params
-      params.require(:stack).permit(:repo_owner, :repo_name)
+      params.require(:stack).permit(:repo_owner, :repo_name, :branch, :environment, :deploy_url, :ignore_ci)
     end
   end
 end
--- a/app/controllers/shipit/webhooks_controller.rb
+++ b/app/controllers/shipit/webhooks_controller.rb
@@ -16,7 +16,8 @@ module Shipit
     private
 
     def verify_signature
-      head(422) unless Shipit.github.verify_webhook_signature(request.headers['X-Hub-Signature'], request.raw_post)
+      github_api = Shipit.github(organization: @stack.owner)
+      head(422) unless github_api.verify_webhook_signature(request.headers['X-Hub-Signature'], request.raw_post)
     end
 
     def check_if_ping
--- a/app/controllers/shipit/webhooks_controller.rb
+++ b/app/controllers/shipit/webhooks_controller.rb
@@ -16,7 +16,7 @@ module Shipit
     private
 
     def verify_signature
-      github_app = @stack&.github_app || Shipit.github # TODO: figure out something cleaner
+      github_app = stack&.github_app
       verified = github_app.verify_webhook_signature(
         request.headers['X-Hub-Signature'],
         request.raw_post
--- a/app/controllers/shipit/webhooks_controller.rb
+++ b/app/controllers/shipit/webhooks_controller.rb
@@ -16,12 +16,11 @@ module Shipit
     private
 
     def verify_signature
-      verified = Shipit.github_organizations.map do |gh_org|
-        Shipit.github(organization: gh_org).verify_webhook_signature(
-          request.headers['X-Hub-Signature'],
-          request.raw_post
-        )
-      end.any?
+      github_app = @stack&.repository&.github_app_instance || Shipit.github # TODO: figure out something cleaner
+      verified = github_app.verify_webhook_signature(
+        request.headers['X-Hub-Signature'],
+        request.raw_post
+      )
       head(422) unless verified
     end
 
--- a/app/controllers/shipit/webhooks_controller.rb
+++ b/app/controllers/shipit/webhooks_controller.rb
@@ -16,8 +16,13 @@ module Shipit
     private
 
     def verify_signature
-      github_api = Shipit.github(organization: @stack.owner)
-      head(422) unless github_api.verify_webhook_signature(request.headers['X-Hub-Signature'], request.raw_post)
+      verified = Shipit.github_organizations.map do |gh_org|
+        Shipit.github(organization: gh_org).verify_webhook_signature(
+          request.headers['X-Hub-Signature'],
+          request.raw_post
+        )
+      end.any?
+      head(422) unless verified
     end
 
     def check_if_ping
--- a/app/controllers/shipit/webhooks_controller.rb
+++ b/app/controllers/shipit/webhooks_controller.rb
@@ -16,7 +16,7 @@ module Shipit
     private
 
     def verify_signature
-      github_app = @stack&.repository&.github_app_instance || Shipit.github # TODO: figure out something cleaner
+      github_app = @stack&.github_app || Shipit.github # TODO: figure out something cleaner
       verified = github_app.verify_webhook_signature(
         request.headers['X-Hub-Signature'],
         request.raw_post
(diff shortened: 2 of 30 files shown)


lib/shipit.rb
```diff
@@ -101,8 +103,19 @@ def redis
...
-    @github ||= GitHubApp.new(secrets.github)
+  def github(organization: github_default_organization)
+    org = organization.to_sym
+    raise GithubOrganizationUnknown if secrets.github[org].nil?
+    @github ||= {}
+    @github[organization] ||= GitHubApp.new(organization, secrets.github[org])
+  end
+
+  def github_default_organization
+    secrets.github.keys.first
```
MrMarvin, line 114: This requires the current structure of the secrets to be slightly remodelled, right?
Looking at [the development secrets example above](https://github.com/Shopify/shipit-engine/pull/1151/files#diff-d077374229ea196a27874f1d46a8c013dce1af4ac5ac7c4ca0ba74bc09f2312aR5-R22), the default app needs to be nested one level deeper now, I think

Ginja: New config can now look like this. However, this branch now supports backward capability for the old-style config.

```yaml
github:
    org1:
      app_id: 
      installation_id:
      webhook_secret:
      private_key:
      oauth:
        id: 
        secret: 
        teams:
    org2:
      app_id: 
      installation_id:
      webhook_secret:
      private_key: 
      oauth:
        id: 
        secret: 
        teams:
```

app/controllers/shipit/stacks_controller.rb
```diff
@@ -180,7 +180,7 @@ def repo_name
     end
 
     def repository_params
-      params.require(:stack).permit(:repo_owner, :repo_name)
+      params.require(:stack).permit(:repo_owner, :repo_name, :branch, :environment, :deploy_url, :ignore_ci)
```
Ginja, line 183: This was updated to prevent the following warning in the logs. Can be easily reverted if necessary/desired.

```
Unpermitted parameters: :branch, :environment, :deploy_url, :ignore_ci
```

app/jobs/shipit/github_sync_job.rb
```diff
@@ -32,7 +32,8 @@ def append_commit(gh_commit)
 
     def fetch_missing_commits(&block)
       commits = []
-      iterator = Shipit::FirstParentCommitsIterator.new(&block)
+      github_api = @stack&.repository&.github_api || Shipit.github.api # TODO: figure out something cleaner
```
gigr, line 35: Is the `||` necessary here? Looks like `@stack` can never be `nil`. Stacks should also have a repository associated with them in all cases.

config/secrets.development.example.yml
```diff
@@ -1,8 +1,10 @@
 host: 'localhost:3000'
 redis_url: 'redis://127.0.0.1:6379/0'
 
+# TODO: document creating a dev app
```
gigr, line 4: https://github.com/Shopify/shipit-engine/blob/master/docs/setup.md#creating-the-github-app should suffice?

gigr:
@byroot may want to take a look as well.

app/controllers/shipit/webhooks_controller.rb
```diff
@@ -16,7 +16,13 @@ def create
     private
 
     def verify_signature
-      head(422) unless Shipit.github.verify_webhook_signature(request.headers['X-Hub-Signature'], request.raw_post)
+      verified = Shipit.github_organizations.map do |gh_org|
```
gigr, line 19: Github webhooks should include the `repo_owner` (org) field. Did you consider using that instead of iterating over available organizations?

app/jobs/shipit/github_sync_job.rb
Ginja: This is purely for making a particular test pass. We've done something similar in other parts of the code (search for the comment).

config/secrets.development.example.yml
Ginja: Possibly! I carried it over from the old config. I'll remove it.

Ginja: Done :)

app/controllers/shipit/webhooks_controller.rb
Ginja: Good catch! I've gone ahead and changed this. I followed the same approach as above in `github_sync_job.rb`. The tests for `webhooks_controller.rb` don't define a stack. You are absolutely right though. In real flows one should always be defined. I don't know if this is the PR that we should clean up/fix those tests though.

app/controllers/shipit/webhooks_controller.rb
```diff
@@ -16,7 +16,12 @@ def create
     private
 
     def verify_signature
-      head(422) unless Shipit.github.verify_webhook_signature(request.headers['X-Hub-Signature'], request.raw_post)
+      github_app = @stack&.repository&.github_app_instance || Shipit.github # TODO: figure out something cleaner
```
casperisfine, line 19: Isn't `@stack` always present?

app/models/shipit/commit.rb
```diff
@@ -143,7 +143,7 @@ def schedule_refresh_check_runs!
 
     def refresh_statuses!
       github_statuses = stack.handle_github_redirections do
-        Shipit.github.api.statuses(github_repo_name, sha, per_page: 100)
+        stack.repository.github_api.statuses(github_repo_name, sha, per_page: 100)
```
casperisfine, line 146: `Stack` should delegate `github_api` to it's `repository` to avoid these long chains everywhere.

app/models/shipit/repository.rb
```diff
@@ -66,16 +66,24 @@ def github_repo_name
       [owner, name].join('/')
     end
 
+    def github_api
+      github_app_instance.api
+    end
+
+    def github_app_instance
```
casperisfine, line 73: The `_instance` suffix is ugly. Why is it needed?

casperisfine (changes requested):

app/models/shipit/repository.rb
Ginja: It's not :) No issues in removing it.

app/models/shipit/commit.rb
Ginja: 👍

app/jobs/shipit/github_sync_job.rb
```diff
@@ -32,7 +32,8 @@ def append_commit(gh_commit)
 
     def fetch_missing_commits(&block)
       commits = []
-      iterator = Shipit::FirstParentCommitsIterator.new(&block)
+      github_api = @stack&.github_api || Shipit.github.api # TODO: figure out something cleaner
```
casperisfine, line 35: Same here.

lib/shipit.rb
```diff
@@ -115,6 +138,7 @@ def user
     if github.bot_login
       User.find_or_create_by_login!(github.bot_login)
     else
+      # TODO: Anything needed here?
```
casperisfine, line 141: Nothing needed here.

casperisfine (commented):
Looks good to me now. It's just a matter of clearing the remaining TODOs

app/controllers/shipit/webhooks_controller.rb
```diff
@@ -16,7 +16,12 @@ def create
     private
 
     def verify_signature
-      head(422) unless Shipit.github.verify_webhook_signature(request.headers['X-Hub-Signature'], request.raw_post)
+      github_app = @stack&.github_app || Shipit.github # TODO: figure out something cleaner
```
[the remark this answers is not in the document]
Ginja: Same reason as above. I’d need to define a stack in a test that doesn’t necessarily require it for what it’s attempting to test. Should be able to mock it though.

https://github.com/Shopify/shipit-engine/pull/1151#discussion_r605967973

PR #1152, opened 2021-03-30
Fix bundler deprecation

gmcgibbon:
Fixes ```[DEPRECATED] `Bundler.clean_env` has been deprecated in favor of `Bundler.unbundled_env`. If you instead want the environment before bundler was originally loaded, use `Bundler.original_env` (called at /Users/gannon/src/github.com/Shopify/shipit-engine/lib/shipit/command.rb:16)``` warnings.

### DIFF
--- a/lib/shipit/command.rb
+++ b/lib/shipit/command.rb
@@ -13,7 +13,7 @@ module Shipit
     Denied = Class.new(Error)
     TimedOut = Class.new(Error)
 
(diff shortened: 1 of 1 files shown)


airhorns:
This'd be really great to get in!

casperisfine:
Ah right sorry, It fell thought the cracks.

PR #1154, opened 2021-04-07
Force standardization of github config keys

Ginja:
`Rails.application.secrets` can return either Symbols or Strings depending on how it's loaded and parsed. For the purpose of ease of evaluation, we'll force all keys to be symbols.

### DIFF
--- a/lib/shipit.rb
+++ b/lib/shipit.rb
@@ -131,7 +131,7 @@ module Shipit
 
   def legacy_github_api
     if secrets&.github_api.present?
-      @legacy_github_api ||= github.new_client(access_token: secrets.github_api['access_token'])
+      @legacy_github_api ||= github.new_client(access_token: secrets.github_api[:access_token])
     end
   end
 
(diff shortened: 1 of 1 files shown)


lib/shipit.rb
```diff
@@ -120,7 +120,7 @@ def github(organization: github_default_organization)
 
   def github_default_organization
     return nil unless secrets&.github
-    org = secrets.github.keys.first
+    org = secrets.github.keys.first.to_s
```
gigr, line 123: There are other places where the secrets are accessed, e.g. https://github.com/Shopify/shipit-engine/blob/1e90347cfd47593db21e84626d4864e5cd7c2e3e/lib/shipit.rb#L107-L119 and https://github.com/Shopify/shipit-engine/blob/1e90347cfd47593db21e84626d4864e5cd7c2e3e/lib/shipit.rb#L129. Did you verify that these don't need to be string/symbol based keys?

casperisfine (approved):

PR #1162, opened 2021-04-19
Drop unhandled events for Webhooks Controller 

Ginja:
Some payloads from Github don't send the `repository`, or `organization` sub-object. This adds another `before_action` method to drop events we don't have a handler for.

### DIFF
--- a/app/controllers/shipit/webhooks_controller.rb
+++ b/app/controllers/shipit/webhooks_controller.rb
@@ -2,7 +2,7 @@
 module Shipit
   class WebhooksController < ActionController::Base
     skip_before_action :verify_authenticity_token, raise: false
-    before_action :check_if_ping, :verify_signature
+    before_action :check_if_ping, :drop_unhandled_event, :verify_signature
 
     respond_to :json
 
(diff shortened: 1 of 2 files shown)


app/controllers/shipit/webhooks_controller.rb
```diff
@@ -43,7 +43,20 @@ def event
 
     def repository_owner
       # Fallback to the organization sub-object if repository isn't included in the payload
-      params.dig('repository', 'owner', 'login') || params.dig('organization', 'login')
+      owner = params.dig('repository', 'owner', 'login') || params.dig('organization', 'login')
+
+      if owner.nil?
+        # Fallback to less common locations
+        if params.dig('installation', 'account', 'type') == 'Organization'
```
casperisfine, line 50: Do we even care about these events?

Couldn't we refactor a bit to first check wether the event has a handler before going down that route?

Ginja: Agree on the need for a refactor. I don't want to tackle that in this PR though.

I like what you proposed @ajshepley. I went with that approach, and also dropped the head response for if `repository_owner` cannot be found. It didn't work as expected, and now for events that we do support we want to be able to handle them successfully (i.e. detecting the organization so we can choose the right Github app).

Ginja: > I can see us moving the repository_owner parsing as a trait of Shipit::Webhooks::Handler or something like that, so that we don't build an endlessly bigger method here, but I don't know if that'd be necessary right now.

:100: The handler in the future should be authoritative for parsing its payload and determining the organization as the payload structure will be known.

ajshepley (approved):

gigr (approved):

PR #1164, opened 2021-04-21
Use a negative `status:delay` config value to skip automatic `MarkDeployHealthyJob` execution

suprnova32:
Use a negative `status:delay` config value to skip automatic `MarkDeployHealthyJob` execution

With this, by setting the delay value to `-1`, projects can configure a validation period that does not expire, and that has to be marked as healthy either via the API or by a human.

Related to https://github.com/Shopify/deploys/issues/48

### DIFF
--- a/app/models/shipit/deploy.rb
+++ b/app/models/shipit/deploy.rb
@@ -209,7 +209,7 @@ module Shipit
     end
 
     def report_complete!
-      if stack.release_status? && stack.release_status_delay.positive?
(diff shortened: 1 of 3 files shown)


gigr (approved):

gigr:
> Should we update any of the documentation? I don't think it goes this in-depth so maybe just a note in the changelog (later on) will be good enough.

The entire `status` feature isn't documented. I don't think it's necessary in this PR, especially while we're working out the logic.

PR #1167, opened 2021-05-04
Guard against stale checkrun updates.

ajshepley:
### The problem

Shipit parses CheckSuite events and uses them to enqueue [refresh jobs](https://github.com/Shopify/shipit-engine/blob/master/app/jobs/shipit/refresh_check_runs_job.rb#L8). These jobs sync checkruns via [`Commit.refresh_check_runs`](https://github.com/Shopify/shipit-engine/blob/master/app/models/shipit/commit.rb#L171).

We're seeing cases where this job seems to be returning stale data - either that, or we receive the correct data for our checkruns api request, but GitHub then does not send us a new CheckSuite event.

This seems to happen when the CheckSuite webhook event is received within one second of a successful checkrun webhook. Perhaps there's an ordering or timing issue, or some sort of debouncing on GitHub's end that would result in no second checksuite event.

More details are available on our internal issues page.

Some potential solutions to this may involve persisting or retrieving checkruns more often (see `Followups`) so guarding against outdated information would be helpful.

### This PR / The Change

This PR changes the Checkrun class to compare an updated checkrun against the db record before we update the record.

We now compare the timestamps between the new GitHub checkrun data and our db record, and:

* If the github data is a new checkrun, we persist it.
* If it's a known checkrun, but has a newer github timestamp than the db record, then we update the record with the new one's data.
* If the new checkrun has a different timestamp but the same conclusion, we ignore it (keeping the Idempotence we have today).
* If the new checkrun has a different conclusion, but has an older (or equal) timestamp, we enqueue a new refresh for a little bit later.
  * Unfortunately, GitHub's timestamp only has seconds-resolution. I think enqueuing a refresh makes sense here. If GitHub has somehow permanently reverted a timestamp, then this could be problematic. But I don't think we've ever seen that before.

### Followups, mergeability

This PR is a draft, as it's not a change that makes sense without deciding to move forward with some other changes, such as handling CheckRun events and persisting a checkrun from those events (like we do with [Statuses](https://github.com/Shopify/shipit-engine/blob/master/app/models/shipit/webhooks/handlers/status_handler.rb#L21)). I'll prepare a PR to demonstrate those ideas, and point it at this branch for simplicity of review.

### DIFF
--- a/app/models/shipit/check_run.rb
+++ b/app/models/shipit/check_run.rb
@@ -5,6 +5,8 @@ module Shipit
     include DeferredTouch
     include Status::Common
 
+    CHECK_RUN_REFRESH_DELAY = 5.seconds
+
     belongs_to :stack, required: true
     belongs_to :commit, required: true
 
@@ -19,11 +21,28 @@ module Shipit
         create!(selector.merge(attributes))
       rescue ActiveRecord::RecordNotUnique
         record = find_by!(selector)
-        record.update!(attributes)
+
+        if record.github_updated_at < attributes[:github_updated_at]
+          record.update!(attributes)
+        elsif attributes[:conclusion] != record.conclusion
+          Rails.logger.warn(
+            "Conflicting stale checkrun received. Checkrun id: #{selector[:github_id]}, Details: #{attributes}"
+          )
+          RefreshCheckRunsJob.set(wait: CHECK_RUN_REFRESH_DELAY).perform_later(commit_id: record.commit_id)
+        end
+
         record
       end
 
       def create_or_update_from_github!(stack_id, github_check_run)
+        checkrun_date = github_check_run.completed_at&.to_s || github_check_run.started_at&.to_s
+
+        unless checkrun_date
+          Rails.logger.warn("No valid timestamp found in checkrun data. Checkrun id: #{github_check_run.id}.")
+          RefreshCheckRunsJob.set(wait: CHECK_RUN_REFRESH_DELAY).perform_later(stack_id: stack_id)
+          return
+        end
+
         create_or_update_by!(
           selector: {
             github_id: github_check_run.id,
@@ -35,6 +54,7 @@ module Shipit
             title: github_check_run.output.title.to_s.truncate(1_000),
             details_url: github_check_run.details_url,
             html_url: github_check_run.html_url,
+            github_updated_at: Time.parse(checkrun_date),
           },
         )
       end
--- a/db/migrate/20210504200438_add_github_updated_at_to_check_runs.rb
+++ b/db/migrate/20210504200438_add_github_updated_at_to_check_runs.rb
@@ -1,0 +1,5 @@
+class AddGithubUpdatedAtToCheckRuns < ActiveRecord::Migration[6.1]
+  def change
+    add_column :check_runs, :github_updated_at, :datetime, default: 0
+  end
+end
--- a/test/dummy/db/schema.rb
+++ b/test/dummy/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_03_25_194053) do
+ActiveRecord::Schema.define(version: 2021_05_04_200438) do
 
   create_table "api_clients", force: :cascade do |t|
     t.text "permissions", limit: 65535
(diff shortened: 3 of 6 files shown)


db/migrate/20210504200438_add_github_updated_at_to_check_runs.rb
```diff
@@ -0,0 +1,5 @@
+class AddGithubUpdatedAtToCheckRuns < ActiveRecord::Migration[6.1]
+  def change
+    add_column :check_runs, :github_updated_at, :datetime, default: 0
```
gigr, line 3: How precise are Github's timestamps? You can store datetime data down to the microsecond with some tweaking.

ajshepley: GitHub's timestamps are to the second. From their example (other PR): https://github.com/Shopify/shipit-engine/blob/e8b5f82b5ea2689d3bf1c5b6b6d542e6bf9b15b6/test/fixtures/payloads/check_run_success.json#L13

app/models/shipit/check_run.rb
```diff
@@ -19,11 +19,23 @@ def create_or_update_by!(selector:, attributes: {})
...
+          )
+          RefreshCheckRunsJob.set(wait: 5.seconds).perform_later(commit_id: record.commit_id)
+        end
+
         record
       end
 
       def create_or_update_from_github!(stack_id, github_check_run)
+        checkrun_date = github_check_run.completed_at&.to_s || github_check_run.started_at&.to_s
+        parsed_updated_at = checkrun_date ? Time.parse(checkrun_date) : Time.now.utc
```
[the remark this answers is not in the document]
ajshepley: They should always have a started_at, but not always a completed_at. It doesn't hurt to be careful I suppose. Enqueuing a refresh in that case seems like a good idea.

ajshepley:
```ruby
Err:24 https://packages.microsoft.com/ubuntu/20.04/prod focal/main amd64 Packages
  File has unexpected size (69255 != 68824). Mirror sync in progress? [IP: 13.66.21.183 443]
...
Fetched 3913 kB in 4s (1037 kB/s)
Reading package lists...
E: Failed to fetch https://packages.microsoft.com/ubuntu/20.04/prod/dists/focal/main/binary-amd64/Packages.bz2  File has unexpected size (69255 != 68824). Mirror sync in progress? [IP: 13.66.21.183 443]
...
E: Some index files failed to download. They have been ignored, or old ones used instead.
Error: Process completed with exit code 100.
```

Ah, GitHub actions.

gigr (approved):

db/migrate/20210504200438_add_github_updated_at_to_check_runs.rb
indiebrain: @casperisfine @ajshepley @gigr , #TIL Using `default: 0` to set a default calculated timestamp in MySQL only works in certain SQL modes - set by the server instance - and can cause errors when running migrations.

> The following examples use DEFAULT 0, a default that can produce warnings or errors depending on whether strict SQL mode or the NO_ZERO_DATE SQL mode is enabled. Be aware that the TRADITIONAL SQL mode includes strict mode and NO_ZERO_DATE.
-- https://dev.mysql.com/doc/refman/8.0/en/timestamp-initialization.html

Our MySQL server instances / clusters include the `NO_ZERO_DATE` SQL mode which means this migration fails without alteration. 

I _think_ the goal of `default: 0` is to set the github_updated_at value for new records to the current database time so that the check in `Shipit::CheckRun#create_or_update_by!`[^1] will avoid messaging nil, and indicate that Shipit _should_ save the CheckRun for the first webhook it receives. Am I close to the mark? I'm really interested in "Have I understood the intent / context correctly?"

In this case, for any folks like me who _might_ run into this issue, I _think_ the host application can  work around this failing migration by rewriting the default to use the `CURRENT_TIMESTAMP` instead of `0` like so:

```ruby
add_column :check_runs, :github_updated_at, :datetime , default: -> { "CURRENT_TIMESTAMP" }
```

References
----------

- https://dev.mysql.com/doc/refman/8.0/en/timestamp-initialization.html
- https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html
- https://dev.mysql.com/doc/refman/5.7/en/timestamp-initialization.html

[^1]: https://github.com/Shopify/shipit-engine/blob/ff4300326a981b9036df3565b719352bbb75c114/app/models/shipit/check_run.rb#L25

casperisfine: Either way, it look like this column is supposed to store GitHub data, so from a quick glance I feel like the default should have been `NULL`.

PR #1169, opened 2021-05-05
Fix corrupt git cache causes "exit status 128"

indiebrain:
When the stack's git cache repository is in an invalid state, tasks and
deployments will fail indicating that `git: exited with status 128`
until a human intervenes to clear the stack's git cache.

For example, when shipit first creates a stack a background job like
CacheDeploySpecJob and/or FetchDeployedRevisionJob will run shortly
after. These make use of the
StackCommands#with_temporary_working_directory method to populate the
local git cache before performing their primary work. When the initial
clone of the git cache fails the git repository may be in an invalid
state. Subsequent attempts to interact with the git repository will
result in the git program exiting with status 128.

A concrete scenario in which we've seen this surface. We recently
experienced instability due to over-scheduling of memory resources on
one of the nodes in our Kubernetes clusters on which these shipit-engine
jobs run. Some of the sidekiq workers running these jobs were OOMKilled
by the kernel due to congestion / contention on node memory. The result
is that these stacks remain in a state where they did not yet, nor would
they ever be able to, cache their deploy spec. This means they appear
"stuck" until a human realizes the on-disk git cache is broken and
manually schedules a ClearGitCacheJob through the UI for the stack.

The idea of behind this bod of work is to make shipit-engine more
resilient - self-healing - in such cases. The approach enhances
shipit-engine's existing file-system checks of the stack's git cache. If
the git cache is a working git repository, then use it. Otherwise throw
away the broken on-disk git cache and re-clone the repository.

### DIFF
--- a/lib/shipit/stack_commands.rb
+++ b/lib/shipit/stack_commands.rb
@@ -15,19 +15,25 @@ module Shipit
 
     def fetch
       create_directories
-      if Dir.exist?(@stack.git_path)
+      if valid_git_repository?(@stack.git_path)
         git('fetch', 'origin', '--tags', @stack.branch, env: env, chdir: @stack.git_path)
       else
+        @stack.clear_git_cache!
         git_clone(@stack.repo_git_url, @stack.git_path, branch: @stack.branch, env: env, chdir: @stack.deploys_path)
       end
     end
 
     def fetched?(commit)
-      git_dir = File.join(@stack.git_path, '.git')
-      if Dir.exist?(git_dir)
+      if valid_git_repository?(@stack.git_path)
         git('rev-parse', '--quiet', '--verify', "#{commit.sha}^{commit}", env: env, chdir: @stack.git_path)
       else
-        Command.new('test', '-d', git_dir, env: env, chdir: @stack.deploys_path)
+        # When the stack's git cache is not valid, the commit is
+        # NOT fetched. To keep the interface of this method
+        # consistent, we must return a Shipit::Command whose #success?
+        # method returns false - has a non-zero exit status. We utilize
+        # the POSIX 'test' command with no arguments which should
+        # always have an exit status of 1.
+        Command.new('test', env: env, chdir: @stack.deploys_path)
       end
     end
 
@@ -69,6 +75,18 @@ module Shipit
         git('checkout', commit.sha, chdir: git_dir).run! if commit
         yield Pathname.new(git_dir)
       end
+    end
+
+    def valid_git_repository?(path)
+      path.exist? &&
+        !path.empty? &&
+        git_cmd_succeeds?(path)
+    end
+
+    def git_cmd_succeeds?(path)
+      git("rev-parse", "--git-dir", chdir: path)
+        .tap(&:run)
+        .success?
     end
 
     def git_clone(url, path, branch: 'master', **kwargs)
--- a/test/unit/deploy_commands_test.rb
+++ b/test/unit/deploy_commands_test.rb
@@ -4,8 +4,9 @@ require 'test_helper'
 module Shipit
   class DeployCommandsTest < ActiveSupport::TestCase
     def setup
-      @stack = shipit_stacks(:shipit)
       @deploy = shipit_deploys(:shipit_pending)
+      @stack = @deploy.stack
+      @stack.stubs(:clear_git_cache!)
       @commands = DeployCommands.new(@deploy)
       @deploy_spec = stub(
         dependencies_steps!: ['bundle install --some-args'],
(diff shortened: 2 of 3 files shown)


casperisfine:
Ack. It might take a couple days to review this. But I agree with the premise of resiliency/self healing.

lib/shipit/stack_commands.rb
```diff
@@ -15,19 +15,25 @@ def env
 
     def fetch
       create_directories
-      if Dir.exist?(@stack.git_path)
+      if valid_git_repository?(@stack.git_path)
         git('fetch', 'origin', '--tags', @stack.branch, env: env, chdir: @stack.git_path)
       else
+        @stack.clear_git_cache!(with_lock: false)
```
indiebrain, line 21: If the stack's git cache repo is present, but corrupt, the subsequent call to `git_clone` will exit with status 128. The previous iteration looks like it assumed that if we entered this branch that the git was absent anyway. I _think_ this means that ensuring the precondition of an empty git cache here would be OK, or even desirable. 

The previous version of the `Stack#clear_git_cache!` method assumed that it is running in a context which has NOT already obtained the file-lock on the git cache. So far as I was able to determine this `#fetch` method is only called in a context where the outer scope has already acquired the git cache filesystem lock - while [executing a task / deployment](https://github.com/Shopify/shipit-engine/blob/master/app/models/shipit/task_execution_strategy/default.rb#L70-L76) and when [fetching fetching a commit's ref into a temporary working directory](https://github.com/Shopify/shipit-engine/blob/0edf50782d5b8d2b6e1976f96d35937a84c4c1b1/lib/shipit/stack_commands.rb#L54-L58). Since the outer scopes which call the `#fetch` method have already acquired the git cache file lock, this needs a way to delete the git cache without trying to reacquire the git cache lock a 2nd time - since this would cause a flock error.

I chose to add the ability to tell `clear_git_cache!` to skip trying to reacquire the git cache lock because we already have it. This prevents duplication of the logic to delete the git cache, but felt a bit goofy. Would love opinions and alternate points of view on this.

lib/shipit/stack_commands.rb
```diff
@@ -15,19 +15,25 @@ def env
...
         git('rev-parse', '--quiet', '--verify', "#{commit.sha}^{commit}", env: env, chdir: @stack.git_path)
       else
-        Command.new('test', '-d', git_dir, env: env, chdir: @stack.deploys_path)
+        # When the stack's git cache is not valid, the commit is
+        # NOT fetched. To keep the interface of this method
+        # consistent, we must return a Shipit::Command whose #success?
+        # method returns false - has a non-zero exit status. We utilize
+        # the POSIX 'test' command with no arguments which should
+        # always have an exit status of 1.
+        Command.new('test', env: env, chdir: @stack.deploys_path)
```
indiebrain, line 36: I attempted to infer the intent of the previous version of this code. I included the comment here to reveal - what I _think_ - the goal of this `Command` objects is. Did I "even come close" to understanding this correctly?

indiebrain (commented):
Thanks for the ACK, @casperisfine . I appreciate the lightning fast reply. 

I intentionally created this Pull Request in a draft state. This "works", but  I've got one potential alteration to the diff I want to chase down - might be able to remove some of it. Additionally, I had some textual context I wanted to include in the _formal_ request for review. My hope is to that the context would help frame some of the thoughts and tradeoffs considered in the formation of this change - maybe make the review process a bit easier. 

I'll mark this as "Ready for Review" (make it a REAL Pull Request) when its fit-for-human-consumption.

indiebrain:
Ok, I _think_ that's everything I wanted to button up. This is "Ready for Review" when you're ready @casperisfine . Thanks again, friend!

lib/shipit/stack_commands.rb
```diff
@@ -71,6 +77,18 @@ def with_temporary_working_directory(commit: nil)
       end
     end
 
+    def valid_git_repository?(path)
+      Dir.exist?(path) &&
```
casperisfine, line 81: I was about to say `path` is a `Pathname` so this code could be cleaned up, but turns out I'm wrong, for some reason `Stack#path` returns a `Pathname`, but `Stack#git_path` returns a `String`... I should clean that up.

lib/shipit/stack_commands.rb
indiebrain: Thanks @casperisfine ,  I was - slowly - working on something similar. I'll rebase and remove the `with_lock` bits. Appreciate the help.

test/unit/deploy_commands_test.rb
```diff
@@ -21,35 +22,79 @@ def setup
     end
 
     test "#fetch calls git fetch if repository cache already exist" do
-      Dir.expects(:exist?).with(@stack.git_path).returns(true)
+      @stack.stubs(:git_path).returns(mock_path)
```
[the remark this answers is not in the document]
casperisfine: > Stack's path objects aren't memoized - maybe for good reason.

Feel free to memoize them.

indiebrain:
@casperisfine , I _think_ I've addressed the feedback provided ~7days or so ago. Pinging to make sure that further discussion isn't being held up by something to which I've failed to adequately respond. Cheers.

casperisfine:
Ah yeah, I have commit push notifications disabled, otherwise I'd just be swarmed. So please ping me back when you address feedback on PRs. I'll have a look now.

PR #1180, opened 2021-05-20
Optionally record Review Stack latest deployed ref

indiebrain:
If enabled, `update_latest_deployed_ref` creates a branch for each
Review Stack that shipit-engine deploys. In a Repository with high
Review Stack turnover this quickly adds up to a lot of "noise" branches
in the git object database- our flagship monorepo currently has over 3k
such branches. However, in some scenarios it is still useful to record
the "latest deployed ref" of "plain old, manually managed" stacks - in
our case we DO still want to track the latest deployed ref of our
staging and production "environments" of this flagship monorepo.

This allows the host application to en/disable the recording of the
latest deployed ref for Review Stacks independently of "plain old,
manually defined" stacks.

### DIFF
--- a/app/models/shipit/review_stack.rb
+++ b/app/models/shipit/review_stack.rb
@@ -22,6 +22,10 @@ module Shipit
       end
     end
 
+    def update_latest_deployed_ref
+      # noop: last deployed ref is useless for review stacks
+    end
+
     model_name.class_eval do
       def route_key
         "stacks"
--- a/test/models/shipit/review_stack_test.rb
+++ b/test/models/shipit/review_stack_test.rb
@@ -74,5 +74,18 @@ module Shipit
         stack.unarchive!
       end
     end
+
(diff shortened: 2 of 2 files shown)


benlangfeld:
We should probably also prune those branches when a stack is archived.

casperisfine:
Hum, I don't think this feature make sense for review stack. I think I'd just disable it entirely for these without a config.

indiebrain:
@casperisfine , I can get behind that. Addressed by https://github.com/Shopify/shipit-engine/pull/1180/commits/c793e3a26b16bc3ca1ac048bd3ae7c8287ab9050

casperisfine (approved):
One style nitpick. Also please squash your commits and I'll merge.

indiebrain:
@casperisfine , thanks for your time and the conversation, friend!

PR #1183, opened 2021-05-27
Update logic to be either next or latest when valid

dahukish:
To simplify the dist tag strategy when deploying libraries we are updating the logic to do two things:

1) Check to see if a tag has a `-` after MAJOR.MINOR.PATCH and treat this as `next` (so not a full fledged release). Otherwise fallback to `latest`.
2) Validate that all package tags are of the same type, so we don't have mixed versions being deployed.

Tested with these tags:

```
const taggedPackages = [
  "v5.37.0",
  "v5.35.1",
  "@shopify/react-web-worker@2.0.0",
  "graphql-typed@1.0.0",
  "v5.37.0-1",
  "v5.35.1-1",
  "@shopify/react-web-worker@2.0.0-1",
  "graphql-typed@1.0.0-1",
  "v5.37.0-rc",
  "v5.35.1-rc",
  "@shopify/react-web-worker@2.0.0-rc",
  "graphql-typed@1.0.0-rc",
  "v5.37.0-beta",
  "v5.35.1-beta",
  "@shopify/react-web-worker@2.0.0-rc",
  "graphql-typed@1.0.0-rc"
];
```
Note: we also changed the ENV var `SHIPIT_LERNA_PUBLISH_MECHANISM` and how it is used a bit.

### DIFF
--- a/lib/snippets/publish-lerna-independent-packages
+++ b/lib/snippets/publish-lerna-independent-packages
@@ -6,39 +6,32 @@ const taggedPackages = execSync("git tag --points-at HEAD")
   .trim()
   .split("\n");
 
-const hasBetaRelease = taggedPackages.some((version) =>
-  version.includes("-beta")
-);
+// anything that matches `-` after MAJOR.MINOR.PATCH
+const validPattern = new RegExp(/\d+\.\d+\.\d+-(\w+)/);
 
-// set this using machine.environment.SHIPIT_LERNA_PUBLISH_FROM_GIT in your shipit.yml
-const useFromGit = process.env.SHIPIT_LERNA_PUBLISH_FROM_GIT || false;
-
-if (
-  hasBetaRelease &&
-  !taggedPackages.every((version) => version.includes("-beta"))
-) {
-  throw Error("All packages must be beta releases. Versions cannot be mixed.");
+// Ensure that all releases use the same prerelease suffix.
+// We want to only release all final versions, or all betas etc
+const allPrereleaseSuffixes =  taggedPackages.map((version) => {
+  const result = validPattern.exec(version);
+  return result ? result[1] : '';
+});
+if ((new Set(allPrereleaseSuffixes)).size > 1) {
+  throw Error("All packages must be of the same type of release. Versions cannot be mixed.");
 }
-const isNext =
-  !hasBetaRelease &&
-  taggedPackages.some((version) => {
-    return ["-alpha", "-rc", "-next"].some((distributionType) =>
-      version.includes(distributionType)
-    );
-  });
 
-function distTag() {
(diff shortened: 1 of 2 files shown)


lib/snippets/publish-lerna-independent-packages
```diff
@@ -10,27 +10,30 @@ const hasBetaRelease = taggedPackages.some((version) =>
...
+    return baseMatch === match;
+  }
+  return baseMatch[1] === match[1];
+});
+
 // set this using machine.environment.SHIPIT_LERNA_PUBLISH_FROM_GIT in your shipit.yml
 const useFromGit = process.env.SHIPIT_LERNA_PUBLISH_FROM_GIT || false;
 
-if (
-  hasBetaRelease &&
```
gigr, line 17: `hasBetaRelease` is still being set on line 9. It can be removed, right?

BPScott (commented):
Reading lerna docs i've spotted the the [`--pre-dist-tag`](https://github.com/lerna/lerna/tree/6cb8ab2d4af7ce25c812e8fb05cd04650105705f/commands/publish#--dist-tag-tag) option.

Which leads me to wonder - are we over complicating all this? Does lerna do the right thing if we give it minimal options?

Could we do the following and bypass writing any of our own "what tag should i use" stuff?

```js
const command = [
  "node_modules/.bin/lerna publish",
  useFromGit ? "from-git" : "from-package",
  "--yes",
  "--dist-tag latest",
  "--pre-dist-tag next"
];
```

lib/snippets/publish-lerna-independent-packages
```diff
@@ -6,31 +6,30 @@ const taggedPackages = execSync("git tag --points-at HEAD")
...
+const isValid = taggedPackages.every((version) => {
+  const match = validPattern.exec(version);
+  if (baseMatch === null) {
+    return baseMatch === match;
+  }
+  return baseMatch[1] === match[1];
+});
 
 // set this using machine.environment.SHIPIT_LERNA_PUBLISH_FROM_GIT in your shipit.yml
 const useFromGit = process.env.SHIPIT_LERNA_PUBLISH_FROM_GIT || false;
```
BPScott, line 24: Not related to this PR but in retrospect perhaps this should have been `SHIPIT_LERNA_PUBLISH_MECHANISM` and have it default to `from-package`, and users could set `SHIPIT_LERNA_PUBLISH_MECHANISM='from-git'`, to allow for possible new approaches in the future.


I'm also not that convinced about from-package being the default - it feels like `from-git` would be preferable if we're doing stuff like asserting that this commit has a bunch of tags on it. `from-package` doesn't use any of that info

gigr (approved):

PR #1184, opened 2021-06-08
Use the newer GitHub checkrun date, and persist conflicting checkruns before refreshing.

ajshepley:
## The problem

Ref: https://github.com/Shopify/shipit-engine/pull/1167, also https://github.com/Shopify/shipit-engine/pull/1182

We observed an issue with stampeding runoff checkrun refreshes.

This appeared to be due to a few things, though I wasn't able to fully verify all of them:

- Conflicting data with new or equivalent timestamps meaning that refreshes would retrieve a checkrun that could trigger a new refresh.
- Checkrun conclusions and states can theoretically go from any, to any. The actual state transition rules do not seem strictly defined, so we need to be robust against this.
- Manual checkrun refreshes would result in webhooks and checkrun data firing that could "conflict" in the moment but was canonically accurate in some sense, and those webhooks could conflict with each other and ensure that eternally-conflicting refresh(es) were enqueued
- The completed_at timestamp in GitHub's webhooks were not necessarily the newest timestamp available, nor were these timestamps always accurate to when the conclusion was reached/when the Checkrun was actually modified (!!!) (Unfortunately, I don't seem to have retained the logs showing these events 😐)

Most importantly, there was effectively no guard against endless refreshes, or perpetually stale data; if the data on github entered what we considered a "conflict state", but that was the actual end state of the checkrun (an app was not going to update the checkrun further), then the code as written would never stop refreshing the checkrun for new data.

## This PR / The Solution

This PR resolves these problems by doing the following:

1. Always persist checkrun data that has been received via webhook. This is what we were doing before https://github.com/Shopify/shipit-engine/pull/1167 to begin with.
2. Parse, utilize/compare and persist the newest date from either started_at or completed_at. We're only interested in "github_updated_at", e.g. the newest date available from the webhook data.
3. If we detect a conflict and the webhook _seems_ older than our record, log it and enqueue a refresh, but still persist it.
4. Add a test that checks that a conflict checkrun gets persisted, and refreshes, but stops refreshing.

This should prevent the aforementioned endless erroneous refreshes, and ensure that we reach a state of finality (insofar as the checkrun is "done" changing on github).

There's still a possibility that we'll persist out-of-order data _and_ that github's API will return stale data for longer than `CHECK_RUN_REFRESH_DELAY`, but I think this is the best compromise given the other limitations with the data they provide for checkruns via the API and the webhooks they emit to us.

---

This should allow us to also ship https://github.com/Shopify/shipit-engine/pull/1182. One option would be a variant of #1182, which enqueues a late refresh rather than persisting data; this PR makes that option safer as well.

### DIFF
--- a/test/models/commits_test.rb
+++ b/test/models/commits_test.rb
@@ -351,6 +351,7 @@ module Shipit
           title: 'Tests build ran successfully',
         ),
         completed_at: Time.now,
+        started_at: Time.now - 1.minute,
       )
       response = mock(
         check_runs: [check_run],
--- a/test/models/shipit/check_run_test.rb
+++ b/test/models/shipit/check_run_test.rb
@@ -10,62 +10,106 @@ module Shipit
     end
 
     test ".create_or_update_from_github! updates successfully" do
-      checkrun_time = Time.now
+      completed_at = Time.now
       assert_difference -> { @commit.check_runs.count }, +1 do
         @commit.check_runs.create_or_update_from_github!(
           @stack.id,
-          github_check_run(conclusion: nil, checkrun_time: '2021-04-29T18:05:12Z')
+          github_check_run(conclusion: nil, completed_at: '2021-04-29T18:05:12Z')
         )
       end
 
       assert_no_enqueued_jobs(only: RefreshCheckRunsJob) do
         @commit.check_runs.create_or_update_from_github!(
           @stack.id,
-          github_check_run(conclusion: 'success', checkrun_time: checkrun_time + 1.minute)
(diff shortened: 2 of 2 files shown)


ajshepley:
>  but some of our update check run mutations are sending updates with a new conclusion but not a new completed_at.
> https://docs.github.com/en/graphql/reference/input-objects#updatecheckruninput

Ah. We have various services that could write checkruns, so it's always possible that some of them may not correctly update both fields every time. Better to be resilient against that eventuality than to rely on ensuring that every app conforms to expectations (though we should contact application owners when we do observe this problem).

Thanks for the insight (and the help debugging a while back)!

PR #1185, opened 2021-06-09
Allow changing the branch name via the API.

gigr:
This updates the `api/stacks#update` endpoint to allow changing the branch name. Changing branches means shipit-engine must re-sync git data for the new branch and it could also lead to issues with internal state. Changing this value while a deploy is ongoing for the stack might cause issues, but changing it while the stack is "at rest" should not.

I'm not sure there's anything else that needs to change in this PR to be honest. When code is checked out in shipit-next it's based on the commit SHA, not the branch name. We'll probably also start dropping webhooks once the branch is changed, but this is by design.

https://github.com/Shopify/shipit-engine/blob/c50c4a400c543dcd738c88d1d56f52006fb597eb/lib/shipit/task_commands.rb#L49-L66

cc @Shopify/deploys

### DIFF
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -86,7 +86,7 @@ module Shipit
     after_commit :sync_github_if_necessary, on: :update
 
     def sync_github_if_necessary
-      if archived_since_previously_changed? && archived_since.nil?
+      if (archived_since_previously_changed? && archived_since.nil?) || branch_previously_changed?
         sync_github
       end
     end
--- a/app/controllers/shipit/api/stacks_controller.rb
+++ b/app/controllers/shipit/api/stacks_controller.rb
@@ -40,6 +40,7 @@ module Shipit
       end
 
       params do
+        accepts :branch, String
         accepts :deploy_url, String
         accepts :ignore_ci, Boolean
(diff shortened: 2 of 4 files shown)


ajshepley:
cc: @casperisfine Any problems with enabling this on the API front? I know renames might cause weirdness with local/disk git state but is there anything else you think we should be concerned about?

casperisfine:
Let's try.

suprnova32 (approved):

app/models/shipit/stack.rb
```diff
@@ -86,7 +86,7 @@ def lock_author=(user)
     after_commit :sync_github_if_necessary, on: :update
 
     def sync_github_if_necessary
-      if archived_since_previously_changed? && archived_since.nil?
+      if (archived_since_previously_changed? && archived_since.nil?) || branch_previously_changed?
```
ajshepley, line 89: Since this method is called on update, is `*_previously_changed?` necessary?

gigr: `on :update` runs on any change. Checking for `branch_previously_changed?` ensures we don't sync on _all_ updates.

ajshepley (approved):

PR #1186, opened 2021-06-10
Give configured `shell_path`'s precedence for commands

DazWorrall:


ajshepley:
Keeping the branch around in case Shipit needs to revert to the current pin later on.

PR #1190, opened 2021-06-28
Added note about max_commits: null.

ajshepley:
You can have CD for any number of commits using an explicit `null` so I figure it's worth mentioning.

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,7 +270,9 @@ deploy:
 ```
 <br>
 
-**<code>deploy.max_commits</code>** defines the maximum number of commits that should be shipped per deploys. Defaults to `8`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -270,7 +270,9 @@ deploy:
 ```
 <br>
 
-**<code>deploy.max_commits</code>** defines the maximum number of commits that should be shipped per deploys. Defaults to `8`.
+**<code>deploy.max_commits</code>** defines the maximum number of commits that should be shipped per deploys. Defaults to `8` if no value is provided.
```
stephsachrajda, line 273:
```suggestion
**<code>deploy.max_commits</code>** defines the maximum number of commits that should be shipped per deploy. Defaults to `8` if no value is provided.
```

This might be wrong, my English is weak

stephsachrajda (approved):

benlangfeld (approved):

PR #1196, opened 2021-07-14
Make locking optional on rollbacks

stephsachrajda:
### What

Adds a param to control whether to lock when a rollback is triggered. Defaults to true so not breaking current behaviour.

### Why

In our use case, deploys will be triggered by a separate application via the api. If a certain commit is not desired, we both rollback the deploy and drop the commit (and all later commits) from the tracked branch. In this case we do not required a lock because the system is working as intended.

### DIFF
--- a/app/models/shipit/deploy.rb
+++ b/app/models/shipit/deploy.rb
@@ -97,14 +97,16 @@ module Shipit
     end
 
     # Rolls the stack back to this deploy
-    def trigger_rollback(user = AnonymousUser.new, env: nil, force: false)
+    def trigger_rollback(user = AnonymousUser.new, env: nil, force: false, lock: true)
       rollback = build_rollback(user, env: env, force: force)
       rollback.save!
       rollback.enqueue
 
-      lock_reason = "A rollback for #{rollback.since_commit.sha} has been triggered. " \
-        "Please make sure the reason for the rollback has been addressed before deploying again."
-      stack.update!(lock_reason: lock_reason, lock_author_id: user.id)
+      if lock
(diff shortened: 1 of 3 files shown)


app/models/shipit/deploy.rb
```diff
@@ -97,14 +97,16 @@ def build_rollback(user = nil, env: nil, force: false)
...
       rollback.save!
       rollback.enqueue
 
-      lock_reason = "A rollback for #{rollback.since_commit.sha} has been triggered. " \
-        "Please make sure the reason for the rollback has been addressed before deploying again."
-      stack.update!(lock_reason: lock_reason, lock_author_id: user.id)
+      if lock
+        lock_reason = "A rollback for #{rollback.since_commit.sha} has been triggered. " \
+          "Please make sure the reason for the rollback has been addressed before deploying again."
+        stack.update!(lock_reason: lock_reason, lock_author_id: user.id)
```
ajshepley, line 108: Do we need to trigger a stack update in a different way now, to ensure any related hooks will still be processed when a rollback is triggered? Or is creating and enqueueing the rollback sufficient?

That is, could we have been relying on this update! call for other side-effects?

stephsachrajda: Looking at the events on update I don't believe we need those since the stack isn't actually changing, but I'm not super familiar with shipit-engine so I'm open to being wrong. When I went through them everything related to the stack actually changing and broadcasting that change which is doesn't in this case (and we don't trigger them on a regular deploy either from what I can see).

FWIW I was able to run this locally and didn't hit any roadblocks, but if there's extra stuff you think I should test that I might not know about, I'm super game!

PR #1202, opened 2021-07-27
Adjust padding in plugins container

kirs:
Before (the div is way bigger than its container):

Screenshot 2021-07-27 at 13 28 03

After (the div is properly sized):

Screenshot 2021-07-27 at 13 27 47

### DIFF
--- a/app/assets/stylesheets/_pages/_deploy.scss
+++ b/app/assets/stylesheets/_pages/_deploy.scss
@@ -196,8 +196,6 @@
   width: 100%;
   overflow-y: auto;
   overflow-x: hidden;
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/_pages/_deploy.scss
```diff
@@ -196,8 +196,6 @@
   width: 100%;
   overflow-y: auto;
   overflow-x: hidden;
-  padding-right: 20px;
```
ajshepley, line 199: I wonder why we have this padding to begin with.

ajshepley (approved):

PR #1207, opened 2021-08-17
Don't mark rollback tasks as faulty

rdji123:
Shipit-engine currently marks successful rollback tasks as faulty. This is because `update_release_status` runs for any changes in transitions. And when the rollback task succeeds the last completed task  (`stack.last_completed_deploy.id`) is itself and thus marks itself as faulty. 

I also overrode `report_complete!` so rollbacks will not be entering validations.

### DIFF
--- a/app/models/shipit/rollback.rb
+++ b/app/models/shipit/rollback.rb
@@ -33,6 +33,10 @@ module Shipit
       'deploys/deploy'
     end
 
+    def report_complete!
+      complete!
+    end
+
     private
 
     def update_release_status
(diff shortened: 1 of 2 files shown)


stephsachrajda (approved):

ajshepley:
> I also overrode report_complete! so rollbacks will not be entering validations.

Is this correct/desired behaviour? I wouldn't be surprised if it was a breaking change for some users.

cc: @casperisfine what do you think?

rdji123:
@ajshepley It doesn't seem to be entering validation in the current state as well. It's marked as faulty right after the rollback succeeds

kwboyd-shopify (approved):

app/models/shipit/rollback.rb
```diff
@@ -33,13 +33,18 @@ def to_partial_path
...
+    end
+
     private
 
     def update_release_status
       return unless stack.release_status?
 
       # When we rollback to a certain revision, assume that all later deploys were faulty
       stack.deploys.newer_than(deploy.id).until(stack.last_completed_deploy.id).to_a.each do |deploy|
+        next if deploy.id == id
```
casperisfine, line 47: I don't get it, `newer_then(deploy.id)` should translate as `id > #{deploy.id}`, so `deploy` shouldn't be yielded by that `each`.

PR #1208, opened 2021-08-17
Allow the deploy banner to stretch so that buttons do not get hidden

kwboyd-shopify:
Right now it's possible for the text of the "Abort" and "Abort and Rollback" buttons to be cut-off if users are viewing this on smaller screens. This PR allows the banner to stretch taller to accommodate those buttons.

The issue:
Screen Shot 2021-08-17 at 4 05 42 PM

### DIFF
--- a/app/assets/stylesheets/_pages/_deploy.scss
+++ b/app/assets/stylesheets/_pages/_deploy.scss
@@ -67,7 +67,8 @@
 }
 
 .deploy-banner {
-  height: 4rem;
+  min-height: 4rem;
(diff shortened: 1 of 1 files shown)


ajshepley (approved):

ajshepley:
Should we set a max height value as well?

kwboyd-shopify:
@ajshepley We shouldn't need to. `auto` means it will automatically adjust to fit whatever's inside of it, so the only way we'd run into issues is if we cram a ton of content into that div.

PR #1210, opened 2021-08-20
Fix Default value type mismatch in add_github_updated_at_to_check_runs migration

### LINKED ISSUE #1209: Default value type mismatch in `add_github_updated_at_to_check_runs` migration
Getting the following error when running the latest migration on a postgresql database:

```
PG::DatatypeMismatch: ERROR:  column "github_updated_at" is of type timestamp without time zone but default expression is of type integer
HINT:  You will need to rewrite or cast the expression.
```

setting the default value to `nil` instead of `0` fixed the issue

pcothenet:
Fixes #1209


pcothenet:
CLA accepted but I am unable to re-run the test.

PR #1218, opened 2021-08-30
Clarify deploy start and finish datetime

maximevaillancourt:
This commit clarifies when a deploy started:

- Before: "20 minutes ago in 3m21"
- After: "Started 20 minutes ago and finished in 3m21s"

Without this commit, it's unclear whether the "20 minutes"
part is when the deploy started or when it finished.

With this commit, it's much clearer.

### DIFF
--- a/app/views/shipit/deploys/_deploy.html.erb
+++ b/app/views/shipit/deploys/_deploy.html.erb
@@ -30,12 +30,12 @@
       </p>
       <p class="commit-meta">
         <% if read_only %>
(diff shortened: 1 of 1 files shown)


casperisfine:
Looks good to me, could you include a screenshot though? I'm just worried the extra size might cause wrapping or things like that.

maximevaillancourt:
Closing: doesn't seem necessary (hey Jean, it's been a while) 😄

PR #1223, opened 2021-09-16
Release v0.36.0

atesgoral:


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Unreleased
 
+# 0.36.0
```
ajshepley, line 3: Would 0.35.1 make sense in this case?

ajshepley: Yeah I noticed there were a few of them. I don't know if there's any standardization here, but it can kinda make it clear that the changes in this version are related to the code in 0.35.0 which I think is useful.

PR #1236, opened 2021-11-22
Detect and refresh legacy GitHub user tokens

DazWorrall:
We received this from GitHub regarding our production instance:

> We noticed that an application, Shopify Shipit, owned by an organization you are an admin of, Shopify, used a token with an outdated format to access the GitHub API ...
> 
> ... we encourage you to reset any authentication tokens used by this app, as well as tokens used by any other apps you may have, with our reset token API.
> 
> Alternatively, you can prompt your users to step through the authorization flow again, as outlined in the docs for either GitHub Apps and OAuth Apps.
> 
> To understand more about this change and why it's important, visit https://github.blog/2021-04-05-behind-githubs-new-authentication-token-formats.

This PR will detect tokens in the legacy format and take users back through the GitHub login flow to fetch an updated one. Because we aren't asking for updated permissions, this will be totally transparent to the end user - just a slightly longer page load as the browser follow some redirects to GitHub and back again. I 🎩'd this locally by updating my user with a garbage token (`user.update!(github_access_token: 'blah')` and reloading the page - it worked as intended.

### DIFF
--- a/app/controllers/concerns/shipit/authentication.rb
+++ b/app/controllers/concerns/shipit/authentication.rb
@@ -17,7 +17,11 @@ module Shipit
     private
 
     def force_github_authentication
-      if Shipit.authentication_disabled? || current_user.logged_in?
+      if current_user.logged_in? && current_user.requires_fresh_login?
+        Rails.logger.warn("User #{current_user.id} requires a fresh login, logging out...")
+        reset_session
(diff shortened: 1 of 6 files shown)


casperisfine (approved):

suprnova32 (approved):

White731964:
Hi

PR #1237, opened 2021-12-09
Update stack environment through api

erik-shopify:
This PR allows a stack's name to updated through the api.

### DIFF
--- a/app/controllers/shipit/api/stacks_controller.rb
+++ b/app/controllers/shipit/api/stacks_controller.rb
@@ -40,6 +40,7 @@ module Shipit
       end
 
       params do
+        accepts :environment, String
         accepts :branch, String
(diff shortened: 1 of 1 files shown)


kwboyd-shopify (approved):

app/controllers/shipit/api/stacks_controller.rb
```diff
@@ -40,6 +40,7 @@ def create
       end
 
       params do
+        accepts :environment, String
```
ajshepley, line 43: Out of curiosity, if we have different shipit.ymls in the repo, and you switch to a new name, does it correctly refresh and pull down the "new" one (`shipit.newenvname.yml`)?

If not, it may be worth enqueueing a refresh when this field is updated.

PR #1247, opened 2022-02-11
Allow nil deploy url

erik-shopify:
Allows `nil` `deploy_url` in stack update and creation.


kwboyd-shopify:
Mind adding a test to `stacks_controller_test.rb`? It'd be helpful to have a test that fails on `master` and passes on this branch so that we can 100% verify this fixes that API error we saw.

kwboyd-shopify (approved):

PR #1258, opened 2022-04-11
Configurable respect bare shipit file behavior

Archetypically:
Resolves https://github.com/Shopify/shipit-next/issues/3563.

# Changes

- Expose a new interface to explicitly disable the handling of bare `shipit.yml` (or `#{app_name}.yml`) files via attribute `Shipit.respect_bare_shipit_file`
   - Behavior is configurable, but **left on by default to avoid breaking changes.**
- Continues to preserve the hierarchy of file naming conventions:
   1. `#{app_name}.#{env}.yml`
   2. `#{app_name}.yml`
   3. `shipit.#{env}.yml`
   4. `shipit.yml`
- Will explicitly override `deploy.pre` with an `exit 1` command and an informative error message when this flag is toggled to off and a deploy is attempted on a Stack with a bare file:

invalid_shipit_file_message

### DIFF
--- a/lib/shipit.rb
+++ b/lib/shipit.rb
@@ -69,6 +69,7 @@ module Shipit
     :preferred_org_emails,
     :task_execution_strategy,
     :task_logger,
+    :respect_bare_shipit_file,
   )
 
   def task_execution_strategy
@@ -234,6 +235,13 @@ module Shipit
   def task_logger
     @task_logger ||= Logger.new(nil)
   end
+
+  def respect_bare_shipit_file
+    @respect_bare_shipit_file = true if @respect_bare_shipit_file.nil?
+    !!@respect_bare_shipit_file
+  end
+
+  alias_method :respect_bare_shipit_file?, :respect_bare_shipit_file
 
   protected
 
--- a/lib/shipit.rb
+++ b/lib/shipit.rb
@@ -63,13 +63,12 @@ module Shipit
 
   delegate :table_name_prefix, to: :secrets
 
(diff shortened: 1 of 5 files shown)


lib/shipit.rb
```diff
@@ -85,6 +85,10 @@ def enable_samesite_middleware?
     ENV['SHIPIT_ENABLE_SAMESITE_NONE'].present?
   end
 
+  def obey_bare_shipit_file?
+    ActiveModel::Type::Boolean.new.cast(ENV['SHIPIT_OBEY_BARE_SHIPIT_FILE'] || true)
```
suprnova32, line 89: _Really_ small nitpick, but the repetition of `SHIPIT` looks a bit weird to me. What do you think about calling it `SHIPIT_OBEY_BARE_CONFIG_FILE`?

suprnova32 (approved):

erik-shopify (approved):

lib/shipit.rb
```diff
@@ -235,6 +236,13 @@ def task_logger
     @task_logger ||= Logger.new(nil)
   end
 
+  def respect_bare_shipit_file
+    @respect_bare_shipit_file = true if @respect_bare_shipit_file.nil?
+    !!@respect_bare_shipit_file
+  end
```
casperisfine, line 242: Just use a regular accessor and then set the default with `self. respect_bare_shipit_file = true`

casperisfine (approved):

PR #1259, opened 2022-04-20
Use deep_merge! versus direct access

Archetypically:
# Changes

- Fixes the edge case where a deployment attempt occurs with `respect_bare_shipit_yml` set to false, but the `shipit.yml` does not contain a `deploy:` top level key.


kwboyd-shopify (approved):
neat!

PR #1260, opened 2022-04-25
Correct param argument for refresh controller

rdji123:
Related to https://github.com/Shopify/shipit/issues/1988

This endpoint was exposed a while ago but was passing the wrong param argument


Archetypically (approved):
:big:

PR #1265, opened 2022-05-17
[Doc] Organise permissions in Github App creation

mathieutu:
Hi,

I was creating the Github app and have seen that the permissions listed are not in the good order, with the same label than Github, so we have to search for them in the page.

This PR just fixes it, so it'll be easier to follow the instructions.

Thanks!


casperisfine:
Thanks.

PR #1270, opened 2022-06-27
Change `emit_lock_hooks` to not check for `previous_changes` in order to enable it running after "Abort & Rollback" is clicked

kwboyd-shopify:
**What:**
This gets rid of the automatic return in `emit_lock_hooks` if `previous_changes` doesn't contain a change to `lock_reason`.

**Why:**
Right now, hitting "Abort & Rollback" on a deploy does not actually end up emitting a lock hook. This is because of [this behavior](https://medium.com/ruby-on-rails/activerecord-transaction-gotchas-277c048dc3ca) in ActiveRecord. `previous_changes` was ending up empty when `emit_lock_hooks` was actually called in the callbacks.

**Risks:**
This does mean we'll be emitting a lot more "lock" webhooks, even when the lock has not changed. Are we okay with this, or would we prefer one of the alternatives below?

**Alternatives:**
- Changing `emit_lock_hooks` to run on `after_save` fixes this bug as well, but is a change in behavior that risks us emitting a "lock" hook before the lock actually gets committed to the DB.
- There's a gem in the link above that we could add to keep track of transaction changes

### DIFF
--- a/app/models/shipit/stack.rb
+++ b/app/models/shipit/stack.rb
@@ -642,9 +642,7 @@ module Shipit
     end
 
     def emit_lock_hooks
-      return unless previous_changes.include?('lock_reason')
-
-      lock_details = if previous_changes['lock_reason'].last.blank?
+      lock_details = if previous_changes.include?('lock_reason') && previous_changes['lock_reason'].last.blank?
         { from: previous_changes['locked_since'].first, until: Time.zone.now }
       end
 
--- a/test/models/deploys_test.rb
+++ b/test/models/deploys_test.rb
@@ -459,6 +459,32 @@ module Shipit
       end
     end
 
+    test "transitioning to aborted locks the stack if a rollback is scheduled" do
+      refute @stack.locked?
+
+      @deploy = shipit_deploys(:shipit_running)
+      @deploy.ping
+      @deploy.pid = 42
+      @deploy.abort!(rollback_once_aborted: true, aborted_by: @user)
+      @deploy.aborted!
+
+      assert @stack.reload.locked?
+      assert_equal @user, @stack.lock_author
+    end
+
+    test "transitioning to aborted emits a lock hook if a rollback is scheduled" do
+      refute @stack.locked?
+
+      @deploy = shipit_deploys(:shipit_running)
+      @deploy.ping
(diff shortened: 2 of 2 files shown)


app/models/shipit/stack.rb
```diff
@@ -642,9 +642,7 @@ def schedule_merges_if_necessary
     end
 
     def emit_lock_hooks
-      return unless previous_changes.include?('lock_reason')
-
-      lock_details = if previous_changes['lock_reason'].last.blank?
+      lock_details = if previous_changes.include?('lock_reason') && previous_changes['lock_reason'].last.blank?
```
stephsachrajda, line 645: Am I understanding this right, that the issue is that with this rails bug, `previous_changes` doesn't include the lock information?

In that case isn't this now just sending blank webhooks (at least as far as locking info)? I feel like that isn't going to be any better than before and might break stuff if it's expecting these fields to be set?

I don't really have a better solution. I'm not opposed to an `after_save` as the transaction can only roll back the DB and any third party api calls are still going to happen right away, so it's not exactly perfect the way it is. 

Maybe there's a way to separate the two so that the rollback isn't a side effect of aborting and instead we do them in succession? That would make them separate transactions. The webhooks are working for regular rollbacks

casperisfine: We can't do it from `after_change`, if we did the ob could be ran before the transaction is committed, and send the wrong data.

casperisfine:
> This does mean we'll be emitting a lot more "lock" webhooks, even when the lock has not changed. Are we okay with this

I don't think we're OK with this. It will mess with chat notifications.

Can we just do an ad hoc fix (e.g. emit a hook after a rollback is created?)

app/models/shipit/stack.rb
kwboyd-shopify: Both of what you said makes sense - yeah, I think making them separate transactions is a cleaner fix but a riskier one. And yes, the issue is that `previous_changes` doesn't include the lock info in the case of "Abort & Rollback." I'm going to create a new PR that just calls `emit_lock_hooks` directly from within `trigger_revert`. I'll have that up in a moment and will close this - thank you both for the feedback!

PR #1219, opened 2021-09-15
Run git commands less noisily

atesgoral:
The initial `git clone` spams the log with progress output:

```
$ git clone --single-branch --recursive --branch master https://github.com/shopify/shipit.git /app/data/stacks/shopify/shipit/production/git
pid: 32373
Cloning into '/app/data/stacks/shopify/shipit/production/git'...
remote: Enumerating objects: 21925, done.        
remote: Counting objects:   0% (1/2009)        
remote: Counting objects:   1% (21/2009)        
remote: Counting objects:   2% (41/2009)        
...
```
_many many lines later_
```
...
Resolving deltas:  99% (14043/14182)   
Resolving deltas: 100% (14182/14182)   
Resolving deltas: 100% (14182/14182), done.
```

Adding a `--quiet` to keep git quieter.

The commit hash checkout results in git giving us a lecture about detached heads:

```
$ git checkout 0216b407c09b8198caec9ec12786e3c6a952cb1c
pid: 32416

Note: checking out '0216b407c09b8198caec9ec12786e3c6a952cb1c'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>
```

Adding `--config advice.detachedHead=false` to turn off this advisory.

PR #1278, opened 2022-08-09
Add --quiet switches to checkouts

atesgoral:
Follow-up to #1219 and #1225. Missed a spot.

Output is garbled due to progress feedback of `checkout`:

image

Tophatted locally:

```
git -c advice.detachedHead=false checkout --quiet 05f5b2e2b72a80491d8ebe3e91a350a7ae09dc5e
```

### DIFF
--- a/lib/shipit/stack_commands.rb
+++ b/lib/shipit/stack_commands.rb
@@ -72,7 +72,14 @@ module Shipit
         ).run!
 
         git_dir = File.join(dir, @stack.repo_name)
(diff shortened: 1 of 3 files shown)


stephsachrajda (approved):

atesgoral:
@ajshepley I've thinking about the same. Other consumers of shipit-engine might still use progress output. Might make sense to revise all the silencing we've added so far and add an option to toggle verbosity / progress.

PR #1282, opened 2022-09-13
Add lock and archive functionality to the stack update API call

mikailkarimi:
### 🤷 What are you trying to accomplish?
We need API support for archive/lock a stack for our project.
Modifying the API to be able to perform archive and lock functionalities on one stack. This includes unarchiving and unlocking functionalities as well. 

### 🛠️ How did you accomplish this and why did you choose this approach?

Modified the api/stacks_controller.rb to align with the current stacks_controller.rb. In order to avoid modification to the API behaviour, when the keys are not provided, we do not perform any modification to the stack. I added testing to the testing file to accommodate this change. 

### 🎩 Tophat
In order to run tests in local, I installed redis.
```
brew install redis
redis-server

# another terminal tab
redis-cli
> ping
```

Then because I'm using M1, so I need to follow the [instruction](https://github.com/Shopify/shipit-engine/pull/1279#issue-1333802477) to install `libpq` before I can `bundle install`
```
brew install libpq
export PATH="/opt/homebrew/opt/libpq/bin:$PATH"
bundle install
```
 
Then all tests should pass.

### DIFF
--- a/app/controllers/shipit/api/stacks_controller.rb
+++ b/app/controllers/shipit/api/stacks_controller.rb
@@ -46,9 +46,13 @@ module Shipit
         accepts :ignore_ci, Boolean
         accepts :merge_queue_enabled, Boolean
         accepts :continuous_deployment, Boolean
+        accepts :archived, Boolean
       end
       def update
-        stack.update(params)
+        stack.update(update_params)
+
+        update_archived
+
         render_resource(stack)
       end
 
@@ -76,6 +80,26 @@ module Shipit
 
       def stack
         @stack ||= stacks.from_param!(params[:id])
+      end
+
+      def update_archived
+        if key?(:archived)
+          if params[:archived]
+            stack.archive!(nil)
+          elsif stack.archived?
+            stack.unarchive!
+          end
+        end
+      end
+
+      def key?(key)
+        params.to_h.key?(key)
+      end
+
+      def update_params
+        params.select do |key, _|
+          %i(environment branch deploy_url ignore_ci merge_queue_enabled continuous_deployment).include?(key)
+        end
       end
 
       def repository
--- a/test/controllers/api/stacks_controller_test.rb
+++ b/test/controllers/api/stacks_controller_test.rb
@@ -114,6 +114,57 @@ module Shipit
         refute @stack.continuous_deployment
       end
 
+      test "#update does not perform archive when key is not provided" do
+        refute_predicate @stack, :archived?
+        refute_predicate @stack, :locked?
+
+        patch :update, params: { id: @stack.to_param }
+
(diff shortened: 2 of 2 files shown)


app/controllers/shipit/api/stacks_controller.rb
```diff
@@ -78,6 +84,37 @@ def stack
...
+            stack.lock(reason, nil)
+          elsif stack.locked?
+            stack.unlock
+          end
+        end
+      end
+
+      def update_archived
+        if key?(:archived)
+          if params[:archived] == "true"
```
casperisfine, line 100: This doesn't seem right at all. With `explicit-parameters` you get proper types. `:archived` is declared as a boolean, you should compare it with a String.

app/controllers/shipit/api/stacks_controller.rb
```diff
@@ -78,6 +84,37 @@ def stack
         @stack ||= stacks.from_param!(params[:id])
       end
 
+      def update_lock
```
casperisfine, line 87: There is already an API to lock/unlock stacks https://github.com/Shopify/shipit-engine/blob/master/app/controllers/shipit/api/locks_controller.rb

casperisfine:
> Do we need to squash all changes into one to keep history clean before merge?

You can have multiple commits if they makes sense. Commits like "fix linter" etc should definitely be squashed.

> Do we need to release a new version for this API change

No. Releases are handled separately. And you don't need to cut a release to get your change deployed on our Shipit instance.

app/controllers/shipit/api/stacks_controller.rb
xiaoxipang: Sorry, my bad, I should keep them consistent

app/controllers/shipit/api/stacks_controller.rb
xiaoxipang: Got it, thank you!

app/controllers/shipit/api/stacks_controller.rb
xiaoxipang: Done

casperisfine: This should be a boolean. The API is always JSON, so it makes no sense to pass it as a string.

casperisfine: That's my fault, the API tests should all pass `as: :json` to better simulate true API requests. I fixed that in https://github.com/Shopify/shipit-engine/pull/1283, if you rebase you should be able to test with true booleans.

zarifmahfuz (approved):
Can you remove the redis database before merging?

PR #1289, opened 2022-12-15
Add a `safeties_enforced` param to rollback and deploy API endpoints which prevents `ignored_safeties` from being set

kwboyd-shopify:
This adds a `safeties_enforced` parameter to the rollbacks and deploys API endpoints. If you pass this in, even if you also pass `force=true`, `ignored_safeties` will be set to `false`. This is for systems that may use Shipit's API and require some of the powers that `force` enables, but have their own safeties requirements that may differ from Shipit's.

**Note:** there is an existing bug in which rolling back using `force` when there is an active task running does not lead to the rollback being marked as `ignored_safeties: true`. Right now, if you force a rollback, the rollback will _only_ be marked as `ignored_safeties: true` if there was not a task already in process. This pull request does not fix that bug.

### DIFF
--- a/app/controllers/shipit/api/deploys_controller.rb
+++ b/app/controllers/shipit/api/deploys_controller.rb
@@ -13,12 +13,14 @@ module Shipit
         accepts :force, Boolean, default: false
         accepts :require_ci, Boolean, default: false
         accepts :env, Hash, default: {}
+        accepts :safeties_enforced, Boolean, default: false
       end
       def create
         commit = stack.commits.by_sha(params.sha) || param_error!(:sha, 'Unknown revision')
         param_error!(:force, "Can't deploy a locked stack") if !params.force && stack.locked?
         param_error!(:require_ci, "Commit is not deployable") if params.require_ci && !commit.deployable?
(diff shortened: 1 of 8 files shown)


app/controllers/shipit/api/deploys_controller.rb
```diff
@@ -13,12 +13,14 @@ def index
         accepts :force, Boolean, default: false
         accepts :require_ci, Boolean, default: false
         accepts :env, Hash, default: {}
+        accepts :safeties_enforced, Boolean, default: false
```
casperisfine, line 16: Should this default to `true`?

kwboyd-shopify: The purpose of this param is a hard override of `ignore_safeties` (as in, we already enforced safeties before this API request) - so it's defaulting to `false` since we want safeties to usually be reported on. Naming this is tricky so I'm open to suggestions!

casperisfine: ah, this is purely to switch the DB field so that it's displayed in the UI? 

Then I don't think adding a parameter is the clean way to do this. It would be much better to automatically set that to true if the `force` was actually needed.

PR #1297, opened 2023-02-22
More explicit wording when failing on missing shipit.yml stanzas

timothysmith0609:
Makes explicit that you should be looking at a _specific_ `shipit.yml` when a deploy fails to locate an appropriate stanza

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -60,9 +60,9 @@ en:
       locked: Locked
   deploy_spec:
     hint:
-      deploy: Impossible to detect how to deploy this application. Please define `deploy.override` in your shipit.yml
-      rollback: Impossible to detect how to rollback this application. Please define `rollback.override` in your shipit.yml
-      fetch: Impossible to detect how to fetch the deployed revision for this application. Please define `fetch` in your shipit.yml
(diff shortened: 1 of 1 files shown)


ruionweb (approved):

casperisfine (dismissed):

config/locales/en.yml
```diff
@@ -60,9 +60,9 @@ en:
       locked: Locked
   deploy_spec:
     hint:
-      deploy: Impossible to detect how to deploy this application. Please define `deploy.override` in your shipit.yml
-      rollback: Impossible to detect how to rollback this application. Please define `rollback.override` in your shipit.yml
-      fetch: Impossible to detect how to fetch the deployed revision for this application. Please define `fetch` in your shipit.yml
+      deploy: Impossible to detect how to deploy this application. Please define `deploy.override` in your shipit.$SHIPIT_ENVIRONMENT.yml
```
casperisfine, line 63: Hum, actually, the strict environment thing is optional, so this may be confusing for other users.

Also this `$SHIPIT_ENVIRONMENT` isn't interpolated, so not sure that really helps much.

casperisfine (commented):
I'm really not sure this would help much.

PR #1300, opened 2023-03-14
Add fetch_commit during checkout_repository

ruionweb:
We've been seeing numerous accounts of `fatal: reference is not a tree:<commit-sha>` error  while checking out the [commit](https://github.com/Shopify/shipit-engine/blob/master/lib/shipit/task_commands.rb#L55), even though we fetch the [branch ref](https://github.com/Shopify/shipit-engine/blob/master/lib/shipit/task_commands.rb#L55) before hand.

Reproducing it locally, fetching the commit before the checkout resolves the error.

### DIFF
--- a/app/models/shipit/task_execution_strategy/default.rb
+++ b/app/models/shipit/task_execution_strategy/default.rb
@@ -70,7 +70,6 @@ module Shipit
           @task.acquire_git_cache_lock do
             @task.ping
             unless @commands.fetched?(@task.until_commit).tap(&:run).success?
-              capture!(@commands.fetch)
               capture!(@commands.fetch_commit(@task.until_commit))
             end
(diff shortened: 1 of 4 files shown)


stephsachrajda (approved):
If this works, that's fine with me, but since you were able to reproduce it locally, were you able to validate if @kwboyd-shopify 's theory on submodules had any weight (https://gist.github.com/gitaarik/8735255#making-it-easier-for-everyone)

ruionweb:
@stephsachrajda I ran `git submodule status --recursive` on a few repos failing with this, but it didn't detect submodules. Also tried cloning with `--recurse-submodules` but still had the same issue. Perhaps I'm not running the commands correctly?

PR #1309, opened 2023-05-29
Update `ContinuousDeliveryJob` to account for concurrent tasks

ruionweb:
When you force deploy via the api controller, it sets `allow_concurrency` to true. So when the pipeline is unlocked and `ContinuousDeliveryJob` runs, it only checks for non-concurrent tasks with the [`current` function](https://github.com/Shopify/shipit-engine/blob/master/app/models/shipit/task.rb#L90). This means that another deployment of the same commit can be triggered by the `ContinuousDeliveryJob` simultaneously.

This updates `ContinuousDeliveryJob` so that it doesn't trigger a deploy if there is a concurrent task running.


app/models/shipit/stack.rb
```diff
@@ -453,6 +453,15 @@ def active_task
       @active_task ||= tasks.current
     end
 
+    def occupied?
+      !!occupied
+    end
+
+    def occupied
+      return @occupied if defined?(@occupied)
```
kartiki975, line 461: Why is this line necessary? Won't line 462 be sufficient as it is?

kartiki975 (approved):

erik-shopify (approved):

PR #1313, opened 2023-07-14
Give `ConcurrentTaskRunning` a clear error message

stephsachrajda:
### What

- Updates `ConcurrentTaskRunning` error to have a clear error message

### Why

- The error name itself is clear, but "Conflict" is not, this way api users can pass along the message

### DIFF
--- a/app/models/shipit/task.rb
+++ b/app/models/shipit/task.rb
@@ -3,7 +3,11 @@ module Shipit
   class Task < Record
     include DeferredTouch
 
-    ConcurrentTaskRunning = Class.new(StandardError)
+    class ConcurrentTaskRunning < StandardError
+      def message
+        "A task is already running."
+      end
+    end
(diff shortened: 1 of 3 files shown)


app/models/shipit/task.rb
```diff
@@ -3,7 +3,11 @@ module Shipit
   class Task < Record
     include DeferredTouch
 
-    ConcurrentTaskRunning = Class.new(StandardError)
+    class ConcurrentTaskRunning < StandardError
```
kartiki975, line 6: I saw that there are a bunch of other standard error subclasses that do not define `message`; e.g. `Error`, etc. Would those also need to be implemented for the `conflict` method in the other file... I am unsure if those other errors may get propogated all the way to that method.

kwboyd-shopify (approved):

PR #1314, opened 2023-07-26
Update to omniauth-github 2.

simi:
:information_source: I'm trying to fix security report reported to RubyGems.org instance of shipit instance. Sadly, since there is no testing shipit instance at RubyGems.org side, I have not tested this deployed, I just did quick test the app boots locally. Is there any chance to test this at Shopify side? If not, feel free to ping me, I'll try to setup proper testing environment and test the whole GH flow.

- per https://github.com/omniauth/omniauth-github/releases/tag/v2.0.0 there is no need to change anything in omniauth-github setup code
- per https://github.com/omniauth/omniauth/wiki/Upgrading-to-2.0 there are some backward incompatibilities


casperisfine:
> Is there any chance to test this at Shopify side?

Hum, I haven't done this in a very long time, but it's possible to test this locally, IIRC you don't need a publicly available instance for authentication.

> per https://github.com/omniauth/omniauth/wiki/Upgrading-to-2.0 there are some backward incompatibilities

> OmniAuth now defaults to only POST as allowed request_phase methods.

I think this will be an issue, as we directly redirect to authentication, without submitting a form or anything.

So we'll either have to disable that default and keep the old behavior (meaning CVE-2015-9284 is still a thing) or add an authentication page.

simi:
@casperisfine is there anything I can do about this to move it forward or any idea for alternative?

casperisfine:
This is all blurry to me now, but I think the comment still hold. Since shipit redirects you (with GET) when you are unauthenticated, we either need to configure OAuth to accept GET requests, or instead redirect to some login page with a form and a button users have to click on.

PR #1318, opened 2023-09-13
Added support for config files nested in .shipit/

derek-watson:
[Shopify/web](https://github.com/Shopify/web) has so many shipit configuration files in its root, they are creating noise and polluting the root folder. Added support for nesting config files in `.shipit/`. 

Screenshot 2023-09-12 at 10 12 04 PM


casperisfine:
Thanks, I corrected the priority order and also updated the list of "bare" names.

PR #1319, opened 2023-10-05
Update master to main

kartiki975:
Closes https://github.com/Shopify/services/issues/5781

After renaming the branch and following the rest of the steps in the guide, this is the last step remaining.

### Focus for reviewers
Should the changes in the second commit be kept?

### DIFF
--- a/lib/shipit/stack_commands.rb
+++ b/lib/shipit/stack_commands.rb
@@ -106,7 +106,7 @@ module Shipit
         .success?
     end
 
-    def git_clone(url, path, branch: 'master', **kwargs)
+    def git_clone(url, path, branch: 'main', **kwargs)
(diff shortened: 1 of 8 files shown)


lib/shipit/stack_commands.rb
```diff
@@ -106,7 +106,7 @@ def git_cmd_succeeds?(path)
         .success?
     end
 
-    def git_clone(url, path, branch: 'master', **kwargs)
+    def git_clone(url, path, branch: 'main', **kwargs)
```
kwboyd-shopify, line 109: I suspect that this change is fine because `branch` is a required field when creating a stack. I don't know that there's a real situation where you would end up here and `git_clone` would be nil.

kwboyd-shopify (approved):

PR #1320, opened 2023-12-14
Allow overrides for default task values via URL parameters

kovyrin:
Allow new task page to accept default value overrides for task variables. This should allow users to construct urls like `https://shipit.example.com/something/production/tasks/restart/new?REGION=gcp-us-east1` and provide their parameters with a default value.

One very important use-case here is links from documentation/playbooks, where we'd want to hardcode specific values for a task and allow the operator to review those and then click the button without having to re-type the values every time (especially dangerous when under fire).

### DIFF
--- a/app/models/shipit/task_definition.rb
+++ b/app/models/shipit/task_definition.rb
@@ -40,14 +40,6 @@ module Shipit
       "This task (title: #{@title}) cannot be shown due to an incorrect variable name. Check your shipit.yml file"
     end
 
-    def override_variables(params)
-      variables.each do |var|
-        if params.key?(var.name)
-          var.override_value(params[var.name])
-        end
-      end
-    end
-
     def allow_concurrency?
       @allow_concurrency
     end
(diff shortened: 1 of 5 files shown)


muz (approved):

app/views/shipit/_variables.html.erb
```diff
@@ -9,7 +9,7 @@
           <% if variable.select %>
             <%= field.select variable.name, options_for_select([["Please select...", { disabled: "disabled" }]] + variable.select, variable.default || "Please select...") %>
           <% else %>
-            <%= field.text_field variable.name, value: variable.default %>
+            <%= field.text_field variable.name, value: variable.value %>
```
casperisfine, line 12: I think you can implement this whole feature in the view with just:

```suggestion
            <%= field.text_field variable.name, value: params[variable.name].presence || variable.default %>
```

No model or controller layer changes required.

app/models/shipit/task_definition.rb
```diff
@@ -40,6 +40,14 @@ def render_title(env)
       "This task (title: #{@title}) cannot be shown due to an incorrect variable name. Check your shipit.yml file"
     end
 
+    def override_variables(params)
```
casperisfine, line 43: But if you go with the controller + model, then I think it would make more sense to call this `override_defaults` and set `@default` rather than a new `@override` variable.

PR #1324, opened 2024-01-26
Upgrade Rails to 7.1.1

kartiki975:
- Upgraded rails 7.1.1 for an internal Shopify ticket
- Removed mini_racer gem; see the comments below for more details


.github/workflows/main.yml
```diff
@@ -150,7 +150,7 @@ jobs:
         export SHIPIT_GEM_PATH="${PWD}"
         mkdir /tmp/new-app
         cd /tmp/new-app
-        gem install rails -v '~> 7.0.2' --no-document
+        gem install rails -v '7.1.1' --no-document
```
casperisfine, line 153:
```suggestion
        gem install rails -v '~> 7.1.1' --no-document
```

shipit-engine.gemspec
```diff
@@ -30,7 +30,7 @@ Gem::Specification.new do |s|
   s.add_dependency('octokit', '~> 4.20')
   s.add_dependency('omniauth-github', '~> 1.4')
   s.add_dependency('pubsubstub', '~> 0.2.0')
-  s.add_dependency('rails', '~> 7.0.0')
+  s.add_dependency('rails', '= 7.1.1')
```
casperisfine, line 33:
```suggestion
  s.add_dependency('rails', '~> 7.1.1')
```

kartiki975:
@casperisfine Post rails upgrade, mini_racer is breaking wherever Ruby version is less than 3.2. Looking at https://github.com/rubyjs/mini_racer?tab=readme-ov-file#supported-ruby-versions--troubleshooting, it seems like there is no workaround where the Ruby version is less than 3.2:
> make sure you have Rubygems >= 3.2.13 and bundler >= 2.2.13 installed: gem update --system

Note, I have tried this with the existing and now, the latest version of mini_racer here: [ba2c64a](https://github.com/Shopify/shipit-engine/pull/1324/commits/ba2c64a58a014e2eb394e9f263d64892a9d3ef0d).

At this point, there two actions I can take:
- upgrade Ruby to 3.2 which would mean I will have to ship the original PR I tagged you in
- removing mini_racer; do you know if this will break other changes?

Thoughts?

casperisfine:
> * removing mini_racer; do you know if this will break other changes?

We can remove mini_racer and if nodejs is around, `ExecJS` will automatically use tha, it's probably better that way.

docs/setup.md
```diff
@@ -14,7 +14,7 @@ In the future we'd like to provide it fully packaged inside a Docker container,
 
 Shipit provides you with a Rails template. To bootstrap your Shipit installation:
 
-1. If you don't have Rails installed, run this command: `gem install rails -v 7.0`
+1. If you don't have Rails installed, run this command: `gem install rails -v 7.1.1`
```
benlangfeld, line 17:
```suggestion
1. If you don't have Rails installed, run this command: `gem install rails -v 7.1`
```

kwboyd-shopify, line 17: Wherever it was previously just major.minor version, is it possible to stick to that? 7.1 instead of 7.1.1

kartiki975: I am not sure if I am quite following this. Is the ask here to switch upgrade to 7.1 instead of 7.1.1?

I will DM you privately to vet this, Kate :)

erik-shopify (approved):

jgaskins (approved):

casperisfine:
Thanks for taking care of the upgrade, LGTM.

However next time please try to keep a clean git history, `wip` and `
Add minor tweaks` commit messages cause pain down the line when investigating the reason for a change.

casperisfine:
> Worth setting the repo to force squash merges?

It's one solution, but has its own downsides as some PR do legitimately contain more than one commit, e.g. a refactoring followed by a feature, etc.

Anyway, this was a general advice for contribution to all projects, not specific to `shipit-engine`.

PR #1350, opened 2024-05-20
Round the `finished in` second count

DazWorrall:
This amount of precision seems a bit excessive, right? Is there a reason we haven't done this before?

Screenshot 2024-05-20 at 13 49 36


casperisfine (approved):
Make sense to me.

PR #1356, opened 2024-07-29
Add stack & accessor to Filesystem

wayt:
This adds `stack` attributes to `Filesystem` in order to use it in plugins to find information about the current stack.

### DIFF
--- a/test/models/deploy_spec_test.rb
+++ b/test/models/deploy_spec_test.rb
@@ -5,7 +5,8 @@ module Shipit
   class DeploySpecTest < ActiveSupport::TestCase
     setup do
       @app_dir = '/tmp/'
-      @spec = DeploySpec::FileSystem.new(@app_dir, 'env')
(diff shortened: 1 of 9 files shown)


DazWorrall (approved):

wayt:
Added a changelog note and `0.40.0` version release.

Never did that here, is that all valid @DazWorrall @casperisfine ?

test/models/deploy_spec_test.rb
```diff
@@ -287,7 +288,8 @@ class DeploySpecTest < ActiveSupport::TestCase
     end
 
     test '#gemspec gives the path of the repo gemspec if present' do
-      spec = DeploySpec::FileSystem.new('foobar/', 'production')
+      # spec = DeploySpec::FileSystem.new('foobar/', 'production')
```
casperisfine, line 291:
```suggestion
```

PR #1358, opened 2024-08-15
Make --quiet flag dynamic in git commands

mfilipe:
At PowerHRG, we're currently running on a fork which removes `--quiet` from git commands to avoid timing out log collection on long git clone operations.

Other consumers of shipit-engine might make use of progress output as discussed at https://github.com/Shopify/shipit-engine/pull/1278.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -220,7 +220,7 @@ GEM
     multi_xml (0.6.0)
     multipart-post (2.3.0)
     mutex_m (0.2.0)
-    mysql2 (0.5.3)
+    mysql2 (0.5.6)
     net-imap (0.4.12)
       date
       net-protocol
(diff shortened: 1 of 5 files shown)


mfilipe:
> Sorry for the delay I was in vacation. I think your use case make sense, but while we're at it we might as well allow to add any arbitrary git flags.

```yml
production:
  git_flags: ['--quiet']
```

Do you mean anything like that?

casperisfine:
> Do you mean anything like that?

Yeah, a list of flags that are just appended. But actualy, if `--quiet` is added by default, no flags can undo it, so you solution might be inevitable if we want `--quiet` to be the default.

I'll do another pass on your PR in a bit.

casperisfine:
Yeah, I figured, but I don't think it's really worth it.

Issue #1100, opened 2020-08-07
[Proposal] "Continuous Deployment Hours"


jayzalowitz:
Continuous deployment is great, but deploy hours are great things to enforce in this system.

Would yall be willing to add something that is basically business hours?

@cloudcustodian has some good examples https://cloudcustodian.io/docs/quickstart/offhours.html

benlangfeld:
We implement this as a pre-deploy check in the shipit.yml. I can drop an example shortly. It works well for us.

benlangfeld:
Apologies for the delay, but here is our implementation:

In shipit.yml:

```yml
review:
  checks:
    - bundle exec bin/prod_deploy_time_limit
```

In `bin/prod_deploy_time_limit`:

```ruby
#!/usr/bin/env ruby

# frozen_string_literal: true

require "active_support/all"
require "tzinfo"
require "tzinfo/data"

tz = Time.find_zone("America/New_York")
now = tz.now

puts "Considering current time #{now}"

if now.saturday? || now.sunday?
  puts "Deploys not allowed at the weekend"
  exit 1
end

if now.friday? && now > tz.parse("16:30 pm")
  puts "Deploys not allowed on Friday after 16:30 Eastern"
  exit 1
end

puts "No time restriction active. Deploys allowed."
exit 0
```

casperisfine:
I think this is an interesting feature, and I do remember some interest for it. 

It can indeed be done with a pre-deploy check, but it could make sense as a more polished feature.

I'm just not sure wether it would make more sense as a global configuration or as a part of `shipit.yml` (or both?)

benlangfeld:
> I'm just not sure wether it would make more sense as a global configuration or as a part of `shipit.yml` (or both?)

In our case this is applied only to a single stack, not even globally to a whole repo, let alone globally to all repos. I think it needs to go in `shipit.yml`.

benlangfeld:
@powerhome Will contribute this feature once #1102 is merged.

casperisfine:
> once #1102 is merged.

Damn, I really need to review that, but it's really too big. Any way you could break it up? e.g. by not submitting the API just yet etc?

benlangfeld:
[...]

PR #1361, opened 2024-08-24
[Feature] Continuous Deployment Hours

davidcornu:
Implements https://github.com/Shopify/shipit-engine/issues/1100.

## User-visible changes

Adds a new settings page where the continuous deployment schedule can be configured.

Stacks settings | Schedule
---- | ----
CleanShot 2024-08-24 at 15 51 35@2x | CleanShot 2024-08-24 at 15 51 43@2x

## Code changes

- Adds a new `continuous_delivery_schedules` table, structured as a series of `*_enabled`, `*_start`, and `*_end` fields for each day of the week. 
    - By default every day is enabled with a time window that covers the whole day
- Adds a new `ContinuousDeliverySchedule` model which wraps the logic to determine whether a deploy should be allowed
- `Stack` records can have one or zero associated `ContinuousDeliverySchedule` records (enforced via unique index)
- Adds new routes, a controller, and a view to render the UI
    - Also adds a little bit of coffeescript to power the <kbd>Copy to all</kbd> button
- Adds a check in `ContinuousDeliveryJob` to bail out if we are currently outside of a deployment window
    -  From what I can tell this job runs every minute so I deemed this sufficient (one might imagine wanting to pause the job until the next deployment window but that'd likely be overkill)

## Notes

- The UI is largely inspired by Google Calendar's settings for working hours
    CleanShot 2024-08-24 at 16 01 16@2x
- I considered allowing the timezone to be configured per-schedule as this would be useful for cross-timezone teams. 
    - Howewer
        -  Doing this with `ActiveRecord::Type::Time` would require very localized timezone overrides (e.g. when parsing form data) which would be tricky to get right, easy to break, and may cause unintended consequences elsewhere.
        - A more robust approach (IMHO) would involve using a custom type (to basically store "HH:MM" as a string) and control all conversions to and from timestamps.
    - I'm happy to explore either of these options but I figured I would submit the most straightforward route as
        - This config is unlikely to change often
        - In most scenarios I would expect the host app to have to most appropriate timezone configured for the people using it

### DIFF
--- a/app/controllers/shipit/continuous_delivery_schedules_controller.rb
+++ b/app/controllers/shipit/continuous_delivery_schedules_controller.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+module Shipit
+  class ContinuousDeliverySchedulesController < ShipitController
+    before_action :load_stack
+
+    def show
+      @continuous_delivery_schedule = @stack.continuous_delivery_schedule || @stack.build_continuous_delivery_schedule
+    end
+
+    def update
+      @continuous_delivery_schedule = @stack.continuous_delivery_schedule || @stack.build_continuous_delivery_schedule
+      @continuous_delivery_schedule.assign_attributes(continuous_delivery_schedule_params)
+
+      if @continuous_delivery_schedule.save
+        flash[:success] = "Successfully updated"
+        redirect_to(stack_continuous_delivery_schedule_path)
+      else
+        flash.now[:warning] = "Check form for errors"
+        render(:show, status: :unprocessable_entity)
+      end
+    end
+
+    private
+
+    def load_stack
+      @stack = Stack.from_param!(params[:id])
+    end
+
+    def continuous_delivery_schedule_params
+      params.require(:continuous_delivery_schedule).permit(
+        :timezone_name,
+        *Shipit::ContinuousDeliverySchedule::DAYS.flat_map do |day|
+          [
+            "#{day}_start",
+            "#{day}_end",
+            "#{day}_enabled",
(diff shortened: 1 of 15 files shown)


davidcornu:
I have signed the CLA!

davidcornu:
I don't believe this CI failure (https://github.com/Shopify/shipit-engine/actions/runs/10541191179/job/29206691546?pr=1361) is related to my changes but please correct me if I'm wrong.

CleanShot 2024-08-24 at 16 14 52@2x

app/controllers/shipit/continuous_delivery_schedules_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    private
+
+    def load_stack
+      @stack = Stack.from_param!(params[:id])
+    end
+
+    def continuous_delivery_schedule_params
+      params.require(:continuous_delivery_schedule).permit(
+        :timezone_name,
```
davidcornu, line 32: Forgot to remove this from a previous implementation

app/controllers/shipit/continuous_delivery_schedules_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+          [
+            "#{day}_start",
+            "#{day}_end",
+            "#{day}_enabled",
+          ]
+        end
+      )
+    end
+
+    def operation_param
```
davidcornu, line 43: This too

casperisfine:
Looks neat. I don't have any concern about the implementation. Did you try it in prod?

PR #1363, opened 2024-08-29
Fix the SetupScript CI step

casperisfine:
Rails 7.1 isn't happy with sqlite3 V2 yet.


casperisfine:
Arf, I think I'll just upgrade to 7.2

PR #1367, opened 2024-09-19
Make "Previous Deploys" wording consistent

benorgan:
The other sections on this page refer to "commits", this PR updates the wording of the previously deployed commits to be consistent to avoid confusion.


casperisfine:
Those are not commits though. It can be confusing because the message of the tip commit is displayed, but they represent a range of commits and if you click on it it gives you the deploy logs.

From shipit modeling these are deploys, not "deployed commits".

PR #1368, opened 2024-09-19
Support subpaths/projects per stack

DazWorrall:
Allow stacks to target a subdirectory of the repo.


wayt:
This could be a great addition to work in pair with https://github.com/Shopify/infrastructure/pull/1266

Issue #1390, opened 2025-04-09
Feature request: Config inheritance


benlangfeld:
ShipIt [currently implements](https://github.com/powerhome/shipit-engine/blob/a6f1948622f6f76e4b6d61fb4f1b2e90dd9fe680/app/models/shipit/deploy_spec/file_system.rb#L106-L120) replacement config files [on a per-environment basis](https://github.com/powerhome/shipit-engine/blob/a6f1948622f6f76e4b6d61fb4f1b2e90dd9fe680/README.md?plain=1#L129-L131). This is useful when configuration is significantly different between environments.

When configuration differences are limited between environments, this can result in significant duplication. It would be useful for the more specific files to be able to contain an inheritance directive to reduce this duplication. This might look something like this:

shipit.production.yml:
```yaml
inherit_from: shipit.yml

machine:
  environment:
    PUBLIC: true
```

shipit.yml:
```yaml
machine:
  environment:
    PUBLIC: false

deploy:
  override:
    - ./some_deployment_process.sh ${PUBLIC}

...more shared stuff...
```

We'd be willing to contribute this upstream if it has a reasonable change of being merged. Looking for in-principle approval of a feature like this from a maintainer.

casperisfine:
> I'm looking for in-principle approval of a feature like this from a maintainer.

Yeah, in-principle it looks fine to me.

benlangfeld:
> > I'm looking for in-principle approval of a feature like this from a maintainer.
> 
> Yeah, in-principle it looks fine to me.

Thanks, we'll have an implementation up next week for your review.

PR #1393, opened 2025-04-15
Allow Config File Inheritance

ryanmrodriguez:
Addresses https://github.com/Shopify/shipit-engine/issues/1390

---

Allow setting an `inherit_from` key in 'shipit.yml' config files.

As of now this key expects a string. It is allowed to be chained:

``` yaml
# shipit_2.yml
inherit_from: shipit_1.yml

c: true
d: 123
```

``` yaml
# shipit_1.yml
inherit_from: shipit_base.yml

c: false

machine:
  environment:
    PUBLIC: true
    TEST: true
```

``` yaml
# shipit_base.yml

a: 123
b: test
c: 1


machine:
  environment:
    PUBLIC: false
```

Loading 'shipit_2.yml' results in:
```rb
{"a"=>123, "b"=>"test", "c"=>true, "machine"=>{"environment"=>{"PUBLIC"=>true, "TEST"=>true}}, "d"=>123}
```

Merge (deep) precedence = child overwrites parent (config file specified in `inherit_from` key).

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ The settings in the `shipit.yml` file relate to the different things you can do 
 * [Custom Tasks](#custom-tasks) (`tasks`)
 * [Custom links](#custom-links) (`links`)
 * [Review Process](#review-process) (`review.checklist`, `review.monitoring`, `review.checks`)
-* [Inherit From](#inherit-from)(`inherit_From`)
+* [Inherit From](#inherit-from)(`inherit_from`)
 
 All the settings in `shipit.yml` are optional. Most applications can be deployed from Shipit without any configuration.
 


app/models/shipit/deploy_spec/file_system.rb
```diff
@@ -136,6 +137,20 @@ def app_name
         @app_name ||= Shipit.app_name.downcase
       end
 
+      SHIPIT_CONFIG_INHERIT_FROM_KEY = "inherit_from"
+      def build_config(path, config_obj)
+        return config_obj if config_obj.blank? || !config_obj.key?(SHIPIT_CONFIG_INHERIT_FROM_KEY)
+
+        inherits_from_path = path.dirname.join(config_obj.delete(SHIPIT_CONFIG_INHERIT_FROM_KEY))
+        if inherits_from_path.exist?
```
casperisfine, line 145: Open question: does a non existing path should noop or result in an error?

The problem is kinda always the same though, we don't have a good way of reporting `shipit.yml` errors, so perhaps nooping is the reasonable thing to do.

casperisfine (approved):

Issue #1371, opened 2024-10-16
Disparate Ruby Versions


Phillita:
Hello ShipIt Team!

First off, thank you for your commitment to OSS. 

I've run into an issue during the setup of the ShipIt project. I've initiated the repo with Ruby 3.1 and went through the setup steps. All was going well until I tried to deploy a test application. This application runs on Ruby 3.4. I get the following error at the bundle step:

![Image](https://github.com/user-attachments/assets/d2dd2316-1cde-47b3-a635-1be9f170f3a6)

I've tried running a setup using the shipit.yml where I set the Ruby version in the `dependencies.pre` and overrode using the `dependencies.override` and have had no luck getting past this issue. Below is me trying to run the install by installing/setting the version before the bundle install:

![Image](https://github.com/user-attachments/assets/3f751c56-99ce-4afc-8de6-e111e23cd167)

Unfortunately, many projects won't have the same version as the ShipIt project. Can ShipIt detect/install/use the Ruby version from the project it is trying to deploy? I'm likely doing something wrong.

casperisfine:
> my Unix system

What system is that?

casperisfine:
I just tried it and it's indeed broken, I suppose `sed` changed on OSX a while ago.

I think we should change it for `/usr/bin/sed -i '' '/^ruby /d' Gemfile`

Phillita:
Tested on my end and that works as expected!

PR #1425, opened 2025-09-24
Enable Post Scripts Process on Failure

Phillita:
# Motivation

I am currently trying to run a PoC of Shipit over our current manual deploy process. I noticed there are pre and post scripts that can be run, but I found out that on command errors, the post script won't be run.

My current system let's us know when deploys fail, and I think allowing post scripts to run on failure will get me the needed information in my secondary tracking system. 

# Changes

* Update the bundler discovery section to let the `sed` command work with modern macs (issue #1371)
* Add config for `post` scripts on both deploy and rollback sections to run `on_error`
* Updated the code to allow for post scripts to run on command errors
* Added extra env config (FAILED/FAILURE_MESSAGE) for use in pre-/post-scripts
* Updated README to include `dependencies.bundler.frozen`
* Updated README to include sections on post scripts with on_error flag
* Refactored README to use mostly pure markdown instead of HTML

## Other fixes

* Restored functionality to the stack show page for previously deployed tasks
    * There was a change to collection rendering which removed a warning but this broke functionality
    * Clicking links when rendering the generic partial led to 500 errors
    * You could no longer rollback to previous deploys or mark health statuses


Phillita:
I'm working on the CLA with my org now.

PR #1432, opened 2025-10-29
Allow rollbacks to specify their own variables

timothysmith0609:
# Context

Deploys and rollbacks are similar, yet separate, concepts in Shipit. In fact, `Rollback` is a subclass of `Deploy`. Currently, deploys and rollbacks reference the same variables, namely those specified in `deploy.variables` in `DeploySpec` ([example](https://github.com/Shopify/shop-server/blob/main/areas/platforms/shop-server/shipit.production.yml#L34)).

However, sometimes we want different values/defaults between deploy and rollback. The motivating example we're using is when a deploy wants to default to a phased rollout strategy, but wants to default to a parallel rollout strategy during rollback.

In order to achieve that, we need to be able to encode separate values for both deploy and rollback. In practice, some apps are already doing this (see shop-server, above), even though their configuration is erroneous and does not behave as they would expect. This PR aims to rectify that shortcoming.

# A note on backwards compatibility

To preserve existing behaviour, Shipit Engine will continue to use `deploy.variables` if no rollback variables are explicitly set. This seems like the 'safest' option, at least in the medium term, to avoid any unexpected surprises downstream.

# Acceptance testing

For these tests I use the same `shipit.yml` file with the following:

```
shared_data:
  variables:
    deploy_strategy: &deploy_strategy
      name: "SHIPIT_DEPLOY_STRATEGY"
      title: "Deploy Strategy"
      default: phased
      select:
        - phased
        - parallel

deploy:
  variables:
  - *deploy_strategy
  retries: 2
  override:
    - echo "SHIPIT DEPLOY STRATEGY ${SHIPIT_DEPLOY_STRATEGY}"
    - sleep 2
    - echo deployed

rollback:
  variables:
    - <<: *deploy_strategy
      default: parallel
```

The key thing to note is that both `deploy` and `rollback` define the same variables, and `rollback` changes the default to `parallel`.

**Before the change (rollback default is phased)**

image

**After the change (rollback now uses `rollback` default of parallel)**

image

### DIFF
--- a/app/models/shipit/deploy_spec.rb
+++ b/app/models/shipit/deploy_spec.rb
@@ -141,10 +141,10 @@ module Shipit
 
     def rollback_variables
       if config('rollback', 'variables').nil?
-        Array.wrap(config('rollback', 'variables')).map(&VariableDefinition.method(:new))
-      else
         # For backwards compatibility, fallback to using deploy_variables if no explicit rollback variables are set
         deploy_variables
+      else
+        Array.wrap(config('rollback', 'variables')).map(&VariableDefinition.method(:new))
       end
     end
 


app/models/shipit/stack.rb
```diff
@@ -519,7 +519,7 @@ def self.from_param!(param)
     end
 
     delegate :plugins, :task_definitions, :hidden_statuses, :required_statuses, :soft_failing_statuses,
-             :blocking_statuses, :deploy_variables, :filter_task_envs, :filter_deploy_envs,
+             :blocking_statuses, :deploy_variables, :rollback_variables, :filter_task_envs, :filter_deploy_envs,
```
bostonaholic: Missing new filter delegation.

```suggestion
             :blocking_statuses, :deploy_variables, :rollback_variables, :filter_task_envs, :filter_deploy_envs, :filter_rollback_envs,
```

timothysmith0609: I've added this, but in practice it's not used, since `#build_rollback` is called on a Deploy object directly, not via `Stack`.

It gets even messier. `Deploy` has `delegate :filter_deploy_envs, to: :stack` but it looks like this is also never called, in practice, either :/

PR #1443, opened 2025-12-10
Upgrade Rails to 8.1

byroot:
I'm trying to upgrade ruby-bench to Rails 8.1 and shipit is one of the benchmarks.


shipit-engine.gemspec
```diff
@@ -34,7 +34,7 @@ Gem::Specification.new do |s|
   s.add_dependency('omniauth-github', '~> 1.4')
   s.add_dependency('paquito')
   s.add_dependency('pubsubstub', '~> 0.2')
-  s.add_dependency('rails', '~> 8.0.1')
+  s.add_dependency('rails', '~> 8.1.2')
```
Edouard-chin: Can we remove the upperbound constraint and only have `rails, >= 8.0.1' ?

PR #1474, opened 2026-03-19
Auto-skip dependency installation for production_platform repos

DudeRandom21:
## Summary
- When a repo has `production_platform` configured in its shipit.yml and all explicitly configured deploy/rollback/task steps match a configurable safe-command allowlist, skip dependency installation automatically
- The allowlist is configured via `Shipit.safe_deploy_command_prefixes` (defaults to empty — feature is opt-in per instance)
- Logs a warning when deps are skipped so it's visible in deploy output and Rails logs
- Explicit `dependencies.override` config is always respected — this only affects auto-discovery

Refs Shopify/continuous-deployment#2454

## Motivation

Repos deploying via `production-platform-next` don't need local Ruby/Node dependencies installed on the shipit worker — the app runs in pre-built Docker containers. But shipit's auto-discovery detects `Gemfile`/`package.json` and runs `bundle install` anyway. When the worker's Ruby version doesn't match the repo's required version, the install fails and blocks deploys entirely, preventing Ruby version upgrades.

## Design decisions

- **Configurable allowlist**: The safe command prefixes are configured at the instance level via `Shipit.safe_deploy_command_prefixes`, defaulting to an empty array (feature disabled). Each shipit deployment sets its own prefixes in an initializer.
- **Opt-in**: When the list is empty, the feature is completely disabled — no behavior change for existing deployments.
- **Allowlist matching**: Only skip when ALL steps match the configured safe commands. Unknown commands fall through to normal dependency installation.
- **Explicit config only**: We only inspect `deploy.override`, `deploy.pre/post`, `rollback.override`, `rollback.pre/post`, and `tasks` from config. Auto-discovered steps are not inspected — if no overrides exist, we conservatively install deps.
- **Empty steps guard**: If no explicit steps are configured at all, we don't skip (avoids `[].all?` vacuous truth).

## Configuration

In your shipit instance's initializer:

```ruby
Shipit.safe_deploy_command_prefixes = %w[
  production-platform-next
  kubernetes-deploy
  kubernetes-restart
]
```

## Test plan
- [x] production_platform + safe steps → deps skipped
- [x] production_platform + unsafe deploy steps → deps NOT skipped
- [x] production_platform + unsafe task steps → deps NOT skipped
- [x] production_platform + unsafe deploy.pre → deps NOT skipped
- [x] production_platform + unsafe rollback → deps NOT skipped
- [x] production_platform absent → normal behavior
- [x] Explicit `dependencies.override` → respected regardless
- [x] Pre/post dependency steps preserved when core install skipped
- [x] kubernetes-deploy prefix → skipped
- [x] Unknown command → falls through to discovery
- [x] No overrides configured at all → conservative, does not skip
- [x] Empty safe_deploy_command_prefixes → feature disabled
- [x] Custom prefixes work correctly
- [x] Shipit.safe_deploy_command_prefixes defaults to empty array
- [x] Shipit.safe_deploy_command_prefixes can be set
- [x] Full test suite passes (172 runs, 315 assertions, 0 failures)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/test/models/deploy_spec_test.rb
+++ b/test/models/deploy_spec_test.rb
@@ -9,6 +9,16 @@ module Shipit
       @stack = shipit_stacks(:shipit)
       @spec = DeploySpec::FileSystem.new(@app_dir, @stack)
       @spec.stubs(:load_config).returns({})
+      @original_safe_deploy_command_prefixes = Shipit.safe_deploy_command_prefixes
+      Shipit.safe_deploy_command_prefixes = %w[
+        production-platform-next
+        kubernetes-deploy
+        kubernetes-restart
+      ]
+    end
+
+    teardown do
+      Shipit.safe_deploy_command_prefixes = @original_safe_deploy_command_prefixes
     end
 
     test '#supports_fetch_deployed_revision? returns false by default' do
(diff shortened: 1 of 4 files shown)


test/models/deploy_spec_test.rb
```diff
@@ -9,6 +9,16 @@ class DeploySpecTest < ActiveSupport::TestCase
       @stack = shipit_stacks(:shipit)
       @spec = DeploySpec::FileSystem.new(@app_dir, @stack)
       @spec.stubs(:load_config).returns({})
+      @original_safe_deploy_command_prefixes = Shipit.safe_deploy_command_prefixes
+      Shipit.safe_deploy_command_prefixes = %w[
+        production-platform-next
+        kubernetes-deploy
+        kubernetes-restart
```
timothysmith0609, line 16: Oof, is anything still actually using `kubernetes-deploy` or `kubernetes-restart`? That's pretty ancient code. That same tool has been renamed to `krane`, fwiw ([repo for ref](https://github.com/Shopify/krane))

timothysmith0609 (changes requested):
I don't think it's appropriate to encode this logic here in the OSS gem code: can we move this to our internal shipit?

DudeRandom21:
> I don't think it's appropriate to encode this logic here in the OSS gem code: can we move this to our internal shipit?

I wasn't sure how I would do this purely in the internal layer but I'll dig into it and likely close this PR. Thanks

DudeRandom21:
Moving this logic entirely to the shipit app — production_platform is a shipit-internal concept, not an engine concern. The skip-dependencies behavior will be implemented via a prepended module in the app, matching the existing pattern for production platform discovery.

test/models/deploy_spec_test.rb
DudeRandom21: Good catch. No nothing seems to be using it anymore. It got picked up because it still exists here https://github.com/Shopify/shipit-engine/blob/main/app/models/shipit/deploy_spec/kubernetes_discovery.rb

If the tool was renamed do you think it's worth fixing the auto discovery then?

PR #1475, opened 2026-03-19
Get rid of coffee-rails

k0kubun:
Same approach as #1467 for sassc-rails: compiled all CoffeeScript files to plain JavaScript using the CoffeeScript 1.12.7 compiler (matching coffee-rails 5.0), then removed the coffee-rails dependency entirely.

This is needed because coffee-rails depends on coffee-script which depends on execjs, causing issues in ruby-bench
(https://github.com/ruby/ruby-bench/issues/493).

- Compiled 18 .coffee files to .js using `coffee -bc` (bare compile)
- Preserved sprockets directives (require, require_tree, require_self)
- Preserved ERB in task/notifications.js.erb for image_path helpers
- Removed coffee-rails from gemspec and lib/shipit.rb
- No changes needed to HTML templates or asset precompile configuration


k0kubun:
> https://github.com/Shopify/shipit-engine/pull/1467
> We may also want to consider doing something similar for coffee-rails

This PR does this. WDYT? @Edouard-chin @etiennebarrie @rafaelfranca cc: @byroot

byroot:
> WDYT?

If it's necessary, we can do it, but I think as others mentioned, ruby-bench can just not compile assets.

`sassc` was important because it was failing during boot, but here it's just a codepath we don't wanna call.

k0kubun:
Thanks for your feedback. I guess https://github.com/ruby/ruby-bench/pull/495 obviates this, so I'm gonna close this.

